PR #7, opened 2020-03-16
Add launch and attach svelte-vscode extension settings to launch.json

jasonlyu123:
Also fix language server module path in extension debugging, the origin method doesn't seems to work with yarn's dependencies hoisting.


orta:
Great! Thanks!

PR #8, opened 2020-03-17
Update service.ts to fix the errors with imports

fatmirsukaliq:
This resolves the following problems that we've encountered.

**Problem 1**
When using `paths` in `tsconfig.json`, e.g

```json
"paths": {
      "@Components/*": ["components/*"]
 }
```

and then attempting to import a file in Svelte, e.g

```
import CoolComponent from '@Components/CoolComponent.svelte'
```

you receive a `2307 ("Cannot find module")` error.

**Problem 2**
When using this [type](https://github.com/pyoner/svelte-typescript/blob/master/packages/types/types/svelte/svelte.d.ts) on `tsconfig.json`

And then trying to import something like e.g

```
import Editor from '/.Editor.svelte'
```


orta:
Looks good - thanks!

PR #13, opened 2020-03-25
Wrap plugin execution in try catch

dummdidumm:
This way, if a plugin for whatever reason fails, it won't destroy the whole promise chain.

Happened to me when node-sass was not found, then the CSSPlugin failed, then everything failed.

### DIFF
--- a/packages/language-server/src/lib/PluginHost.ts
+++ b/packages/language-server/src/lib/PluginHost.ts
@@ -46,16 +46,20 @@ export class PluginHost {
         switch (mode) {
             case ExecuteMode.FirstNonNull:
                 for (const plugin of plugins) {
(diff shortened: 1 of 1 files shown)


orta:
Yeah, this is a solid idea

PatrickG:
But now they are executed one after the other?

dummdidumm:
I think they still get executed in parallel. The map function has no built in way of doing "I wait for the previous item if it's a promise", so it just maps all plugins synchronously and then hands the array of promises to Promise.all, which then handles them in parallel.

PatrickG:
Yea, you are right.
Sorry for the interruption :D

dummdidumm:
No problem :)

PR #14, opened 2020-03-25
Only use certain trigger characters for Typescript

dummdidumm:
The language service throws an error if the character is not a valid trigger character.
Also, the completions are worse.
Therefore, only use the characters the typescript compiler treats as valid.


orta:
👍

PR #16, opened 2020-03-26
drop svelte 2 support

### LINKED ISSUE #12: Remove Svelte 2 support?
The current implementation contains code for dealing with Svelte 2 files. Do we really support Svelte 2 going forward? I feel that this would be a lot of work. Maybe remove the Svelte 2-related code?

dummdidumm:
Closes #12 

I hope I caught all occurences. I also adjusted the tests even if that was not really necessary, but it will cause less confusion later on ("why are the tests with .html-endings?").


orta:
Looks good to me! Thanks

Issue #11, opened 2020-03-24
Overview of approaches on a language server


dummdidumm:
This thread is intended for an overview of approaches on how to implement a language server for svelte files. Discussion on a prefered solution can be done here aswell.

# Current state of the language server

Currently syntax highlighting and some basic autocompletion works for svelte files.
Rich intellisense however is not provided, does not know about all the typescript files in the workspace, and also has no type information about the svelte components.
The same is true for the html part where special svelte syntax is highlighted but intellisense is not working. The vscode-html-languageservice is used which does not know about the special syntax by default.
The current language server does use the svelte compiler for parsing files and getting diagnostics like "no alt attribute on this img tag".
A more indepth analysis: [Comment](https://github.com/sveltejs/language-tools/issues/11#issuecomment-606508488)

# Existing approaches/solutions

Disclaimer: I'm not the author of any of the existing solutions, I did look at the code however. If some of the information is wrong or not detailed enough, or you are missing a solution, feel free to comment, I will adjust it.

###  https://github.com/alexprey/sveltedoc-parser 

Uses [htmlparser2](https://github.com/fb55/htmlparser2) to parse the html part of a svelte file. Uses hooks of the html parser to add custom parsing and extract relevant information.
Uses [espree](https://github.com/eslint/espree) to parse the script parts of a svelte file. Then walks the AST to extract relevant information.
Provides the results in a JSON-Format.

https://github.com/ArdenIvanov/svelte-intellisense uses it to parse svelte files. It uses the result to attach some additional behavior.

Since this uses a javascript parser, it will not work with typescript.

## https://github.com/halfnelson/svelte2tsx

Uses svelte's compiler to parse the HTMLx part of a svelte component. Then transforms the HTMLx plus the original script part to a tsx file with a self-written transformer.

https://github.com/simlrh/svelte-language-server/blob/feature/extend-ts-support/src/plugins/ts-svelte/service.ts (a fork from svelte-language-server) uses it to create shadow-tsx-files of svelte-files. It then uses typescript's language service, but proxies the requests: It replaces the original file path with the file path to the generated tsx-file. The typescript language server therefore checks the generated tsx files. Results are transformed back to get correct document positions. 

https://github.com/halfnelson/svelte-type-checker-vscode is another language server using the svelte2tsx-library.

## https://github.com/marcus-sa/svelte-ts

Description taken from [implementer's comment](https://github.com/sveltejs/svelte/issues/3677#issuecomment-541039240):

Compiler:

[...]

PR #19, opened 2020-03-26
Fixing reactive values in typescript

fatmirsukaliq:
Hi there!

When using store reactive values that are prefix with $ sign, TypeScripts reports the 2552 error that states the following:

`'Cannot find name '{0}'. Did you mean '{1}'?'`

although the variable is declared and valid, so this PR will suppress those errors in case they are defined and start with $.

Please let me know your thoughts!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -65,3 +65,5 @@ typings/
 
 # tdsx
 dist
+
+.DS_STORE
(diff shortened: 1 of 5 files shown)


orta:
Hi fatmir - I think this is probably blocked on #11  because handling this aspect of the code is one of the biggest challenges facing these tool

dummdidumm:
With #57 we now have support for the `$` syntax. Still, thanks for your efforts!

PR #24, opened 2020-04-03
Fix and refactor setting/getting config

dummdidumm:
- Simplifying retrieval of capabilities
- Passing initial config to language server and reading it (was not done before).
- Refactor how config is accessed. Plugins no longer contain parts of the config, they now all request them from the PluginHost. Also typing config. Makes code clearer.


orta:
Rockin!

PR #27, opened 2020-04-10
Fix README.md VSCode extension description

carloslfu:


Conduitry:
afaik, the goal is in fact to produce an editor-agnostic language server, not just to create a VS Code extension.

orta:
This PR is correct - the repo has two projects right now, the language server which is described as:

> The language server for Svelte. Built from UnwrittenFun/svelte-language-server to become the official language server for the language.

and the vscode extension which this PR changes to not be a copy pasta of the above 👍

orta:
I'm open for other tools which use the LSP in this repo to migrate into here also over a long time period too - if there are people which will maintain it

PR #30, opened 2020-04-20
(fix) fix html dataprovider

dummdidumm:
Reenables svelte specific suggestions like "on:click".

The plugin relies on some internal variables of the vscode-html-languageservice. These internals changed, so I updated the dataprovider to rely on the new internals.

I also asked the vscode-html-languageservice maintainers if they could make it public https://github.com/microsoft/vscode-html-languageservice/issues/81


orta:
Great work!

PR #38, opened 2020-04-24
Fix positions of ts getDefinitions and getCodeActions

dummdidumm:
Don't wrap the other documents with TypescriptDocument since these are not svelte files.

This slipped in during the big code restructuring.


orta:
Cool, good spot

PR #39, opened 2020-04-25
[Discussion] Setup prettier and pre-commit hook

jasonlyu123:
Since this repo already has prettier config in the root. Maybe we could make it the official format of this repo so that we don't have to resolve conflicts that are just formatting issues.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,1 @@
+package.json
--- a/package.json
+++ b/package.json
@@ -10,9 +10,20 @@
   "scripts": {
     "build": "tsc -b",
     "test": "CI=true yarn workspaces run test",
-    "watch": "tsc -b -watch"
+    "watch": "tsc -b -watch",
+    "prettier": "prettier --write **/src/**/*.{ts,js}"
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "pretty-quick --staged --pattern '**/src/**/*.{ts,js}'"
+    }
   },
   "dependencies": {
     "axios": "0.19.2"
+  },
+  "devDependencies": {
+    "husky": "^4.2.5",
+    "prettier": "^2.0.5",
+    "pretty-quick": "^2.0.1"
   }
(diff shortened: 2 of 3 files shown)


Conduitry:
I'm probably not going to be contributing much to this particular repo myself, but I'm rather against the idea of automatically installing pre-commit Git hooks. These have come up a couple of times in the main Svelte repo [here](https://github.com/sveltejs/svelte/pull/4123) and [here](https://github.com/sveltejs/svelte/pull/4378).

jasonlyu123:
Fair point. I do sometimes get annoyed by prettier + eslint and the pre-commit-hook setup in our company's project.
Or maybe we could remove the `.prettierrc` and move to eslint approach like the main svelte repo.

dummdidumm:
I really like prettier because I don't have to care about formatting anymore. And if the formatted code sometimes looks shitty, so be it. I also think that some kind of enforced style guide wouldn't hurt. But from reading the other issues, that could have problems on its own. Maybe moving that to the CI would be a good thing. Since I never have done that, how exactly would that work?

orta:
I'm on team 'lets cargo-cult the main repo' for this stuff

Though personally I dive real deep [on commit](https://github.com/microsoft/TypeScript-Website/blob/v2/.huskyrc.js) [hooks](https://github.com/danger/peril/blob/master/api/package.json#L156-L172) to avoid finding out 30m later on CI I didn't do something trivial.

dummdidumm:
Fair Point. What would following the cult mean for us?
- add the eslint rc, eslint ignore and editorconfig  from main repo
- add lint task and use that as a check in the CI

Anything else?
I would like to keep the prettierrc just so that those who use it at least have the same formatting.

orta:
Yeah, and keeping the prettierrc makes sense to me (so long as its settings conform to whatever style rules are in the eslint )

jasonlyu123:
Ok! I'll close this.

PR #40, opened 2020-04-25
Adds a deploy to npm step for the language-server

### LINKED ISSUE #37: Publish language-server to npm
There are some requests to use the language-server in other IDE environments (#34). We could have all plugins for these IDEs in this repo, but this might get cumbersome. So it would be better to provide the language server as a npm package.

Do we want to do this, and if yes, when?
If yes, what's the package name? (bikeshedding incoming)

orta:
Fixes #37 

Adds the same 'deploy if there's something to deploy' rules to the language service on a nightly basis

### DIFF
--- a/.github/workflows/Deploy.yml
+++ b/.github/workflows/Deploy.yml
@@ -34,11 +34,22 @@ jobs:
       # So, remove the workspace
       - run: "rm package.json yarn.lock"
 
-      # Re-run the yarn install outside of the workspaceSetup the language server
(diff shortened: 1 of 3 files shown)


dummdidumm:
Seems like it did not work, https://www.npmjs.com/package/svelte-language-server still has "last publish 7 months ago".

orta:
lols, https://github.com/sveltejs/language-tools/commit/693aed33c55e26609ca4b3d78bc5282aac8d4652

PR #45, opened 2020-04-28
Add dev tools like eslint, .editorconfig

jasonlyu123:
More info see the discussion of  #39 and the conclusion https://github.com/sveltejs/language-tools/pull/39#issuecomment-619376180

I moved the eslint config and .editorconfig from the main svelte repo and tune it to fit the coding style of this repo.

Does everyone have any opinion on the rules? Or what do I missed from the main repo?

this is the `eslintrc` of the main svelte repo:
https://github.com/sveltejs/svelte/blob/master/.eslintrc.js

Fix exist lint error would come later.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,12 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+indent_style = space
+indent_size = 4
+charset = utf-8
+trim_trailing_whitespace = true
+
+[package.json]
+indent_size = 2
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,2 @@
+**/dist/*
+node_modules
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,2 +1,3 @@
 **/dist/*
 node_modules
+**/scripts/*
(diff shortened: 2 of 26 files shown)


dummdidumm:
For me it seems good.

dummdidumm:
We also need to add it to the CI script https://github.com/sveltejs/language-tools/blob/master/.github/workflows/CI.yml

jasonlyu123:
Good! it failed

dummdidumm:
I think the only stuff we need to have a closer look at is the "forbidden non-null assertion"-errors, because they require some more code.

jasonlyu123:
I am also hesitant about `@typescript-eslint/consistent-type-assertions`. The default is `assertionStyle: 'as'`. but I change it to `assertionStyle: 'angle-bracket'` as it seems we have more angle-bracket in the existing codebase

dummdidumm:
As @orta said I would "cargo cult" the main repo here. If they use the `as` syntax, we will, too.

orta:
Could also turn off that option - it doesn't feel like something which is a critical lint

dummdidumm:
This looks good to me now. The non-null-assertions can be gradually resolved (or ignored) later.

jasonlyu123:
yeap! I'll make this ready for review.
I think most of them could be fixed.
Some of them can be handled more gracefully, or just change it to optional chaining when possible.

orta:
This looks great

PR #48, opened 2020-04-29
fix regex in finding the last opened svelte tag

jasonlyu123:
I found this while fixing linting errors. The regex for finding last opened and last closed tag escape `s` instead of `\s`


dummdidumm:
Good catch, thanks!

PR #50, opened 2020-04-30
refine eslint

jasonlyu123:
1. disable `no-non-null-assertion` on test
2. lint bin, scripts and disable `no-var-requires`
    on these files
3. disable `max-len` on bump-beta-version.js
4. fix lint errors in node scripts

Rather than disable lint on node scripts, I would like to change it to have custom lint rules.
About the `no-non-null-assertion` rule, I don't think it makes sense in test files. Rather than fixing them and use another assert to test it, I would just disable it.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,3 +1,2 @@
 **/dist/*
 node_modules
-**/scripts/*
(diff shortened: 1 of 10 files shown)


orta:
Cool, let's get this in 👍

PR #52, opened 2020-05-02
fix code action with non-empty next line

jasonlyu123:
Fix bug where code-action-changes with non-empty next line will not be executed.

example:
```js
function Foo() {
    this.bar = 1;
}
const foo = new Foo();
```
When chose the code action to convert to es2015 class, nothing will happen.
※This code action will only appear in js not in ts


orta:
Seems reasonable to me 👍

Issue #43, opened 2020-04-27
Tags starting with script/style wrongfully extracted 


dummdidumm:
Any tag that starts with `style` or `script` is wrongfully treated as a script/style tag.

```html
<!-- should be extracted -->
<style>h1{color:blue}</style>
```
```html
<!-- should not be extracted -->
<styless></styless>
<p>I'm below the styless component. Since the parser wrongfully thinks that was a style opening tag, I get css linting errors now</p>
```

Also see https://github.com/UnwrittenFun/svelte-language-server/issues/21

Extractor is in https://github.com/sveltejs/language-tools/blob/master/packages/language-server/src/lib/documents/utils.ts

dummdidumm:
Closed via #53

PR #53, opened 2020-05-03
Fix misconfigured attribute matching regex in tag extraction

sjafferi:
Fix bug where `extractTag` would wrongfully extract tags starting with script/style (https://github.com/sveltejs/language-tools/issues/43).

example:

```html
<styles>p{ color: blue; }</styles>
<p>bla</p>
 ></style>
```

would incorrectly match `<styles>....</style>`

### DIFF
--- a/packages/language-server/src/lib/documents/utils.ts
+++ b/packages/language-server/src/lib/documents/utils.ts
@@ -23,8 +23,9 @@ function parseAttributes(str: string): Record<string, string> {
  * @param source text content to extract tag from
  * @param tag the tag to extract
  */
+
 export function extractTag(source: string, tag: 'script' | 'style') {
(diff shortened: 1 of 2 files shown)


jasonlyu123:
Can the white space be replaced by `\s`?

dummdidumm:
Also one could put more than one space between `<style` and tags (f.e `<style   lang="scss">`). Could you modify the regex to handle this and add a test for that? Thanks! Apart from that it look great.

sjafferi:
> Also one could put more than one space between `<style` and tags (f.e `<style lang="scss">`). Could you modify the regex to handle this and what @jasonlyu123 said and add a test for that? Thanks! Apart from that it look great.

Sure! Just so I understand correctly, this is the case where multiple spaces are placed after `<style` and in between attributes (i.e. `<style     lang="scss">`)? If so, the current regex supports this, I've added a few test cases to ensure that. Let me know if this is what you were referring to.

dummdidumm:
Perfect, thanks!

PR #55, opened 2020-05-03
Copies the auth token ENV from make-monaco-builds

orta:
In a [different action I use to deploy to npm](https://github.com/orta/make-monaco-builds/blob/f17ca4a84b74aee03e4fc02d10a36571c7aa36f8/.github/workflows/nightly.yml#L13) - it uses `NODE_AUTH_TOKEN` (which just feels like a really icky name, considering node doesn't have auth) 

So let's copy that


dummdidumm:
It worked https://www.npmjs.com/package/svelte-language-server 🥳

PR #58, opened 2020-05-05
only extract attribute until tag end

jasonlyu123:
fix the issue where the attribute group of the extract tag would end at `>` in the script
example: 
```svelte
<script type="typescript">
let a = () => 'a'
</script>
```


dummdidumm:
Good catch!

PR #60, opened 2020-05-05
(fix) update prettier, make import dependent on user workspace

### LINKED ISSUE #59: svelte-vscode conflicts with prettier@2
**Describe the bug**
Projects using `prettier@2` + `prettier-plugin-svelte` for code formatting see different results from running `prettier --write` vs `svelte-vscode` + `formatOnSave`

**To Reproduce**
```
<script>
const foo = bar().baz().baf();
</script>
```
`svelte-vscode` formats this as:
```
<script>
    const foo = bar()
        .baz()
        .baf();
</script>
```
`prettier --write` formats this as
```
<script>
    const foo = bar().baz().baf();
</script>
```
**Expected behaviour**
Expect the formatting to be consistent.

Ideally, I would expect `svelte-vscode` to use some instance of `prettier` installed locally or globally before anything else.

 - OS: macOS
 - IDE: VSCode
 - Plugin Svelte Beta

**Additional context**
Looks very much like the `svelte-vscode` is formatting like `prettier@1`

dummdidumm:
This updates `prettier` and `prettier-plugin-svelte` to the latest version. Also makes `prettier` imports dynamic, similar to how `svelte` is imported.

Closes #59

.. but maybe not entirely, since `prettier-plugin-svelte` does not do dynamic imports. But it should be good enough for now.


Conduitry:
If there's something else prettier-plugin-svelte ought to be doing, do let me know. Right now the [shipped code](https://unpkg.com/prettier-plugin-svelte@1.1.0/plugin.js) just has `require('prettier')` and `require('svelte/compiler').parse(text)` which should do what I would expect we want to do - resolve the dependencies relative to wherever the plugin module lives.

dummdidumm:
Yes, it resolves where the `prettier` / `svelte` relative to where the plugin module lives - but that is the `node_modules` folder of the plugin, not the user's workspace. You can check that by using `require.resolve` and look at what the path is. To use `prettier` / `svelte` relative to the user's workspace, logic similar to this PR is needed. I can create an Issue over at `prettier-plugin-svelte` so we can discuss pros/cons in more detail if you like.

PR #61, opened 2020-05-06
Fix import completion and add better completion documentaion

jasonlyu123:
feature:
1. Previously, completion for import statement only show file, now it would show directory and file with the extension (so that it'll have icon before it).
圖片
2. Add auto-import capability, it'll now show a message suggesting completion item would be auto imported
圖片
3. Show documentation of completion item
圖片

code structure:
1. Add `resolveCompletion` to `CompletionsProvider`
2. type `CompletionItem`'s data property

### DIFF
--- a/packages/language-server/src/plugins/typescript/TypeScriptPlugin.ts
+++ b/packages/language-server/src/plugins/typescript/TypeScriptPlugin.ts
@@ -195,13 +195,20 @@ export class TypeScriptPlugin
         // The language service throws an error if the character is not a valid trigger character.
         // Also, the completions are worse.
         // Therefore, only use the characters the typescript compiler treats as valid.
-        const validTriggerCharacter = ['.', '"', "'", '`', '/', '@', '<', '#'].includes(
-            triggerCharacter!,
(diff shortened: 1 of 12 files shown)


dummdidumm:
This is awesome!

Just one thing: Could you move the completion-logic into a new file (same for the tests then)? `plugins/typescript/features/getCompletions` or something like that, similar to how its done in `SveltePlugin`. `TypescriptPlugin` gets too big with all these great features 😄

packages/language-server/src/plugins/typescript/TypeScriptPlugin.ts
```diff
@@ -485,6 +337,37 @@ export class TypeScriptPlugin
...
+
+    private featureEnabled(feature: keyof LSTypescriptConfig) {
+        return (
+            this.configManager.enabled('typescript.enable') &&
+            this.configManager.enabled(`typescript.${feature}.enable`)
+        );
+    }
+}
+
+export class LSAndTSDocResovler {
```
dummdidumm, line 351: Could you move this class to its own file? This would resolve the circular dependency between `TypescriptPlugin` and `CompletionProvider`.

jasonlyu123:
ok it's ready

PR #66, opened 2020-05-07
(feat) hover info for svelte tags

dummdidumm:
hover info for tags such a #if, :else, :then, ...


jasonlyu123:
A little suggestion. Adding anchor links to offical doc would be nice

dummdidumm:
Good idea, will do!

Issue #63, opened 2020-05-07
Some questions


benmccann:
Hi, this looks like a really interesting project!

Does this project handle TypeScript in `.svelte` files currently or just JavaScript at the moment?

If it does handle TypeScript, can it run the TypeScript compiler and output JavaScript on to the file system or is it only meant to be used in Visual Studio?

If it's not yet supported, is handling TypeScript a goal someday and what would be required to compile TypeScript and output it to the file system? I potentially could help in this direction depending what's required. Most likely through testing and providing feedback, but maybe also writing tests or helping with some implementation details depending how much time I end up having

Thanks!

jasonlyu123:
From what I understand
1. It doesn't fully support typescript yet, users would still need to setup preprocessor for the moment. 
2. It's not a build tool but a dev tool. The language server is meant to be used by clients that support language-server-protocol like vs code. 
The building still depends on the community preprocessor and the svelte compiler. But I think it would nice to have an official typescript build tool that can emit declaration files. 

The building step is like this:
1. Svelte detect what should be preprocessed and preprocess with provided preprocessor
2. Svelte compile with transpiled js. 

So as the current build steps. separate build and type check seems more reasonable. Based on this approach what I imagine an integrate ts build tool would be:
1. have a preprocessor to transpile ts in the script to js
2. analysis and type check script and template
3. emit declaration files

dummdidumm:
@jasonlyu123 is right.
The Svelte Plugin will always just be a IDE Dev Tool to provide Intellisense for Svelte files. Right now Typescript is supported, but the support is still in early stages, so expect some inconveniences. But we are actively working on making the experience better!
Transpiling Typescript to Javascript and outputting this to the disk is done through preprocessors like https://github.com/kaisermann/svelte-preprocess .

Does this answer your question?

benmccann:
Thank you!

Will this tool need to do the same preprocessing as a build tool would need to? I'm wondering if that part would be a shared library that would be collaborated on between this project and a build tool effort

dummdidumm:
Yes, it needs to do the same preprocessing, but the preprocessors used are those defined by the user, which should be the same ones he also uses to build his project. The preprocessors are not part of this repo.

benmccann:
Oh, that's super interesting and not what I expected as someone with little experience with language servers. Do you use source map output from the pre-processor then to figure out how to mark up issues in the IDE?

[...]

PR #67, opened 2020-05-07
Add note to README about needing a preprocessor

benmccann:
Thanks for providing this info in https://github.com/sveltejs/language-tools/issues/63. Do you think it's appropriate to add it to the README to help the next person?

I also cleaned up a couple other things I noticed like inconsistent ways of saying VSCode, a run-on sentence, and a missing word

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,14 +86,13 @@ There are two ways to work on this project: either by working against an existin
 
 ## Running the Dev Language Server Against Your App
 
-To run the developer version of both the language server and the vscode extension:
+To run the developer version of both the language server and the VSCode extension:
 
-- open the root of this repo in VS Code
+- open the root of this repo in VSCode
 - Go to the debugging panel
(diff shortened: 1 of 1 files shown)


dummdidumm:
The main `README` is general information about this repository. Information about preprocessors is actually already present in the [Svelte VSCode README](https://github.com/sveltejs/language-tools/tree/master/packages/svelte-vscode). Still I think this is a good addition.

I just have one small suggestion: You don't need a preprocessor if you only use javascript/css. So a better wording might be "If you don't use pure Javascript and CSS, but languages like Typescript or SCSS, your project will need a Svelte preprocessor setup such as [svelte-preprocess](https://github.com/kaisermann/svelte-preprocess)."

dummdidumm:
Yeah I think this would be even better, to add a short link to these docs instead of to the docs of the `svelte-preprocess`-repo.

orta:
👍

Issue #68, opened 2020-05-07
CLI for validating project


orta:
**Is your feature request related to a problem? Please describe.**

Add a new package to the mono-repo which could be something like `svelte-check`. Perhaps being extra verbose is good here, because I'm not certain that in the long term this module would be the place where it lives. (e.g. I think if there's a `svelte` cli, then this would be a subcommand in there.

**Describe the solution you'd like**

The package would use the LSP shipped in this repo to ping every Svelte file in the current directory, and to request diagnostics (a.k.a errors) for that file. This is effectively the equivalent of running TypeScript and/or other tools we support in the LSP over each file.

**Describe alternatives you've considered**
In theory this could be built without the LSP, but that abstraction is solid and already has [proven itself in vetur](https://github.com/vuejs/vetur/blob/master/vti/src/cli.ts)

**Additional context**
Came out of https://github.com/sveltejs/svelte/issues/4518#issuecomment-625345746

benmccann:
Thanks @orta this is super helpful. So this would basically have type checking decoupled from building if I understand correctly. That gives you multiple ways to set things up, but one setup might look like:

* Setup a pre-processor (unless the functionality is merged into core https://github.com/sveltejs/svelte/issues/4797)
* Tell TypeScript to emit JS, but don't do type checking for faster building. That makes hot module reloading work faster in the browser and then you can rely on the IDE for errors
* If you're using a CI server, then you can run `svelte check` to fail the build if there are errors. You could also set this up to happen locally either while building or as a git precommit hook if you wanted

orta:
I don't know enough about how people use Svelte to talk about that middle step, in my head all of that was happening at IDE level 

But yeah, that's basically the entire gist 👍

halfnelson:
For some code to cut and paste or to use in the short term I had made one using svelte2tsx. https://github.com/halfnelson/svelte-type-checker maybe it can be ported to use language server

benmccann:
It looks to me like the [vetur implementation](https://github.com/vuejs/vetur/blob/master/vti/src/cli.ts) could be copy-pasted and used almost directly for svelte. We'd just need to tell it to look at `.svelte` files instead of `.vue` files and set the `languageId` and `version` appropriately.

[...]

PR #69, opened 2020-05-08
Set the languageId to svelte

benmccann:
I'm guessing the `languageId` shouldn't be empty and was only set that way to get things off the ground

https://code.visualstudio.com/docs/languages/identifiers suggests it should match the package.json:
https://github.com/sveltejs/language-tools/blob/98fc363d079a095b54c8d5c304cdbc6cb62c9d76/packages/svelte-vscode/package.json#L197

This is the first step of https://github.com/sveltejs/language-tools/issues/68

### DIFF
--- a/packages/language-server/src/lib/documents/Document.ts
+++ b/packages/language-server/src/lib/documents/Document.ts
@@ -118,7 +118,7 @@ export abstract class Document implements TextDocument {
         return this.getText().split(/\r?\n/).length;
     }
 
-    languageId = '';
+    abstract languageId: string;
 }
 
 /**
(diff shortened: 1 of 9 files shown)


packages/language-server/src/lib/documents/Document.ts
```diff
@@ -118,7 +118,7 @@ export abstract class Document implements TextDocument {
         return this.getText().split(/\r?\n/).length;
     }
 
-    languageId = '';
+    languageId = 'svelte';
```
jasonlyu123, line 121: This is actually the base class of multiple language's sub-document. Could you also help checking does setting this makes any difference?

benmccann: Fixed.

I don't think it actually matters except in the case where a file extension can possibly hold multiple types of content and you need to differentiate. Still it seems like good practice to set it appropriately

dummdidumm:
I honestly don't know if any of the plugin's language services actually uses that `documentId`. But these changes shouldn't hurt. Thanks!

PR #73, opened 2020-05-08
remove end brackets from completion trigger character

jasonlyu123:
The `)` and `]` are listed as completion trigger characters for language-server. But this causes ts to give competition at weird places.

for example:
```ts
Promise.resolve()
``` 
This end parentheses would trigger completion and ts would give back global objects as completion then if you type `.`, the completion will be committed causing unwanted completion.
圖片
圖片

The comment said it's for emmet, but I don't know any emmet completion that should be trigger by these two. So I think we could safely remove both of them. If anyone knows any, you could point out for me. I would change it be only filter out by the ts completion.

### DIFF
--- a/packages/language-server/src/server.ts
+++ b/packages/language-server/src/server.ts
@@ -66,9 +66,7 @@ export function startServer() {
                         '+',
                         '^',
                         '(',
-                        ')',
                         '[',
-                        ']',
                         '@',
                         '-',
 


dummdidumm:
According to [this doc](https://docs.emmet.io/abbreviations/syntax/) `)` and `]` are closing tags of some valid syntax. So maybe we play it safe and filter them out in the ts completion instead, like you proposed.

dummdidumm:
Thought about it some more. These characters represent the end of something. In that sense, they don't trigger something new, so I think the change is fine like this.

Issue #74, opened 2020-05-08
WebStorm / IntelliJ support?


shirotech:
It seems like there is a plugin for LSP https://plugins.jetbrains.com/plugin/10209-lsp-support
https://github.com/gtache/intellij-lsp

I tried running the executable `/usr/local/bin/svelteserver` but nothing happens, can anyone point me to the right direction? Thanks.

<img width="1091" alt="Screen Shot 2020-05-09 at 2 37 10 am" src="https://user-images.githubusercontent.com/1149825/81427337-069c7a80-919e-11ea-98a4-013336ea142c.png">

![Screen Shot 2020-05-09 at 2 48 00 am](https://user-images.githubusercontent.com/1149825/81429164-bf63b900-91a0-11ea-8076-5047c9b7736f.png)

dummdidumm:
I honestly don't know if the IntelliJ plugin even uses this language-server right here. Since this is a question related to that plugin, please ask the question over there.

shirotech:
I thought all LSP are standard, but anyways thanks for the quick response, I'll try asking over there. Cheers.

dummdidumm:
Maybe related #34

jasonlyu123:
I also tried it on vetur's vls. same result.

benmccann:
> I thought all LSP are standard

I'm not sure, but the code seems to refer to VSCode specifically:

https://github.com/sveltejs/language-tools/blob/a14cf1b61b5958159e342f96a391b9ac94eb5aa1/packages/language-server/src/server.ts#L9

It's pretty confusing that the repo is generically named "language-tools" instead of something like "vscode". I don't know if there are plans for other plugins to live in this repo? If not, it might be worth renaming this one. I would think each should have their own repo.

benmccann:
Actually, it does look like most IDEs support LSP. IDEA IntelliJ / WebStorm don't, but most others including Atom do.

https://microsoft.github.io/language-server-protocol/implementors/tools/

orta:
This should conform to the standard LSP interface (so it should work), though I'm afraid I have no advice on how to run the module as a standalone CLI tool. I don't know, I'd recommend digging into the source and seeing how it works.

dummdidumm:
[From this comment](https://github.com/sveltejs/language-tools/issues/361#issuecomment-671846219) it seems that it will work if you add the `stdio` flag.

dummdidumm:
Closing in favor of #361

PR #78, opened 2020-05-10
Add a link to production version of VSCode plugin

benmccann:
It's hard to make your way from this repo to the plugin. We should make it easier to navigate by adding a link.

I originally was going to put the link in https://github.com/sveltejs/language-tools/blob/master/packages/svelte-vscode/README.md in https://github.com/sveltejs/language-tools/pull/76 but it appears that README is used directly in https://marketplace.visualstudio.com/items?itemName=svelte.svelte-vscode so then I decided that wasn't the best place to put it

I also am clarifying in this PR that this repo is for a VSCode language server and that language servers aren't generic across IDEs. This was confusing to me and others. E.g. see https://github.com/sveltejs/language-tools/issues/74#issuecomment-625913226

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,8 @@
 
 ## What is Svelte Language Tools?
 
+This repository contains the code that powers the [VSCode extension](https://marketplace.visualstudio.com/items?itemName=svelte.svelte-vscode).
+
 A `.svelte` file would look something like this:
 
 ```html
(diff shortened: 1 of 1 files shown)


jasonlyu123:
I think the rationale behind 
>tools which provide editor integration

is more like 'it suppose to support, if it ain't we should found the reason and fix it'

benmccann:
The thing that confused me was the the majority of files refer to VSCode packages, so I though it'd only support VSCode. I guess the naming for the LSP packages isn't as generic as it should be. I updated the description to add more details about this and revert some of the earlier changes.

If this gets merged you'll probably want to "Squash and merge" since there are multiple commits on the PR.

dummdidumm:
Good clarification, thanks!

Issue #42, opened 2020-04-27
Only extract top level script


dummdidumm:
Currently, script tags are extracted no matter where they are within the svelte component. But only the top-level script should be treated as the instance script. 

Test cases:

```html
{#if name}

   <script>
	console.log('I should not be treated as top level')
   </script>

{/if}

<script>
    let name = 'I am top level!';
</script>
```

```html
{@html `<script type="application/ld+json">${JSON.stringify(iAmNotTopLevel)}</script>`}
```

Extractor is in https://github.com/sveltejs/language-tools/blob/master/packages/language-server/src/lib/documents/utils.ts

Also see https://github.com/UnwrittenFun/svelte-language-server/issues/4

dummdidumm:
Closed via #81

PR #81, opened 2020-05-12
Only extract top level script

sjafferi:
Fix for: https://github.com/sveltejs/language-tools/issues/42

This PR changes `extractTag` such that `{#if xx }` statements are ignored. 

But I'm not sure all of these statements should be ignored (for example, `{#if true}`). I'm also not sure if this covers all the possibilities where a script tag should not be considered top level. 

Let me know what you guys think.

### DIFF
--- a/packages/language-server/test/lib/documents/utils.test.ts
+++ b/packages/language-server/test/lib/documents/utils.test.ts
@@ -90,45 +90,47 @@ describe('document/utils', () => {
         });
         it('extracts top level script tag only', () => {
             const text = `
-            {#if name}
-                <script>
-                    console.log('not top level')
-                </script>
-            {/if}
-            <ul>
-                {#each cats as cat}
+                {#if name}
                     <script>
                         console.log('not top level')
                     </script>
-                {/each}
-            </ul>
-            {#await promise}
-                <script>
-                    console.log('not top level')
-                </script>
-            {:then number}
-                <script>
-                    console.log('not top level')
-                </script>
-            {:catch error}
-                <script>
-                    console.log('not top level')
-                </script>
-            {/await}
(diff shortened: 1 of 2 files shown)


dummdidumm:
This looks good, thanks! Could you add the same logic for `#each`, `#await` and `@html`? I think then all use cases where an inner script could occur should be covered.

packages/language-server/test/lib/documents/utils.test.ts
```diff
@@ -74,5 +74,27 @@ describe('document/utils', () => {
...
+            <script>top level script</script>
+            `;
+            assert.deepStrictEqual(extractTag(text, 'script'), {
+                content: 'top level script',
+                attributes: {},
+                start: 359,
+                end: 375,
+                container: { start: 351, end: 384 },
+            });
+        });
```
orta, line 98: 👍

sjafferi:
> This looks good, thanks! Could you add the same logic for `#each`, `#await` and `@html`? I think then all use cases where an inner script could occur should be covered.

Awesome, added these cases in.

packages/language-server/test/lib/documents/utils.test.ts
```diff
@@ -88,5 +88,50 @@ describe('document/utils', () => {
                 container: { start: 17, end: 73 },
             });
         });
+        it('extracts top level script tag only', () => {
+            const text = `
+                {#if name}
```
sjafferi, line 93: wondering if i should split these up into separate tests...

dummdidumm:
Great work! I don't mind the test being that big. If there are ever bugs in that area we can always split them up to narrow down the error.

PR #82, opened 2020-05-12
fix #80, filtering completion trigger character

### LINKED ISSUE #80: Weird autocompletion behaviour
**Describe the bug**
Some weird autocompletion happens, especially when using ( )
```
<script>
// I import some objects/components/functions...
import { Button } from '@UI';

...

// Then for example I write an `if`
// and the autocompletion becomes weird
if Button() // <- I have the Button that is put even if I didnt select it
```
for example. There seems to be a lot of weird autocompletion going on. Above case is from someone else on the discord but I started experiences such weird autocompletion while writing code yesterday. Will update the issue when I track down more cases when it happens.

**Expected behavior**
Correct autocompletion like in Svelte plugin by James Birtle, had no issues there.

 - OS: Windows
 - IDE: Vscode
 - Plugin: Svelte Beta

jasonlyu123:
1. `getCompletions` now use CompletionContext instead of just `triggerCharacter`
2. TS and CSS now only allow specific trigger characters and will not try to get global completion if completion is invoked by a trigger character in which specified in server capability(https://github.com/sveltejs/language-tools/blob/4a8bdd7f9fb04ba6e11fd1d6a87120fb0d9f4fb9/packages/language-server/src/server.ts#L51)

### DIFF
--- a/packages/language-server/src/plugins/PluginHost.ts
+++ b/packages/language-server/src/plugins/PluginHost.ts
@@ -16,6 +16,7 @@ import {
     TextEdit,
     FileChangeType,
     CompletionItem,
+    CompletionContext,
 } from 'vscode-languageserver';
 import { LSConfig, LSConfigManager } from '../ls-config';
 import { DocumentManager } from '../lib/documents';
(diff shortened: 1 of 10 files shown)


dummdidumm:
Great job!

Issue #86, opened 2020-05-13
Get more robust/user friendly with preprocessors


dummdidumm:
**Is your feature request related to a problem? Please describe.**
Many people seem to have problems setting up preprocessors for the plugin - I guess partly because they install the extension and just expect it to work.

**Describe the solution you'd like**
Maybe we can get more robust/user friendly by doing the following things:
- If we don't detect a `svelte.config.js`, we just do a best guess by using https://github.com/kaisermann/svelte-preprocess
- If we detect that inside javascript "unexpected token"-errors occur, we enhance the error message with something like "have typed `lang="typescript"` in script? have you setup preprocessors correctly? see \<somelink\> for more info"

**Describe alternatives you've considered**
Make docs clearer.

orta:
These are good calls 👍

jasonlyu123:
Does the 'do a best guess' means:
1. adding `svelte-preprocess` as a dependency of language-server and make it the default preprocessor 

or

2. only try to import it from the user's workspace?

dummdidumm:
It would mean adding it as a dependency and using it as the fallback preprocessor.

dummdidumm:
As already stated in #87,  @jasonlyu123's second option could also be a good solution. Would save us from dependencies.

benmccann:
I think just detecting if there are no preprocessors setup and telling them to add one with a helpful error message would be pretty helpful. It'd be a bit less magical and more explicit too, which I always think is nice

PR #87, opened 2020-05-14
(feat) add svelte-preprocess as fallback

dummdidumm:
#86 

No preprocessor apart from ts supported yet

We have to discuss what to support out of the box and what not, because the preprocessor (rightfully) says it is your responsibility to install `less` / `babel` etc. So if we would support all of them, we would add quite a lot of dependencies to our project and also bloat the package size.

So what would be the subset we support?

For me mandatory:
- `typescript` (already supported)

Optional:
- `babel`
- `less`
- `sass` (does that work? There are some very system specific things going on with that package..)

As an alternative, we could [maybe do some hacks](https://gist.github.com/branneman/8048520) to reroute the module resolution of `svelte-preprocess` to the user workspace. Chances are very high the user would have installed the needed preprocessors because he needs them either way to get the project to compile.

Another alternative would be to not add this dependency to our project but instead try to find the dependency inside the user's workspace. I think most users use `svelte-preprocess` in their webpack/rollup config, so there is a high chance they have it (and all required other dependencies) installed.

Opinions?

Side note: I added `skipLibCheck` to `tsConfig` because `svelte-preprocess` fails the strict type checks, and also because of the peer dependency types. I hope that is okay for you?

### DIFF
--- a/packages/language-server/package.json
+++ b/packages/language-server/package.json
@@ -55,6 +55,7 @@
         "source-map": "^0.7.3",
         "svelte": "3.19.2",
         "svelte2tsx": "~0.1.4",
+        "svelte-preprocess": "~3.7.4",
         "typescript": "*",
         "vscode-css-languageservice": "4.1.0",
         "vscode-emmet-helper": "1.2.17",
--- a/packages/language-server/src/lib/documents/utils.ts
+++ b/packages/language-server/src/lib/documents/utils.ts
@@ -1,5 +1,5 @@
-import { clamp } from '../../utils';
-import { Position } from 'vscode-languageserver';
+import { clamp, isInRange } from '../../utils';
+import { Position, Range } from 'vscode-languageserver';
 
 export interface TagInformation {
     content: string;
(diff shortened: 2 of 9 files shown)


jasonlyu123:
I think we don't need babel. If the syntax can't be handle by typescript, it probably is some tc39 proposal syntax that needed to be installed separately as a plugin by the user.

if we don't include scss and less. the style features would still work. but the compiler will throw parsing error. That can be suppressed by preprocessing everything in the style tag to white space. In that way, the 'unused-CSS' warning would be sacrificed

dummdidumm:
Babel: good point.

If we don't include less/scss then ... : Good idea. Additionally we could then add a warning diagnostic like "you need to add a `svelte.config.js` to get style svelte diagnostics".

jasonlyu123:
Supplement my thought on babel. If the user is using js syntax that is not supported by svelte, for example, optional chaining, We probably can hack `svelte-preprocessor` to transpile as if it's typescript.

dummdidumm:
I now added a "try finding `svelte-preprocess`-fallback and enhanced the error messages with some suggestions on possible error causes. I did not add "try parsing babel as typescript" or "blank out style and try again" logic. This can be done in a different PR, if we feel it's necessary.

Before I merge this in: Can someone please read the enhanced error messages and tell me if they are concise/understandable?

dummdidumm:
No objections it seems, so let's try it with this one.

PR #90, opened 2020-05-16
(fix) don't trigger html completions inside moustache tags

### LINKED ISSUE #89: Autocompletion inside on:click arrowfunction
**Describe the bug**
When making event handler (i.e. on:click), the arrow function autocompletion inside seems to be malfunctioning.

**To Reproduce**
Try typing:
```
<div class="answer" on:click={() =}
```
when you do `>` it autocomplete the div inside the arrow function, see gif below:

**Expected behavior**
No autocompletion, let me do the arrow function.

**Screenshots**

[Link to imgur gif](https://imgur.com/a/zhgBEKo)

 - OS: Windows
 - IDE: VSCode
 - Plugin: Svelte Beta

dummdidumm:
Fixes #89

@jasonlyu123 does this seem like a reasonable fix for you?

### DIFF
--- a/packages/language-server/src/plugins/html/HTMLPlugin.ts
+++ b/packages/language-server/src/plugins/html/HTMLPlugin.ts
@@ -14,7 +14,7 @@ export class HTMLPlugin implements OnRegister, HoverProvider, CompletionsProvide
 
     onRegister(docManager: DocumentManager, configManager: LSConfigManager) {
         this.configManager = configManager;
-        docManager.on('documentChange', document => {
+        docManager.on('documentChange', (document) => {
             const html = this.lang.parseHTMLDocument(document);
             this.documents.set(document, html);
         });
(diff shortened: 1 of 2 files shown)


jasonlyu123:
I think it could be moved to `HTMLPlugin`'s `doTagComplete`. Also, there's also another thing that could trigger this tag close.
```
<input type="checkbox" checked={ a > b } />
```

Maybe using a solution similar to what you did to find last open tag to find if there is any unended  `{` before `>` . And just play it safe and disabling the auto close even if `{` or  `}` is possibly in a string literal.

dummdidumm:
You are right with both things. Will do, thanks!

jasonlyu123:
Find an advantage of doing it in the `HtmlPlugin`. we could utilize HTML language service's AST to find where current node start
```
const node = html.findNodeAt(offsetAt(position, document.getText()));
```

dummdidumm:
Good find! I will try to add it to the logic.

dummdidumm:
I did add your suggestions, works really well. Thanks again!

PR #98, opened 2020-05-20
(feat) prepare for svelte-check

dummdidumm:
- add options to turn off logging and to pass connection
- export a util used in svelte-check

Part of #68 

After this is merged and the npm package is updated, a PR with a new package Svelte-Check will land.


orta:
very cool

PR #102, opened 2020-05-21
Migrates svelte2tsx into the monorepo

orta:
This retains all history, so it'll look a bit messy - everything until https://github.com/sveltejs/language-tools/commit/b3575a55643878f7fa89fe15354220b0c8116bbe is the import

Anything afterwards is getting it passed the linters etc


orta:
Only thing like is that I have to think about deploys - TBH, I am at a point where I'll probably build a quick tool for doing deployments, I've got these same scripts in a few different repos now and they all share a lot of code

README.md
```diff
@@ -59,6 +59,10 @@ The language server for Svelte. Built from [UnwrittenFun/svelte-language-server]
 
 The official vscode extension for Svelte. Built from [UnwrittenFun/svelte-vscode](https://github.com/UnwrittenFun/svelte-vscode) to become the official vscode extension for the language.
 
+#### [`svelte2tsx`](packages/svelte2tsx)
+
+Converts a .svelte file into a legal TypeScript file. Built from [halfnelson/svelte2tsx](https://github.com/UnwrittenFun/halfnelson/svelte2tsx) to provide the auto-complete and import mapping inside the language server.
```
halfnelson, line 64: url should be https://github.com/halfnelson/svelte2tsx
I wouldn't bother with the attribution, the old repo will be archived.

orta:
I also lean towards yes on importing it that way 👍 

Good point, yeah, I can run prettier over everything

PR #113, opened 2020-05-26
(feat) update imports for renames/moves files

### LINKED ISSUE #111: Support rename/move of Svelte files
Right now renaming/moving svelte files does not update the imports in other svelte files.

Renaming/moving svelte files should update the imports in other svelte files.

**Additional context**
I dug around the language server protocol a little but found no event that says "file X moved from A to B" or "file X was renamed to Y". Instead, `onDidCloseTextDocument` and `onDidOpenTextDocument` events fire, and also the `onCodeAction` event is invoked with the new file path and a range with start/end of 0. One could infer a file rename/move from these events, if they appear shortly after each other, but I'm not sure if this is a hack or the only possible solution. If someone knows more about this, please comment.

dummdidumm:
closes #111

~~This works, but it's really slow. It takes 30 seconds minimum to update the paths. The time is entirely spent in the `getEditsForFileRename` of the ts language service. Is this always this slow or am I doing something wrong?~~ Update: Bumping ts to version 3.9 did the trick. Luckily for us, the ts team did speed up that exact area drastically.

I also did manage to speed up the tests quite a bit by adding a `tsconfig.json` to the testfiles folder. Before that the ts services inside the tests did know about the whole project with all its ts files, which lead to the slow tests.


jasonlyu123:
I notice an issue. For some reason, sometime it would apply the edit to the old path. so vscode would open a new tab with unsaved changes on a deleted file.

dummdidumm:
Good catch! I somehow have to tell the language service to update its document beforehand so he knows the old one no longer exists.

PR #119, opened 2020-05-27
Also processes markup in transpile.

MaartenX:
Like how the Svelte compiler works, we should first process the markup,
before we can process the script and style, to make sure we get a
representative result.

This fixes an issue where the TypeScript preprocessor needs to look at 
the markup to update the code 
(see this PR: https://github.com/sveltejs/svelte-preprocess/pull/155).

Since the mentioned PR is not merged yet, I did not create an issue,  but
I think the language server should expose the same behavior as the 
Svelte compiler.


dummdidumm:
Thanks!

PR #120, opened 2020-05-27
All props ($$props) issue

### LINKED ISSUE #104: Variable names that begin with $$ results in a problem message
Variable names that begin with `$$` (`$$props` for example) results in a message "Type annotations can only be used in TypeScript files."

**Screenshots**
изображение

**System:**
 - OS: Kubuntu 20.04
 - IDE: VSCode
 - Plugin: Svelte Beta

jasonlyu123:
fix #104 

Move the type annotation of `$$props` to a dummy declare function and refer type from function return type. So that there will not be any typescript annotation in emitted `tsx`.

Also, add support for `$$restProps`. I didn't create another flag for it because these two both provide access to undefined props, I think it can be simply checked at the same step.

### DIFF
--- a/packages/svelte2tsx/src/svelte2tsx.ts
+++ b/packages/svelte2tsx/src/svelte2tsx.ts
@@ -543,7 +543,7 @@ function createRenderFunction(
     uses$$props: boolean,
 ) {
     const htmlx = str.original;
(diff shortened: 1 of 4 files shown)


dummdidumm:
Nice!

Issue #124, opened 2020-05-28
svelte-check: make output machine-readable


rondonjon:
**Is your feature request related to a problem? Please describe.**

The current output of `svelte-check` is human-friendly (e.g. colors, source code snippets) and verbose but at the same time very machine-unfriendly. There is no specification of the output format in the README / manpage / `--help`, and the included source code extracts could be challenging for a parser.

**Describe the solution you'd like**

Please make the output more configurable through CLI parameters, e.g.:

 * a `--ci` parameter for: less verbosity, no colors, and a clearly specified line format, e.g.:
```
    FILE <path>
    <ERROR|WARNING> <line>:<column> "<description (no linebreaks, quotes escaped with backslash)>"
   ...
```
 * a `--verbose` flag for the *current* behaviour

**Describe alternatives you've considered**

In comparison, parsing the current output seems very circumstantial.

rondonjon:
Find pull request [here](https://github.com/sveltejs/language-tools/pull/126)

orta:
Fixed in #126

PR #126, opened 2020-05-28
Refactor console output & add machine-readable output option

rondonjon:
Patch for enhancement #124

### DIFF
--- a/packages/svelte-check/src/writers.ts
+++ b/packages/svelte-check/src/writers.ts
@@ -32,8 +32,7 @@ export class HumanFriendlyWriter implements Writer {
 
     file(diagnostics: Diagnostic[], filename: string, text: string): void {
         if (diagnostics.length > 0) {
-            this.stream.write('\n');
-            this.stream.write(`${chalk.green('File')} : ${chalk.green(filename)}\n`);
+            this.stream.write(`${chalk.green('File')}: ${chalk.green(filename)}\n`);
 
             diagnostics.forEach((diagnostic) => {
                 const source = diagnostic.source ? `(${diagnostic.source})` : '';
--- a/packages/svelte-check/src/writers.ts
+++ b/packages/svelte-check/src/writers.ts
@@ -1,4 +1,5 @@
 import * as chalk from 'chalk';
+import { sep } from 'path';
 import { Writable } from "stream";
 import {
     Diagnostic,
@@ -7,8 +8,8 @@ import {
 import { offsetAt } from 'svelte-language-server';
 
 export interface Writer {
-    start: (workspaceUri: string) => void;
-    file: (d: Diagnostic[], filename: string, text: string) => void;
+    start: (workspaceDir: string) => void;
+    file: (d: Diagnostic[], workspaceDir: string, filename: string, text: string) => void;
     completion: (fileCount: number, errorCount: number, warningCount: number) => void;
     failure: (err: Error) => void;
 }
@@ -17,10 +18,10 @@ export class HumanFriendlyWriter implements Writer {
     constructor(private stream: Writable, private isVerbose = true) {
     }
 
-    start(workspaceUri: string) {
+    start(workspaceDir: string) {
         if (this.isVerbose) {
             this.stream.write('\n');
-            this.stream.write(`Loading svelte-check in workspace: ${workspaceUri}`);
+            this.stream.write(`Loading svelte-check in workspace: ${workspaceDir}`);
             this.stream.write('\n');
             this.stream.write('Getting Svelte diagnostics...\n');
             this.stream.write('====================================\n');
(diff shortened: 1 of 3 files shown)


packages/svelte-check/src/writers.ts
```diff
@@ -0,0 +1,124 @@
...
+                const codeHighlight = chalk.magenta(text.substring(startOffset, endOffset));
+                const codePost = chalk.cyan(text.substring(endOffset, endOffset + 10));
+                const code = codePrev + codeHighlight + codePost;
+                let msg;
+
+                if (this._isVerbose) {
+                    msg = `${diagnostic.message} ${source}\n${position}\n${chalk.cyan(code)}`;
+                }
+                else {
+                    msg = diagnostic.message;
```
dummdidumm, line 63: I think line and character should also be shown in the non-verbose output.

packages/svelte-check/src/writers.ts
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    file(diagnostics: Diagnostic[], filename: string, _text: string) {
+        diagnostics.forEach((d) => {
+            const { message, severity, range } = d;
+            const type = severity === DiagnosticSeverity.Error ? "ERROR" : severity === DiagnosticSeverity.Warning ? "WARNING" : null;
+
+            if (type) {
+                const { line, character } = range.start;
+                this._log(`${type} ${JSON.stringify(filename)} ${line}:${character} ${JSON.stringify(message)}`);
```
dummdidumm: Just curious: Why do you use `JSON.stringify` here? If it's to have `"` around the content, maybe just do `"${filename}"` instead.

packages/svelte-check/src/writers.ts
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    completion(fileCount: number, errorCount: number, warningCount: number) {
+        this._log(`COMPLETED ${fileCount} FILES ${errorCount} ERRORS ${warningCount} WARNINGS`);
+    }
+
+    failure(err: Error) {
+        this._log(`FAILURE ${JSON.stringify(err.message)}`);
+    }
+}
```
dummdidumm, line 132: Very nice!
To remove the lint errors, just do `// eslint-disable-next-line max-len`.
Also minor code style thing: Could you refactor it so that it's not `_<variable>`, instead just `<variable>` and in the constructor do `constructor(private <variable>: <type>) {}`

packages/svelte-check/src/writers.ts
rondonjon: That's to escape special characters, e.g. quotes

packages/svelte-check/src/writers.ts
rondonjon: done.

rondonjon:
> Looks good, thank you! Only have some minor things that should be addressed, see comments in code.
> 
> Could you also update the README to tell people about the new options?

done.

packages/svelte-check/src/writers.ts
rondonjon: done.

packages/svelte-check/src/writers.ts
```diff
@@ -14,29 +14,26 @@ export interface Writer {
...
 
 export class HumanFriendlyWriter implements Writer {
-    _stream: Writable;
-    _isVerbose: boolean;
-
-    constructor(stream: Writable, isVerbose = true) {
-        this._stream = stream;
-        this._isVerbose = isVerbose;
+    constructor(private stream: Writable, private isVerbose = true) {
+        this.stream = stream;
```
dummdidumm, line 18: `this.X = X` can be omitted, ts will do that for us 😃 
So it just becomes `constructor(private stream: Writable, private isVerbose = true) {}`

rondonjon: done with previous commit.

dummdidumm: Not quite, it's 

```
constructor(private stream: Writable, private isVerbose = true) {
        this.stream = stream;
        this.isVerbose = isVerbose;
    }
```

but can be 
`constructor(private stream: Writable, private isVerbose = true) {} // <- empty constructor body`

Or github is showing me something wrong here 😄

rondonjon: Ah, sorry, misunderstood your original request. Done now.

rondonjon:
@dummdidumm would you mind if I changed this

```
File: importcompletions6.svelte
Error: Unexpected end of input (svelte):Line: 3, Character: 7
```

to this?

```
File: importcompletions6.svelte
Error: [3:7] Unexpected end of input (svelte)
```

dummdidumm:
Sounds good!

rondonjon:
> Sounds good!

OK, cool. Just added that.

rondonjon:
@dummdidumm 
No more changes intended from my side. Feel free to merge or take over ...

dummdidumm:
The idea with IDE-Links is very cool. However for me it does only work if I have the absolute file path in there. Is that different for you?

rondonjon:
> The idea with IDE-Links is very cool. However for me it does only work if I have the absolute file path in there. Is that different for you?

It works for me with relative paths, **if** the current folder in the console is also used as the workspace folder in svelte-check, i.e. without `--workspace` in the command line

dummdidumm:
For me it only works if the location where the command is run and the location where my IDE is open are the same (using VSCode).
What do you think about the absolute file path? Would that be too much noise for you?

rondonjon:
> For me it only works if the location where the command is run and the location where my IDE is open are the same (using VSCode).
> What do you think about the absolute file path? Would that be too much noise for you?

OK, please check

dummdidumm:
Great work!

orta:
Nice work

PR #147, opened 2020-06-01
(fix) supress used before assigned error

### LINKED ISSUE #146: svelte props, used before assigned msg
I'm not sure if this is bug or a known issue or i'm doing something wrong or perhaps expected behavior...but it seems like i have to give svelte props default values or it complains.

image

dummdidumm:
Also fix other type error which appears in strict mode when transforming event forwarding.

Fixes #146


jasonlyu123:
I am kind of against suppressing used-before-assigned error because it could indeed be undefined if no value is passed in from props. People using strict mode may want the type check to be as annoying as possible, so some of the users might want to be warned if props are possibly undefined.

dummdidumm:
Mhm that's true..

dummdidumm:
Added a new props construction like discussed in the issue.

PR #152, opened 2020-06-02
fix style, script bare attribute parsing

### LINKED ISSUE #136: Markup incorrectly parsed after <style global> SCSS flag
**Describe the bug**
HTML Svelte interpolated code with comparisons `{condition === 2}` starts to make the rest of the file to be wrongly parsed after <style> tag if 'global' flag is provided at the end of the style tag for SCSS preprocessors.
_This does not affect builds and the **same svelte.config is both used for rollup and for vscode with the preprocessor**._
**To Reproduce**
1. The order of the Svelte file has to be `script - style - markup`.
2. Style **has to use `global` preprocessor flag attribute after any other attrib** like `<style src="./path.scss" global>`
3. In the markup, inside interpolated Svelte code ({expressions}) there should be a boolean comparison (this is the only case I found but there might be more) like `{x === 2}`.

Minimal reproducible example:
```svelte
<script>let whatever = 1; /* it does not matter */</script>
<style src="./aFile.scss" global> // <-- 'global' has to be last
{whatever === 1} // <-- last '}' will give you error, then almost any other not native html code
```

**Expected behavior**
Nothing should be prompted unless it is a real error.

**Screenshots**
Screenshot 2020-05-29 at 13 37 47

 - OS: Mojave. Mac
 - IDE: VSCode
 - Plugin/Package: only "Svelte Beta" (No "Svelte" or "Svelte Type Check" or similar)

**WORKAROUND**
I found out that if `global` is set just before the `src=""` attribute, the compiler will work. 
This clarifies that this issue is, in fact, from the language-tool, perhaps as a conflict with other built-in plugins from VSCode

```svelte
<script>let whatever = 1; /* it does not matter */</script>
<style global src="./aFile.scss">
{whatever === 1} // This works
```

jasonlyu123:
fixes #136 
also fix typescript errors

### DIFF
--- a/packages/svelte2tsx/src/svelte2tsx.ts
+++ b/packages/svelte2tsx/src/svelte2tsx.ts
@@ -1,7 +1,7 @@
 import MagicString from 'magic-string';
 import { parseHtmlx } from './htmlxparser';
 import { convertHtmlxToJsx } from './htmlxtojsx';
-import { Node } from 'svelte/compiler';
+import { Node } from 'estree-walker';
 import * as ts from 'typescript';
 
 function AttributeValueAsJsExpression(htmlx: string, attr: Node): string {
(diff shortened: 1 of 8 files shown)


packages/svelte2tsx/src/svelte2tsx.ts
```diff
@@ -452,15 +452,18 @@ function processInstanceScriptContent(str: MagicString, script: Node): InstanceS
         }
 
         if (ts.isExportDeclaration(node)) {
-            for (const ne of node.exportClause.elements) {
-                if (ne.propertyName) {
-                    addExport(ne.propertyName, ne.name);
-                } else {
-                    addExport(ne.name);
+            const { exportClause } = node;
```
dummdidumm, line 455: Just curious: For what is this update to the code needed?

dummdidumm: Ah, nevermind, it's for getting rid of the type errors.

jasonlyu123: This just adds another check if the `exportClause` is a NamedExports so that typescript won't complain that `elements` doesn't exist on type `NamespaceExport`.

PR #159, opened 2020-06-04
(feat) bump to es2018

### LINKED ISSUE #155: Consider updating the Typescript target?
Currently the language server targets ES5: https://github.com/sveltejs/language-tools/blob/master/packages/language-server/tsconfig.json#L3

This includes all sorts of interesting code to deal with async/await. Can we assume that the editor is using a sufficiently modern node.js?

**Update** Sorry this should be a feature, not a bug.

dummdidumm:
Also add info that node 12 is required (although node 10 should also work)

Closes #155

### DIFF
--- a/packages/language-server/README.md
+++ b/packages/language-server/README.md
@@ -3,6 +3,8 @@
 A language server (implementing the [language server protocol](https://microsoft.github.io/language-server-protocol/))
 for Svelte.
 
+Requires Node 12 or later.
+
 ## What is a language server?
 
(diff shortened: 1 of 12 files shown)


dummdidumm:
Seems like the `svelte2tsx` setup does not honor the `extends` property of `tsconfig.json`. That's why some tests fail, because the created javascript code is really strange. I will not use the `extends` there, then. Ok for you @orta ?

orta:
Sure

PR #160, opened 2020-06-04
Integrate Svelte Atom

orta:
Once I read the code, I realized that this really is a simple LSP wrapper, and there's not much code here.

TODO: 
 - deploys


orta:
deployment is tricky, because it relates to the repo and having GitHub org access

orta:
OK, I might be able to deploy using the default GitHub action token - https://help.github.com/en/actions/configuring-and-managing-workflows/authenticating-with-the-github_token#permissions-for-the-github_token

orta:
I think with the production deploys PR in this is less unblocked in my mind, I'd want this to only make deploys when there are prod deploys of the extension

dummdidumm:
FYI I had to do some fixes in #239 to get lint/test running again. This will likely cause some merge conflicts for this PR.

orta:
Ah, hrm, that shouldn't be there - must have been an accidental push somewhere

orta:
When I looked at this again, I think I was blocked on the fact that we have a monorepo and GH makes some assumptions about your repo in the atom deployment process, I'll need to look again

the0neWhoKnocks:
Looks like it's been almost 2 weeks since the last activity on this PR. As an Atom user, just curious if this is getting abandoned.

orta:
It's just work I've not prioritized, that's all

orta:
An update, as. I gave it another shot over the weekend.

There's some assumptions with how GitHub treats atom extensions which requires having everything set at the root of the git directory even with scripting the files to be at the fs root then. I think we can skip the generation of tags and releases which are just for the extension but having the fs at the root is a pretty big blocker.

I'm going to move this PR to draft, it's blocked on monorepo support inside the atom publishing CLI [apm](https://github.com/atom/apm). 

I'm not really game for diving into a big coffeescript codebase to try and add monorepo support, so if this extension is something you want to see happen - now is a good chance to start contributing.

dummdidumm:
You are also welcome to copy the setup code (`main.js`, `package.json`, `readme.md`) and open a separate repo and deploy the extension yourself to work around the monorepo limitation.

dummdidumm:
Just noticed, there is also this official repo https://github.com/sveltejs/svelte-atom which we could use. @orta what do you think? If you think it's okay to have it in a separate repo for deployment reasons, mind taking another stab at this in that repo?

dummdidumm:
Closing in favor of https://github.com/sveltejs/svelte-atom/pull/10

Issue #161, opened 2020-06-04
Verify and document typescript + svelte-preprocessor working out of the box


orta:
We think you can make a svelte project and just include the node modules for svelte-preprocessor to get `lang='ts'` support working. Verify that works and add some notes 👍

dummdidumm:
Works like a charm https://github.com/dummdidumm/template

### Differences to the javascript version

- Following npm packages are added: `svelte-preprocess typescript tslib @rollup/plugin-typescript`
- `rollup.config.js` is enhanced:

```js
// ...
import sveltePreprocess from 'svelte-preprocess';
import typescript from '@rollup/plugin-typescript';

// ...
  plugins: [
    svelte({
      // ...
      preprocess: sveltePreprocess(), // <--
    }),

    // ...
    commonjs(),
    typescript(), // <-- added below commonjs
    // ...
```

### Caveats:

- entry file (`main.js`) has still to be a javascript file

sw-yx:
tried this and verified!

elvticc:
__Thanks for all your work!__ It seems that the file `tsconfig.json` is missing in the template:

```
rollup v1.32.1
bundles src/main.js → public/build/bundle.js...
(!) Plugin typescript: @rollup/plugin-typescript TS18003: No inputs were found in config file 'tsconfig.json'. Specified 'include' paths were '["**/*"]' and 'exclude' paths were '[]'.
[!] (plugin typescript) Error: @rollup/plugin-typescript: Couldn't process compiler options
Error: @rollup/plugin-typescript: Couldn't process compiler options
```

The following `tsconfig.json` content is it OK?

```json
{
  "include": ["src/**/*"],
  "exclude": ["node_modules/*"],
  "compilerOptions": {
    "target": "es2015",
    "module": "es2015",
    "types": ["svelte"]
  }
}
```
Source: https://dev.to/mhaecker/use-typescript-with-svelte-sapper-45n8

dummdidumm:
Interesting ... I didn't even need a `tsconfig.json` so I just left it out. I did not get either of those errors.

hellpirat:
I trying to use it with sapper and now I  got this error. 

```
$ sapper dev
✗ server
@rollup/plugin-typescript: Couldn't process compiler options
✗ client
@rollup/plugin-typescript: Couldn't process compiler options
internal/modules/cjs/loader.js:584
    throw err;

```

```
export const preprocess = sveltePreprocess({
  postcss: true,
[...]

PR #162, opened 2020-06-04
document TS working OOTB with svelte preprocess

sw-yx:
closes https://github.com/sveltejs/language-tools/issues/161

### DIFF
--- a/packages/svelte-vscode/docs/preprocessors/typescript.md
+++ b/packages/svelte-vscode/docs/preprocessors/typescript.md
@@ -8,17 +8,40 @@ To tell us to treat your script tags as typescript, add a `type` or `lang` attri
     export let name: string;
 </script>
 
-<!-- Or add lang="typescript" -->
+<!-- Or add lang="typescript" or lang="ts" -->
 <script lang="typescript">
     export let name: string;
 </script>
 ```
 
-Now you'll need to add a `svelte.config.js` file at the root of your project to tell svelte how to convert your TypeScript into JavaScript that it understands.
+This will work with Svelte and TypeScript as long as you enable `svelte-preprocess` and `@rollup/plugin-typescript`:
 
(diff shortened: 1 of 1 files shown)


dummdidumm:
Thanks! 😄

packages/svelte-vscode/docs/preprocessors/typescript.md
```diff
@@ -8,17 +8,40 @@ To tell us to treat your script tags as typescript, add a `type` or `lang` attri
...
+    commonjs(),
+    typescript(), // <-- added below commonjs
+    // ...
+```
+
+And this should work to enable full TypeScript checking in your Svelte files. For further discussion and a clonable template [see this issue](https://github.com/sveltejs/language-tools/issues/161).
+
+> Caveat: Your entry file (`main.js`) has still to be a javascript file
+
+You may optionally want to extract this out to a standalone `svelte.config.js` file - this used to be necessary but is no longer required. However you may wish to do this to avoid duplication of config.
```
joelmukuthu, line 44: Hi, it might be worthwhile adding an example for this. At first read I got the impression that I could move the config to a `svelte.config.js` file and not even have to import it into the Rollup config. Or is that possible now? 
And nit-picky, but I would remove the "this used to be necessary bit" :)

dummdidumm: I updated the doc a little, see https://github.com/sveltejs/language-tools/blob/master/packages/svelte-vscode/docs/preprocessors/typescript.md . I removed both the "used to be" reference and the "you can deduplicate" info for brevity.

joelmukuthu: Perfect! With that change, I think (hope) the user will know to refactor their rollup config if they use the svelte config file :)

PR #165, opened 2020-06-05
fixing memory issues

jasonlyu123:
1. [x] fix onWatchFileChanges add ignored files.
2. [x] default to ignore `__sapper__`, `node_modules` and maybe `dist` .
3. [ ] reparse config when file got added.
4. [ ] fix memory leak when user add/remove/rename an jsconfig.json/ tsconfig.json

### DIFF
--- a/packages/language-server/src/plugins/typescript/SnapshotManager.ts
+++ b/packages/language-server/src/plugins/typescript/SnapshotManager.ts
@@ -8,7 +8,7 @@ export class SnapshotManager {
     private documents: Map<string, DocumentSnapshot> = new Map();
 
     updateByFileName(fileName: string, options: SvelteSnapshotOptions) {
-        if (!this.projectFiles.includes(fileName)) {
+        if (!this.has(fileName)) {
             return;
         }
 
@@ -24,6 +24,10 @@ export class SnapshotManager {
         }
 
         this.set(fileName, newSnapshot);
+    }
+
+    has(fileName: string) {
+        return this.projectFiles.includes(fileName) || this.getFileNames().includes(fileName);
     }
 
     set(fileName: string, snapshot: DocumentSnapshot) {
--- a/packages/language-server/src/plugins/typescript/SnapshotManager.ts
+++ b/packages/language-server/src/plugins/typescript/SnapshotManager.ts
@@ -1,20 +1,30 @@
-import { DocumentSnapshot } from './DocumentSnapshot';
+import { DocumentSnapshot, SvelteSnapshotOptions } from './DocumentSnapshot';
 
 export class SnapshotManager {
-    private static managerContainer: Map<string, SnapshotManager> = new Map();
-
-    static getFromTsConfigPath(tsconfigPath: string): SnapshotManager {
-        let manager = this.managerContainer.get(tsconfigPath);
-
-        if (!manager) {
-            manager = new SnapshotManager();
-            this.managerContainer.set(tsconfigPath, manager);
-        }
-
-        return manager;
-    }
+    constructor(
+        private projectFiles: string[]
+    ) { }
 
     private documents: Map<string, DocumentSnapshot> = new Map();
+
+    updateByFileName(fileName: string, options: SvelteSnapshotOptions) {
+        if (!this.projectFiles.includes(fileName)) {
+            return;
+        }
+
+        const newSnapshot = DocumentSnapshot.fromFilePath(fileName, options);
+        const previousSnapshot = this.get(fileName);
+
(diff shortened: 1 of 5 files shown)


packages/language-server/src/plugins/typescript/SnapshotManager.ts
```diff
@@ -1,20 +1,30 @@
...
 
-    static getFromTsConfigPath(tsconfigPath: string): SnapshotManager {
-        let manager = this.managerContainer.get(tsconfigPath);
+    private documents: Map<string, DocumentSnapshot> = new Map();
 
-        if (!manager) {
-            manager = new SnapshotManager();
-            this.managerContainer.set(tsconfigPath, manager);
+    updateByFileName(fileName: string, options: SvelteSnapshotOptions) {
+        if (!this.projectFiles.includes(fileName)) {
```
dummdidumm, line 11: Haven't tested it, but will this work if you add a new `ts` file later on? Or will such a file always be ignored even if it's in a valid place?

jasonlyu123: Yeah. the new project file needed to be reparsed show.

dummdidumm: [What Vetur does](https://github.com/vuejs/vetur/blob/fdf7009cf39c5cb55a2f327042dd10ae751b56be/server/src/services/typescriptService/serviceHost.ts#L182) is to delete the old snapshot, so that the next time snapshots are searched for, it retrieves a new version. Maybe we could do something similar if the reparsing is hard/too performance heavy. For example we could mark the snapshot as "needs update" and the next time it's retrieved it's retrieved from file system again and the version is upped by one, or save the versions in another map and decide depending on that.

jasonlyu123: I think it would also be ok if we bump it here. and when language-service requests a new snapshot cached it and add it to the white list. In this way, we don't have to separate the snapshot and the version.

jasonlyu123:
Known issue, like vetur, new js/ts file currently won't have auto-import. but once it got manually imported it would be normal again.

dummdidumm:
If the user adds an `include` in its tsconfig/jsconfig, maybe we could also use that to test for "should we let this file pass" inside `SnapshotManager`. What do you think? Same thing could be done the other way round for `exclude`. We would need to compute the absolute path by appending the relative paths in `include / exclude` to the path of the jsconfig/tsconfig, and then use those to test.

jasonlyu123:
I was thinking if the `parseJsonConfigFileContent` does it. If so we could change the `onWatchFileChanges` to accept array of changes and only reparse config once to update the project filenames

dummdidumm:
Maybe we don't even need to reparse it, if we do it like so:

SnapshotManager:
```
updateByFileName(name: string) {
   const isIncluded = this.include.some(include => name.startsWith(include));
   const isExcluded = this.exclude.some(exclude => name.startsWith(exclude));
   if (isIncluded && !isExcluded) {
      // ...
   }
```

and `this.include/exclude` would be set as arrays of absolute paths from the ts config on creation of the `SnapshotManager`.

jasonlyu123:
Yeah that true. It could save some io operation and probably not a very complicated logic to implement. But the method you mention should be changed to use glob to test and also the `files` option also needs to be considered.

packages/language-server/src/plugins/typescript/service.ts
```diff
@@ -155,21 +160,26 @@ export function createLanguageService(
...
-            compilerOptions = { ...compilerOptions, ...parsedConfig.options };
-            files = parsedConfig.fileNames;
-        }
+
+        const defaultExclude = [
+            '__sapper__',
+            'node_modules'
+        ];
+
+        const config = Object.assign({ exclude: defaultExclude }, configJson);
```
jasonlyu123, line 169: https://www.typescriptlang.org/docs/handbook/tsconfig-json.html#configuration-inheritance-with-extends
This will overwrite the base config where the tsconfig extends from

jasonlyu123:
Both default to exclude `__sapper__` and check if a file should be included in `snapshotManager` turns out to be not so easy because of `extends`. I only found some typescript internal api to get the merged files-related options.

dummdidumm:
If this gets too time consuming to implement right now, we should start with the "typescript files are not found for imports until they are imported in a svelte file for the first time" behavior and do a separate PR for exploring other ways.

jasonlyu123:
Ok! That's deal with in another PR

dummdidumm:
👍

Issue #143, opened 2020-05-30
Unexpected token. Did you mean `{'>'}` or `&gt;`?


mgholam:
![svelte](https://user-images.githubusercontent.com/460565/83326338-7d520280-a288-11ea-8265-a5303c1bb390.png)

**Describe the bug**
On a perfectly good `app.svelte ` file I'm getting the above.

**System (please complete the following information):**
 - OS: Windows
 - IDE: VSCode
 - Plugin/Package: Svelte Beta

dummdidumm:
Could you paste the complete code of that file?

jasonlyu123:
This is probably because of the code before it got transformed with a missing brackets. Can you provide the code before it?

mgholam:
```js
<script>
  import { onMount } from "svelte";
  import u from "./Helper/utils.js";
  import msgbox from "./Components/msgbox.js";
  import Tabs from "./Components/tabs.svelte";
  import Button from "./Components/Button.svelte";
  import InboxTab from "./Main/Inbox.svelte";
  import ArchiveTab from "./Main/Archive.svelte";
  import HelpTab from "./Main/Help.svelte";
  import SettingsTab from "./Main/Settings.svelte";
  import DocOutTab from "./Main/DocOut.svelte";

  let mainid = "mainid";
  let server = "";
  let TABS;
  let timer = null;
  // let RTL = false;

  onMount(() => {
    u.addPrototypes();

    TABS.addtab({
      name: "inbox",
      description: "$Inbox$",
      icon: "gi-save-file", //icons.faFolderOpen
    });
    TABS.addtab({
      name: "docout",
      description: "$Docs Out$",
      icon: "gi-open-file", //icons.faFilter
    });
    TABS.addtab({
      name: "archive",
      description: "$Search Letters$",
      icon: "gi-search", //icons.faFilter
    });
    TABS.addtab({
      name: "settings",
      description: "$Settings$",
      icon: "gi-cog", // icons.faCog
    });
    TABS.addtab({
      name: "help",
      description: "$Help$",
      icon: "gi-question-sign",
    });

    setTimeout(() => {
      // kludge to get the server for fetch
      u.GET("/api/getuser").then((r) => (server = r));

      // server = window.ServerURL;
      u.GET("/api/inbox.getsettings").then((data) => {
        if (data.Color) {
          u.SetTheme(data.Color);
        }
        if (data.Language == "fa") {
          document.body.dir = "rtl";
[...]

PR #167, opened 2020-06-06
(fix) refactor extractTag to use parse5 instead of RegEx

sw-yx:
closes https://github.com/sveltejs/language-tools/issues/143

Because parse5 parses tags in a very specific way, i had to tweak one of the tests, I will comment below.

### DIFF
--- a/packages/language-server/src/lib/documents/utils.ts
+++ b/packages/language-server/src/lib/documents/utils.ts
@@ -30,12 +30,12 @@ type ParsedNode = {
     parentNode: ParsedNode;
     sourceCodeLocation: Location & { startTag: Location; endTag: Location };
 };
-const re_if = new RegExp('{#if (.*?)*}', 'igms');
-const re_ifend = new RegExp('{/if}', 'igms');
-const re_each = new RegExp('{#each (.*?)*}', 'igms');
-const re_eachend = new RegExp('{/each}', 'igms');
-const re_await = new RegExp('{#await (.*?)*}', 'igms');
-const re_awaitend = new RegExp('{/await}', 'igms');
+const regexIf = new RegExp('{#if (.*?)*}', 'igms');
+const regexIfEnd = new RegExp('{/if}', 'igms');
+const regexEach = new RegExp('{#each (.*?)*}', 'igms');
+const regexEachEnd = new RegExp('{/each}', 'igms');
+const regexAwait = new RegExp('{#await (.*?)*}', 'igms');
+const regexAwaitEnd = new RegExp('{/await}', 'igms');
 /**
  * Extracts a tag (style or script) from the given text
  * and returns its start, end and the attributes on that tag.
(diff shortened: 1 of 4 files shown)


packages/language-server/test/lib/documents/utils.test.ts
```diff
@@ -129,22 +129,50 @@ describe('document/utils', () => {
...
+                {@debug myvar}
                 <!-- p{ color: blue; }</script> -->
                 <!--<script lang="scss">
                 p{ color: blue; }
                 </script> -->
-                <scrit>blah</script>
+                <scrit>blah</scrit>
                 <script>top level script</script>
             `;
+            // Note: cannot test <scrit>blah</scriPt> as that breaks parse5 parsing for top level script!
```
sw-yx, line 141: this used to work under the old regex impl, but doesn't work when using parse5, because the `</script>` causes parse5 to swallow the subsequent `<script>top level script</script>`, making it undetectable for our purposes. Better to abandon support for this funky error case imo and have a better overall parsing strategy.

packages/language-server/test/lib/documents/utils.test.ts
```diff
@@ -129,22 +129,50 @@ describe('document/utils', () => {
                     </script>
                 {/await}
                 <p>{@html <script> consolelog('not top level')</script>}</p>
+                {@html mycontent}
+                {@debug myvar}
```
sw-yx, line 133: added a couple missing directives for good measure

packages/language-server/src/lib/documents/utils.ts
```diff
@@ -11,29 +12,30 @@ export interface TagInformation {
...
+    childNodes: ParsedNode[];
+    parentNode: ParsedNode;
+    sourceCodeLocation: Location & { startTag: Location; endTag: Location };
+};
+const re_if = new RegExp('{#if (.*?)*}', 'igms');
+const re_ifend = new RegExp('{/if}', 'igms');
+const re_each = new RegExp('{#each (.*?)*}', 'igms');
+const re_eachend = new RegExp('{/each}', 'igms');
+const re_await = new RegExp('{#await (.*?)*}', 'igms');
+const re_awaitend = new RegExp('{/await}', 'igms');
```
sw-yx, line 38: placed outside to hopefully improve performance. not sure if the regex can be better here

packages/language-server/test/lib/documents/utils.test.ts
```diff
@@ -129,21 +129,49 @@ describe('document/utils', () => {
...
-                startPos: Position.create(32, 24),
-                endPos: Position.create(32, 40),
-                container: { start: 1140, end: 1173 },
+                start: 254,
+                end: 270,
+                startPos: Position.create(7, 20),
+                endPos: Position.create(7, 36),
+                container: { start: 246, end: 279 },
             });
         });
```
sw-yx, line 176: this is a new test, the usecase we are actually trying to fix.  github misattributes the changed numbers here from the test above

packages/language-server/src/lib/documents/utils.ts
dummdidumm: I replaced the whitespace with `\s`, other than that looks good.

dummdidumm:
Great work, thanks!

sw-yx:
sorry about the failing test, i actually saw it and fixed it, but failed to push it up to my branch 🤦

Issue #110, opened 2020-05-23
Support rename in Svelte files


dummdidumm:
**Is your feature request related to a problem? Please describe.**
Currently, I cannot do rename refactorings in svelte files.

**Describe the solution you'd like**
Allow rename refactorings inside svelte files to work so that
- I can change prop names and all dependent files get updated
- I can change variables/functions inside svelte files
- I can change variables/functions outside in js/ts files and if they are used inside svelte files, those are updated aswell

**Additional context**
I started prototyping this. Renames of variables/functions inside svelte files can be implemented quickly. 

However implementing prop renames requires changes to `svelte2tsx`. It seems that right now there is no mapping from `export let X` to the generated jsx-prop. 

Another problem is that when you do a rename inside ts/js, the language server does not pick that up because he is only registered to react to changes of svelte files. To fix that, it would need to register for js/ts files, too, which requires some work to differ between svelte and ts/js documents and not for example run diagnostics of the `HTMLPlugin` on js/ts files. Maybe this could come as a second step, after renaming inside svelte files works.

dummdidumm:
Rename so far now works for:
- Variables inside a svelte file's script tag
- Component props, but they can only be renamed inside the component which exports the prop

It does not yet work for:
- Component props if you want to rename the prop but are in another file which uses that component
- Generally attributes/bindings inside the template cannot be renamed when you want to trigger the rename in said template. Works inside script tag though.

imekachi:
There is another scenario, where a svelte file imports a variable from another file (.js, .ts)
and you rename that variable in other file, it doesn't update variable imported in .svelte

I think it has something to do with referencing variables, because it cannot find reference from .svelte file as well
![image](https://user-images.githubusercontent.com/5445243/115429944-1c2f7680-a22e-11eb-9677-a303e8ab98a9.png)

dummdidumm:
To support this we need to write a TypeScript plugin, which is tracked in #580

dummdidumm:
There exists a TypeScript plugin now which comes packaged with the VS Code extension and which you need to enable [through the settings](https://github.com/sveltejs/language-tools/tree/master/packages/svelte-vscode#svelteenable-ts-plugin). It also is [available standalone as a npm package](https://www.npmjs.com/package/typescript-svelte-plugin) if you need to use it outside of VS Code.
The plugin fixes the issue of renaming stuff inside TS/JS files and updating Svelte files, too.

imekachi:
[...]

PR #172, opened 2020-06-06
(feat) basic rename

dummdidumm:
#110

- [x] rename where everything is inside same svelte file works
- [x] rename of variable inside svelte which is also in another ts/js file works
- [ ] rename of svelte prop works when doing the rename for prop of component A in component B which uses that prop
- [x] rename of svelte prop works when doing the rename for prop of component A in component A

Out of scope for this PR:
-  rename in ts file which would affect svelte file -> out of this scope because we would also need to trigger the language server on ts/js files then


jasonlyu123:
Found an issue. if I rename a props that is html attribute like disabled. It would do a rename on the html attribute no matter that props exist on the component or not.

dummdidumm:
Good catch, thank you. There is also a prop rename problem. I think I need to enhance `svelte2tsx` with some diagnostic capabilities like "what other svelte components does this file import?" or "what props does this file have?".

dummdidumm:
I ended up using `getRenameInfo` of the ts language service. The downside is that component properties can only be renamed within that component now. I added a TODO about this, this can be tackled later. I also added `prepareRename` support so it shows the user early that this cannot be renamed.

dummdidumm:
@jasonlyu123 could you take one more look? 😃 Also code-wise, I had to jump through some hoops to get it working.

packages/language-server/src/lib/documents/utils.ts
```diff
@@ -215,3 +215,33 @@ export function isInTag(position: Position, tagInfo: TagInformation | null): boo
 export function getTextInRange(range: Range, text: string) {
     return text.substring(offsetAt(range.start, text), offsetAt(range.end, text));
 }
+
+export function getVariableAtPosition(position: Position, text: string) {
+    return getVariableAtOffset(offsetAt(position, text), text);
+}
+
+export function getVariableAtOffset(offset: number, text: string) {
```
jasonlyu123, line 223: also variable.b, variable) and operators without space in between.

jasonlyu123:
```svelte
<input bind:value />
```
would be renamed to
```svelte
<input bind:valuvalues />
```

packages/language-server/src/plugins/typescript/features/RenameProvider.ts
```diff
@@ -0,0 +1,325 @@
...
+        fragment: SvelteSnapshotFragment,
+        position: Position,
+        convertedRenameLocations: (ts.RenameLocation & { range: Range })[],
+        fragments: Map<string, SnapshotFragment>,
+        lang: ts.LanguageService,
+    ) {
+        // First find out if it's really the "rename prop inside component with that prop" case
+        // Use original document for that because only there the `export` is present.
+        const regex = new RegExp(
+            `export\\s+(const|let)\\s+${getVariableAtPosition(
```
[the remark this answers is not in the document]
jasonlyu123: @sw-yx do you know about export const, function, class? I only found it to be available in the component instance, if so it probably doesn't needs to be checked.

packages/language-server/src/lib/documents/utils.ts
jasonlyu123: I found a hack that can be used to find the identifier/symbol in the generated tsx:
```ts
const offset = fragment.offsetAt(fragment.getGeneratedPosition(position));
const { start, length } = lang.getSmartSelectionRange(tsDoc.filePath, offset).textSpan;
const identifier = tsDoc.getText(start, start + length);
```

dummdidumm:
This is a source map problem, seems like `svelte2tsx` mappings are off after conversion. I was able to fix it. I also added your suggestion for the variable retrieval and added a TODO that `export {x as y}` syntax is currently not supported for rename. The function/class thing ... let's just wait till someone shows up with this request 😄

Issue #170, opened 2020-06-06
Warning: Store + Typescript + increments\decrements


Th0rN13:
**Describe the bug**
Store with number type warned about increments\decrements
`$count++` -
`The operand of an increment or decrement operator must be a variable or a property access. ts(2357)`

**To Reproduce**
I use template https://github.com/dummdidumm/template

App.svelte
```
<script lang="ts">
  import { writable } from 'svelte/store';
  const count = writable<number>(0);
</script>

<button on:click={() => $count++}>add</button>
<button on:click={() => $count = 0}>clr</button>
{$count}
```

**Expected behavior**
No warnings

**System (please complete the following information):**
 - OS: Windows
 - IDE: VSCode
 - Plugin: "Svelte Beta"

sw-yx:
interesting. haven't investigated, but what does typescript think $count is?

Th0rN13:
On hover shows 
```
(alias) const count: Writable<number>
import count
```

dummdidumm:
We have to check what `svelte2tsx` transforms this to. There is a  `__sveltets_store_get` function, maybe it gets in the way there.

jasonlyu123:
It got transformed to `__sveltets_store_get($count)++`. Need to add a check for imcrement/decrement and change it to  `__sveltets_store_get($count)+ 1` or something else

dummdidumm:
`$store += 1 ` will also throw then I guess.

sw-yx:
@dummdidumm i also addressed `$store += 1` in #173. ended up taking more code than the original issue haha

PR #173, opened 2020-06-06
(fix) svelte2tsx handling increment postfix, prefix and Assignment operators

sw-yx:
closes https://github.com/sveltejs/language-tools/issues/170

this was a fun one to solve :) `svelte2tsx.ts` has some regrettable duplication between `processInstanceScriptContent` and `processSvelteTemplate` which use `ts.createSourceFile` and `convertHtmlxToJsx` separately, hence every new language feature supported must be done for 2 different ASTs. its fine now but could grow to unmanageable scale. i am wondering whether or not it makes sense to significantly reduce the scope of `convertHtmlxToJsx` and use ts' AST internally - but havent really investigated enough to know for sure


packages/svelte2tsx/src/svelte2tsx.ts
```diff
@@ -83,6 +83,41 @@ function processSvelteTemplate(str: MagicString): TemplateProcessResult {
...
+        // handle $store++, $store--, ++$store, --$store
+        // https://github.com/sveltejs/language-tools/issues/170
+        if (parent.type == 'UpdateExpression') {
+            let appendor;
+            if (parent.operator === '++') appendor = '+';
+            if (parent.operator === '--') appendor = '-';
+            if (appendor === undefined)
+                throw new Error(
+                    'warning - unrecognized UpdateExpression operator! Probably some edge case unaccounted for in svelte2tsx, please file an issue.',
+                );
```
sw-yx, line 112: not actually sure if this is needed but just being cautious - are there any other operators like ++ and --? i dont know.

packages/svelte2tsx/src/svelte2tsx.ts
```diff
@@ -83,6 +83,40 @@ function processSvelteTemplate(str: MagicString): TemplateProcessResult {
             str.appendLeft(parent.end, ')');
             return;
         }
+        // handle Assignment operators ($store +=, -=, *=, /=, %=, **=)
+        const operators = ['+=', '-=', '*=', '/=', '%=', '**='];
```
Th0rN13, line 87: What about bitwise operators?
From [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Expressions_and_Operators)
`<<= >>= >>>= &= ^= |=`

sw-yx: haha. honestly doubt anyone uses bitwise operators with stores...

halfnelson:
Nice fix. Thanks!

packages/svelte2tsx/src/svelte2tsx.ts
sw-yx: image
dammit look what u made me do

dummdidumm:
Nice one!

About the duplicated code: Maybe we can find a intermediate representation which both AST and ts map to, and then the logic is executed only once from this intermediate representation.

PR #183, opened 2020-06-11
VSCode: command to preview compiled js code

hackape:
Implement a command that open a readonly editor to the side, content of which is the compiled js code.


packages/language-server/src/plugins/svelte/SveltePlugin.ts
```diff
@@ -25,6 +25,29 @@ import { getHoverInfo } from './features/getHoverInfo';
 import { SvelteDocument } from './SvelteDocument';
 import { Logger } from '../../logger';
 
+export type SvelteCompiledResult = {
```
jasonlyu123, line 28: What do you think about just use
```ts
import type { compile } from 'svelte/compiler';

export type SvelteCompiledResult = ReturnType<typeof compile>
```

hackape: Good point. I thought that would also import `compile` function as value, that's why I manually wrote it. Turns out TS is smart enough to figure out that value can be dropped.

hackape: Do you think it's better to use type only import?
```ts
import type { compile } from 'svelte/compiler';
```
It requires TS v3.8+ tho.

sw-yx:
nice - can you also add a mention of this to the docs/readme so that its more easily discoverable? also, does it live update as i type?

hackape:
@sw-yx eh, no... I didn't impl live update. Good point.

packages/language-server/src/plugins/svelte/SveltePlugin.ts
jasonlyu123: We are using ts 3.9 now. but I think using normal import could probably be the same. It could be removed.

sw-yx:
might as well replicate the repl :)

hackape:
@sw-yx looking forward to do that some day. but that'd be too much work for me now. I've just push the impl of live-updating compiled code content.

sw-yx:
@hackape happy to help you document, thats the easy part :)

dummdidumm:
So nice! I will take care of resolving the conflict and add add sth to the readme.

PR #200, opened 2020-06-17
Fix error when no content is in the script

### LINKED ISSUE #197: Error when no content is in the script
**Describe the bug**
Error when no content is in the script,

error logged:

Notification handler 'textDocument/didChange' failed with message: Cannot read property 'startOffset' of undefined
```
TypeError: Cannot read property 'startOffset' of undefined
(C:\Projects\language-tools\packages\language-server\dist\src\lib\documents\utils.js:102:28)
    at SvelteDocumentSnapshot.get scriptKind [as scriptKind] (C:\Projects\language-tools\packages\language-server\dist\src\plugins\typescript\DocumentSnapshot.js:105:92)
    at Object.updateDocument (C:\Projects\language-tools\packages\language-server\dist\src\plugins\typescript\service.js:87:67)
...
```

**To Reproduce**
```svelte
<script></script>
```

**Expected behavior**
no error

 - Plugin/Package: `svelte-language-server`,

bradyjoslin:
Close #197 
When a parse error is caught, check if script is empty, and if so don't display the error.

### DIFF
--- a/packages/language-server/src/plugins/typescript/DocumentSnapshot.ts
+++ b/packages/language-server/src/plugins/typescript/DocumentSnapshot.ts
@@ -132,11 +132,17 @@ function preprocessSvelteFile(document: Document, options: SvelteSnapshotOptions
             character: e.start?.column ?? 0,
         };
         const end: Position = e.end ? { line: e.end.line - 1, character: e.end.column } : start;
-        parserError = {
-            range: { start, end },
-            message: e.message,
-            code: -1,
-        };
+
+        // Suppress parse error on empty script.
+        parserError =
+            document.scriptInfo?.content === ''
+                ? null
+                : {
+                      range: { start, end },
+                      message: e.message,
+                      code: -1,
+                  };
+
         // fall back to extracted script, if any
         text = document.scriptInfo ? document.scriptInfo.content : '';
     }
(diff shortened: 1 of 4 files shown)


dummdidumm:
Thank you for tackling this! But I'd rather have this fixed inside `svelte2tsx` where the error originates. More precisely, it comes from [this line](https://github.com/sveltejs/language-tools/blob/de3af247ba7e145198f5775d824e7d26c5fe6b87/packages/svelte2tsx/src/svelte2tsx.ts#L348).

You can add a new test at `svelte2tsx/test/svelte2tsx/samples` (maybe call it `empty-script`), make it fail (run `yarn test` inside `svelte2tsx`) and then try fixing it.

sw-yx:
@dummdidumm are we sure it's that line? because the error stack trace in the original issue is:

which corresponds to this line https://github.com/sveltejs/language-tools/blob/e9880c7f83c63e415e48be00914e8e2d0b608ab2/packages/language-server/src/lib/documents/utils.ts#L118 

typescript shouldve caught this one tbh, its an issue of bad typing from the parse5 side :(

dummdidumm:
Oh then maybe it's both, because when adding said test to `svelte2tsx` it throws the error at that line.

jasonlyu123:
I also think it's both. This issue exists for quite a while.

bradyjoslin:
Thanks for the feedback.  Made the update to fix the issue instead of just suppressing error and added a test.

dummdidumm:
Great work, thanks!

sw-yx:
nice work @bradyjoslin !

PR #202, opened 2020-06-18
Support for Yarn 2 (Adding language server path option)

SagnikPradhan:
For the extension to run in harmony with yarn 2, a `pnp`-fied version of the svelte language server needs to be used instead of the native one. In order to use this custom version of the language server, a config property had to be exposed through which we load the custom language server.

On a side note, I haven't really checked whether my changes work with the "normal" projects. If someone could test my changes that would be great!

### DIFF
--- a/packages/svelte-vscode/package.json
+++ b/packages/svelte-vscode/package.json
@@ -51,6 +51,11 @@
                     "title": "Language Server Runtime",
                     "description": "Path to the node executable to use to spawn the language server"
                 },
+                "svelte.language-server.ls-path": {
+                    "type": "string",
+                    "title": "Language Server Path",
+                    "description": "Path to the langauge server file"
+                },
                 "svelte.language-server.port": {
                     "type": "number",
                     "title": "Language Server Port",
--- a/packages/svelte-vscode/package.json
+++ b/packages/svelte-vscode/package.json
@@ -54,7 +54,7 @@
                 "svelte.language-server.ls-path": {
                     "type": "string",
                     "title": "Language Server Path",
-                    "description": "Path to the langauge server file"
+                    "description": "Path to the langauge server file (either a relative path from the workspace root or an absolute path). Can be used to use a custom variant of the language server."
                 },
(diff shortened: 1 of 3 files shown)


SagnikPradhan:
Okay, it still doesn't work with relative paths, I am looking into it.

packages/svelte-vscode/src/extension.ts
```diff
@@ -32,9 +33,24 @@ namespace TagCloseRequest {
...
     const runtimeConfig = workspace.getConfiguration('svelte.language-server');
 
+    const { workspaceFolders } = workspace;
+    const rootPath = Array.isArray(workspaceFolders) ? workspaceFolders[0].uri.fsPath : undefined;
+
+    const tempLsPath = runtimeConfig.get<string>('ls-path');
+    // Returns undefined if path is empty string
+    // Return absolute path if not already
+    const lsPath =
+        tempLsPath && tempLsPath.trim() != ''
```
dummdidumm, line 45: Very minor thing: Could you use ``!==` instead?

packages/svelte-vscode/package.json
```diff
@@ -51,6 +51,11 @@
                     "title": "Language Server Runtime",
                     "description": "Path to the node executable to use to spawn the language server"
                 },
+                "svelte.language-server.ls-path": {
+                    "type": "string",
+                    "title": "Language Server Path",
+                    "description": "Path to the langauge server file"
```
dummdidumm, line 57: Could you add more details here? Like "either relative to the workspace root, or absolute. Use it when ...". Could you also add the same to the README (there is a settings section)?

SagnikPradhan:
@dummdidumm Made my changes. Is that okay?

dummdidumm:
Looks good 👍 Anything else you want to add? If not, I merge this in.

SagnikPradhan:
> 
> 
> Looks good +1 Anything else you want to add? If not, I merge this in.

I don't think there's anything else to add to this PR.

PR #213, opened 2020-06-20
fix(204): strip doctype before jsx is generated

skippednote:


dummdidumm:
Nice! Could you add a test within `svelte2tsx/test/htmlx2jsx` to confirm the new behavior?

skippednote:
Thanks, @dummdidumm! I've added a test.

dummdidumm:
🎉

Issue #77, opened 2020-05-10
Svelte Beta vs JamesBirtles / UnwrittenFun


benmccann:
Are we ready to have people start using https://marketplace.visualstudio.com/items?itemName=svelte.svelte-vscode ?

https://github.com/sveltejs/integrations points to the old plugin location at https://marketplace.visualstudio.com/items?itemName=JamesBirtles.svelte-vscode

The old plugin has about 90x as many installs. I'm wondering how we can notify these users that there's a new plugin. Some ideas in order of effectiveness:

* delete the new plugin. transfer the old plugin to sveltejs and make it the official one. now all users are using the new plugin
* push an update to the old plugin that spits out warnings somewhere that you're using the old plugin
* remove the old plugin from the marketplace and hope users eventually figure out they have the old one

dhonx:
I prefer the second option

dummdidumm:
I don't know where but I think @orta said something along the lines of "take over Svelte Plugin at some point". So this would mean the (currently old) Svelte Plugin would get an update.

As to when that is: In my opinion not yet, but soon. There is still lots of refactorings/groundwork going on so I think this repo is in a little too early state.

orta:
Yeah, I was figuring we'd ship an update to the birtle version to recommend folks move over once it "Svelte Beta" got stable and comfy (which I don't know when that is, I figure someone will probably tell me, but it's not now for sure )

benmccann:
I would guess the "Beta" one already works better than the non-beta one. Are there any known issues that make the "Beta" one less stable or a worse overall experience?

dummdidumm:
The experience is definitely better, but it's not that stable yet, as in we add new code a lot and that may break some things along the way. For example we added `svelte2tsx` to get better typing not just inside `script` but also inside template, and that that brought in some new bugs which were reported.

benmccann:
I guess even in the long-term it could be helpful to have a "beta" one and "stable" one. Then there'd be an extension that could be used to push out changes and get early feedback before giving it to all users

sw-yx:
do we have publishing control over james birtles' one? does he want to hand it over?

orta:
I don't think we do, but he's game to have it deprecated

dummdidumm:
I think we can start pushing this, Svelte Beta is getting more stable every day and people are still way too often using the old version (almost 100k by now).

orta:
Alright, agreed - OK, so the TODO IMO:

[...]

PR #219, opened 2020-06-21
Adds deploy github actions for prod builds

orta:
Separates out the metadata and build process for 'production builds' of the vscode extension


orta:
re: #77

packages/svelte-vscode/package.json
```diff
@@ -23,7 +23,7 @@
         "url": "https://github.com/sveltejs/language-tools/issues"
     },
     "homepage": "https://github.com/sveltejs/language-tools#readme",
-    "displayName": "Svelte Beta",
+    "displayName": "Svelte for VS Code",
```
dummdidumm, line 26: Bikeshedding: What do you think about naming it `Svelte (Official)`?

.github/workflows/Deploy.yml
```diff
@@ -29,6 +29,9 @@ jobs:
       # Setup the environment
       - run: 'json -I -f packages/svelte-vscode/package.json -e "this.version=\`99.0.0\`"'
       - run: 'json -I -f packages/svelte-vscode/package.json -e "this.preview=true"'
+      - run: 'json -I -f packages/svelte-vscode/package.json -e "this.name=\`svelte-vscode-nightly\`"'
+      - run: 'json -I -f packages/svelte-vscode/package.json -e "this.icon=\`icons/logo-nightly.png\`"'
+      - run: 'json -I -f packages/svelte-vscode/package.json -e "this.displayName=\`Svelte Language Tools Nightly Builds\`"'
```
dummdidumm, line 34: Bikeshedding#2: I think we should name the nightly almost the same as the stable one, just with a Nightly at the end. So for example `Svelte for VSCode (Nightly)` or `Svelte (Official, Nightly)`.

packages/svelte-vscode/package.json
orta: Not a fan personally, it feels a bit like calling yourself cool

Given the birtle's one will be deprecated it doesn't really need to stand out in that way

.github/workflows/Deploy.yml
orta: I'm not against it, but I choose this originally _because_ it's a bit more technically intimidating - people shouldn't really need to use the daily build very often and giving it the more unfriendly name was to discourage folks. We do the same for [TS' nightly builds](https://marketplace.visualstudio.com/items?itemName=ms-vscode.vscode-typescript-next) in vscode

packages/svelte-vscode/package.json
dummdidumm: Fair enough. After Birtle's is deprecated/renamed/removed, we could think about renaming it to just "Svelte"?

.github/workflows/Deploy.yml
dummdidumm: I get the intimidating part, but I still feel like people should at least be able to discover it easily. What about `<Original name> Unstable Nightly Builds` instead? The `Unstable` will trigger most people to not use it in their day-to-day-work.

orta: Works for me

packages/svelte-vscode/package.json
orta: Legit

orta:
I've shipped some changes for the old extension: https://github.com/orta/svelte-vscode-unwrittenfun/pull/1

PR #222, opened 2020-06-22
(fix) filter completions ourselves in error state

### LINKED ISSUE #217: textDocument/completion response incorrectly sets isIncomplete to true
**Describe the bug**
The [LSP specification](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_completion) says the following about the `isIncomplete` property on `textDocument/completion` responses:
> This list is not complete. Further typing should result in recomputing this list.

svelte-language-server is setting this property to true on responses, however seems to be sending a complete list of completions every time. VSCode seems to filter returned completions anyway, but some other clients don't, in particular [lsp-mode](https://github.com/emacs-lsp/lsp-mode) for Emacs. Because of this, the completion list is not narrowed, and the same (often very large, I observed 970 possible completions in a new sveltejs/template project) list of completions is sent every time a character is typed.

**To Reproduce**

1. Start a new Svelte template project (`npx degit sveltejs/template test-project`)
2. Enable verbose logging for the language server (in VSCode, this can be done by setting `svelte.trace.server` to `"verbose"`)
3. Open `src/App.svelte` and start typing something to trigger completions (eg. `console` in the `<script>` block)
4. View the LSP log (in VSCode, this can be done by pressing Ctrl+Shift+U and selecting "Svelte" in the dropdown)
5. Observe how responses to `textDocument/completion` are returning the same list for the same term, with `isIncomplete` set to `true`

**Expected behavior**
Either:

1. `isIncomplete` should be set to `false`, and the full list of completions should be set, or
2. `isIncomplete` should remain `true`, but the list of completions should be filtered based on what the user has typed.

The first solution is likely the easiest to implement.

 - OS: Ubuntu 18.04 (under WSL 2)
 - IDE: All IDEs with LSP client implementations that don't filter when `isIncomplete` is `true`, initially noticed on Emacs 28 with lsp-mode
 - Plugin/Package: `svelte-language-server`

**Additional context**
If you want to test this in Emacs, the `lsp-svelte` branch of [my fork of lsp-mode](https://github.com/Sorixelle/lsp-mode/tree/lsp-svelte) has support for the Svelte language server.

dummdidumm:
Fixes #217

### DIFF
--- a/packages/language-server/src/plugins/typescript/features/CompletionProvider.ts
+++ b/packages/language-server/src/plugins/typescript/features/CompletionProvider.ts
@@ -9,7 +9,7 @@ import {
     TextEdit,
 } from 'vscode-languageserver';
 import { Document, mapCompletionItemToOriginal, mapRangeToOriginal } from '../../../lib/documents';
-import { isNotNullOrUndefined, pathToUrl } from '../../../utils';
+import { isNotNullOrUndefined, pathToUrl, regexLastIndexOf } from '../../../utils';
 import { AppCompletionItem, AppCompletionList, CompletionsProvider } from '../../interfaces';
 import { SvelteSnapshotFragment } from '../DocumentSnapshot';
 import { LSAndTSDocResolver } from '../LSAndTSDocResolver';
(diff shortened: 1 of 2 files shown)


Sorixelle:
I noticed this commit only changes code in the typescript plugin. It also affects CSS, HTML tag/attribute and Svelte `@` (html, debug) blocks (but not `:`, `#` or `/`) autocomplete as well. Does this code fix those too? Should have made that clearer in the issue, sorry.

dummdidumm:
Only the ts/js-related autocompletions have `incomplete: true` set. The others don't, so they don't need filtering. At least, this is how I understood your issue.

Issue #207, opened 2020-06-18
Svelte Beta breaks other extensions like vscode-javascript and prettier-vscode


rob-balfre:
**Describe the bug**
[JavaScriptSnippets](https://marketplace.visualstudio.com/items?itemName=xabikos.JavaScriptSnippets) stops working or becomes very slow in Svelte components. Also constantly breaking prettier-vscode formatting.

**To Reproduce**
Install [JavaScriptSnippets](https://marketplace.visualstudio.com/items?itemName=xabikos.JavaScriptSnippets), open a reasonably large svelte file and try to trigger a snippet.

**Expected behavior**
Snippets should open instantly like they do in js files not hang for seconds / indefinitely

**System (please complete the following information):**
 - OS: MacOS 10.15.5
 - IDE: VSCode 1.46.1
 - Plugin/Package: Svelte Beta and JavaScriptSnippets

jasonlyu123:
Svelte Beta is also a format provider, we have our own set of prettier using prettier's npm package. As far as I know, you can't use both. Is your setup works at all or it works but extremely slow.

rob-balfre:
I removed prettier but JavaScriptSnippets still doesn't work.

rob-balfre:
Video of the problem: https://streamable.com/23vdzs

Takes 40 seconds for the snippet to appear!

jasonlyu123:
What I said is only about prettier though. These two are separate issues.

About JavaScriptSnippets, does other javascript completion also that slow?

rob-balfre:
@jarrodldavis fair enough, I was guessing it's the language server hanging causing vscode (including extensions) to hang.

Yes all auto-completion is slower, just not quite 40s slow, more 5/10 Secs

jarrodldavis:
@rob-balfre Did you mean to mention me or @jasonlyu123?

rob-balfre:
Yep, total accident. Sorry

dummdidumm:
I opened a ~150 lines component and the suggestion appeared instantly.

Could you reproduce the error after a window reload and then paste the VSCode-log of Output-Svelte in here?

rob-balfre:
```Initialize language server at  /Users/xx/xxui
Initialize new ts service at  
Warning - unrecognized UnaryExpression operator 53! 
                This is an edge case unaccounted for in svelte2tsx, please file an issue:
                https://github.com/sveltejs/language-tools/issues/new/choose
                 !$xxelections
[...]

Issue #151, opened 2020-06-01
Heavy memory usage of the Language Server


illright:
This is some sort of documentation of my experience with Svelte Language Server (SLS) that I think needs attention.

In my moderately large Svelte project I decided to try using the Svelte Atom extension (VS Code has proven to be the same in regards to this issue, which makes sense – it's not the editor's fault). After a few moments of coding I've noticed severe performance drops and system freezes. Turns out that the SLS process was using up to 2 GB of RAM. After pulling the project apart I've discovered that the files that are causing such heavy memory consumption are `/jsconfig.json` and `/__sapper__/*`. It's worth noting that at the time I had both the development and the production builds compiled, so the `__sapper__` folder was twice the size.

`jsconfig.json` had a particularly interesting field:
```json
{
  "exclude": ["node_modules", "dist"]
}
```
Turns out that SLS was honoring this field somehow, in that:
* removing `jsconfig.json` altogether  brought the memory consumption to acceptable (~350 MB)
* keeping it in this way was giving the ridiculous 2 GB of memory usage
* adding `"__sapper__"` into the array of excludes also brought the memory consumption to acceptable (same as when removing)

My thoughts on this:
* why is SLS affected by a bunch of `.js` files in the `__sapper__` folder? (note that copypasting the same file in `__sapper__/dev/client` didn't increase the memory consumption, so it's not simply linearly increasing with respect to the amount of files)
* is all this consumed memory really necessary?
* how exactly does `jsconfig.json` affect SLS?
* perhaps it should be somehow made clearer how to tweak memory consumption if it gets out of control, possibly by documenting the usage of `jsconfig.json`

I don't really know what to make of this issue, hoping for some sort of discussion and possibly documentation improvement, but feel free to close this.

jasonlyu123:
The language server use typescript's language service behind the scenes. I'll guess the high memory is because typescript tries to parse all files [configured](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html#details), in the `jsconfig.json`, to be included. Because the typescript language service is in the same process as the svelte language server, I don't know how much memory is used by typescript language service.

I am curious about how big is your `__sapper__` folder. My project has a `jsconfig.json` and set to include about 200 js source files and only use 150MB. Can't imagine how does your `__sapper__` folder cause the language server to use more than 2GB of memory.

illright:
[...]

PR #199, opened 2020-06-17
add support for svelte component import completion

jasonlyu123:
圖片
close #195 
Basically it is just telling ts language service to also looks for `.svelte` file

PR #223, opened 2020-06-22
(fix) add include when there's no js/ts config

dummdidumm:
To not flood the initial files with potentially completely unrelated .js/.ts files

#207 , #151

@jasonlyu123 as discussed.


jasonlyu123:
I just thought of another question. Wouldn't `include: [**/*.svelte]`  just the same as `include: []` here? 
Because `.svelte` is not a supported extension for typescript. And if we use the same technique as #199. It would load all svelte files in the startup, but currently, it's not.

dummdidumm:
Just checked, yes you are right. Because it's not a supported extension, nothing is loaded. I will do `include: []` then for clarity.

Issue #177, opened 2020-06-09
Typescript Support: Doesn't re-interperet typescript imports as they change. Must restart Svelte server every time.


samdesota:
**Describe the bug**
Thanks for investing time on this, auto imports alone are massively valuable with svelte.

Unfortunately, as I continue to change typescript files, the svelte imports don't update. If I add a new function or update a functions type in a typescript file outside the svelte file, svelte doesn't recognize it, event when I save the changes. 

To get svelte to update, I restart the language server.

**To Reproduce**
Create project, tsconfig / svelte config below. Use latest version of the VSCode. Observe language server does not observe changes from imported typescript files.

tsconfig:
```
{
  "compilerOptions": {
    "module": "commonjs",
    "downlevelIteration": true,
    "noImplicitAny": false,
    "removeComments": true,
    "preserveConstEnums": true,
    "sourceMap": true,
    "baseUrl": "src",
    "target": "ES2019",
    "esModuleInterop": true,
    "isolatedModules": true,
    "lib": ["ES2019", "dom"],
    "types": ["svelte", "@sapper", "global"],
    "typeRoots": ["typings"]
  },

  "include": ["src/**/*"]
}
```

svelte.config.js
```
const preprocess = require("svelte-preprocess");

module.exports = {
  preprocess: [preprocess],
};
```
**System (please complete the following information):**
 - OS: macOS 10.15.4
 - IDE: VSCode
 - Plugin/Package: Svelte Plugin v. 0.9.3

jasonlyu123:
[svelte beta](https://marketplace.visualstudio.com/items?itemName=svelte.svelte-vscode) has this now.

samdesota:
Ah, got it. They looked like the same extension. I guess I'll close this if it's on the pipeline, using Svelte Beta now and it's working well. Thank you!

samdesota:
Hmm, after some more testing I'm still not seeing updates propagate into svelte modules.

jasonlyu123:
There's a known issue that the diagnostics don't get updated if dependent file changes, other things like completion or hover should be fine. You could leave this open for diagnostics.

dummdidumm:
If it's about typescript files you don't see changes of: you have to a) save the files (editing is not enough) and b) if it's about diagnostics, edit the svelte file (like put a space in or sth). Reason: the language server only triggers on svelte files. Typescript file updates are only propagated through a file watcher and that needs to see the changes on disk, not just in the editor.

[...]

PR #237, opened 2020-06-28
update all opened document diagnostics on save

jasonlyu123:
related to #177 

update the diagnostics of all opened svelte files when a svelte file or js file is saved to disk. Mainly for typescript type errors.

### DIFF
--- a/packages/language-server/src/lib/documents/DocumentManager.ts
+++ b/packages/language-server/src/lib/documents/DocumentManager.ts
@@ -13,7 +13,7 @@ export type DocumentEvent = 'documentOpen' | 'documentChange' | 'documentClose';
  */
 export class DocumentManager {
     private emitter = new EventEmitter();
-    private openedByServer = new Set<string>();
+    private openedInClient = new Set<string>();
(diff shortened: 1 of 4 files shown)


packages/language-server/src/lib/documents/DocumentManager.ts
```diff
@@ -13,13 +13,15 @@ export type DocumentEvent = 'documentOpen' | 'documentChange' | 'documentClose';
...
     private emitter = new EventEmitter();
+    private openedByServer = new Set<string>();
     public documents: Map<string, Document> = new Map();
     public locked = new Set<string>();
     public deleteCandidates = new Set<string>();
 
     constructor(private createDocument: (textDocument: TextDocumentItem) => Document) {}
 
     openDocument(textDocument: TextDocumentItem): Document {
+        console.log(textDocument.uri);
```
dummdidumm, line 24: Left-over-console.log 😄

jasonlyu123: thanks! forgot that

PR #246, opened 2020-06-29
(fix) svelte-preprocess name typo

kaisermann:
Just installed `v100` to give it a try and, first of all, congrats for the hark work put into this 🍾 🎉 ! 

However, I've noticed the diagnostic tip to add `lang="..."` to `style`/`script` tags is referring to a `svelte-preprocessor` which I believe is a typo. If not, feel free to close this pull request 😁


dummdidumm:
Totally a typo, thanks! 😄

PR #258, opened 2020-07-01
(fix) Comma syntax in guard for If Block

### LINKED ISSUE #231: Parentheses with comma operator in #if causes "Expression expected"
**Describe the bug**
Comma operator in parentheses causes the error "Expression expected"
I found it in `{#if }`, but there may be similar behavior in other places.
It not very usable, but can be used for debug

**To Reproduce**
```
{#if (true, false)}
  content
{/if}
```

**Expected behavior**
No error

 - OS: Windows
 - IDE: VSCode
 - Plugin/Package: Svelte Beta

arpitkalla:
Adding comma syntax in the guard for If blocks. #231. 

Original svelte code:
```
{#if (true, false)}
  content
{/if}
```

Gets converted to:
```
<>{() => {if (true, false){<>
content
</>}}}</>
```

Caveat: any guard with multiple parentheses for example:
```
{#if (((true, false)))}
  content
{/if}
```

Would get converted to:
```
<>{() => {if (true, false){<>
<h1>Hello {name}</h1>
</>}}}</>
```

### DIFF
--- a/packages/svelte2tsx/src/htmlxtojsx.ts
+++ b/packages/svelte2tsx/src/htmlxtojsx.ts
@@ -445,11 +445,10 @@ export function convertHtmlxToJsx(
             return;
         }
         // {#if expr} ->
-        // {() => { if (expr) { <>
+        // {() => { if (expr){ <>
         str.overwrite(ifBlock.start, ifBlock.expression.start, '{() => {if (');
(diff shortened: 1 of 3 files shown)


dummdidumm:
Great work, thanks!

Issue #214, opened 2020-06-21
Supports union types in props and reactive statement


jasonlyu123:
**Is your feature request related to a problem? Please describe.**
The following code will have a typescript error due to typescript's control flow analysis.
But this is very useful when it comes to a select-wrapper component.
```svelte
<script lang="typescript">
  export let value: string | number = '';

  if (typeof value === 'number') {
    console.log(value.toFixed())
  }

</script>

<select bind:value={value}></select>
```

**Describe the solution you'd like**
In the `svelte2tsx`, the `props` and reactive statement `$: {  }` needs to be taken out of the control flow context maybe by wrapping it with a callback like this:

```ts
declare function __sveltets_invalidate<T>(getValue: () => T): T

/*export*/ let value: string | number = __sveltets_invalidate(() => '')

let b: 'A' | 'B' = 'A';

() => {
$: {
    console.log(b)
}
}
```

**Additional context**
![圖片](https://user-images.githubusercontent.com/36730922/85215731-49fa2380-b3af-11ea-9bbf-39c028a3ddbb.png)

dummdidumm:
Good idea with the invalidate-wrapper around props. Maybe we don't even need to add another wrapper to reactive statements then? Or are there any cases where this will throw control flow errors?

jasonlyu123:
change the `export let` to `let` and wrap the statement in `$: {}` will have the same error

```svelte
<script lang="typescript">
  let value: string | number = '';

  $: {
    if (typeof value === 'number') {
      console.log(value.toFixed())
    }
  }
</script>
```

dummdidumm:
Ah, did not think about that. Thanks!

jasonlyu123:
The function callback for props won't work lol. Needs to find another way.

jasonlyu123:
By the way, this wrapping can be used in $: variable if we change the first $: label to let like we discussed earlier in the overall strategy thread. Because its type is depending on another `let`. So it would works.

jasonlyu123:
Another problem, this union type props will also have a type of number in the exported prop def because of the same reason

jasonlyu123:
```ts
export let someProps : string | number = ''
```

To 
```ts
declare function __sveltets__invalidateWithDefault<T>(getValue: () => T | undefined): T;

let someProps: string | number = ''
someProps = __sveltets__invalidateWithDefault(() => someProps);

```

This works, but kind of hacky

PatrickG:
```ts
[...]

PR #259, opened 2020-07-01
better union type for svelte specific syntax

jasonlyu123:
1. Reactive assignment/statement/ expression ($: ) is now wrapped with a function to get a better fit with runtime behaviour. This also means a more strict null check in strict mode. 
2. If props got an explicit type and default value. The type of it would not be the default value but the explicit type of what the user has defined.

credit @PatrickG for the idea

related #214


packages/svelte2tsx/src/svelte2tsx.ts
```diff
@@ -401,6 +400,50 @@ function processInstanceScriptContent(str: MagicString, script: Node): InstanceS
...
+            const jsDocType = ts.getJSDocType(declaration);
+            const type = tsType || jsDocType;
+
+            if (!ts.isIdentifier(identifier) || !type) {
+                return;
+            }
+            const name = identifier.getText();
+            const end = declaration.end + astOffset;
+
+            str.appendLeft(end, `;name = __sveltets_any(${name});`);
```
dummdidumm, line 423: I guess you meant to write `${name} =` instead of `name =`? If I name my exports different than `name`, I get a strange error.

PR #261, opened 2020-07-02
(fix) add compilerOptions.baseUrl if compilerOptions.paths is set

### LINKED ISSUE #228: Types for built-in elements getting overriden
**Describe the bug**
If one installs some `@types/*` npm module that somehow overrides React types then type-checking gets messed up. I can imagine why this happens, but it would be nice if it could be fixed somehow. In my case, I'm using storybook (with Svelte) but it installs a bunch of `@types/*` modules which somehow mess up the types that the language server users. I haven't identified which of the types modules is the culprit. But ideally the language server types should trump other types. 

**To Reproduce**

1. Install the @storybook/addon-knobs npm module
2. Trigger TS types to be loaded somehow, e.g. by importing the module
3. Simple HTML now triggers type errors

Example code that triggers the error:

```htmlx
<script lang="typescript">
  import * as foo from '@storybook/addon-knobs';
</script>

<div>
  <p></p>
</div>
```

**Expected behavior**
That the HTML would not trigger an error.

**Screenshots**
image

 - OS: Linux
 - IDE: VSCode
 - Plugin/Package: "Svelte Beta", `svelte-language-server`, `svelte-check`

**Additional context**
`svelte-check` also triggers the same errors that show up in the IDE.

joelmukuthu:
This ensures that the `compilerOptions.paths` setting is "accepted" by the typescript compiler. Even though the path to the `react` override is an absolute path, it seems that if no `compilerOptions.baseUrl` is set, the compiler just ignores `compilerOptions.paths` altogether. 
 
Follow up to #260 and #234. Fixes #228.

### DIFF
--- a/packages/language-server/src/plugins/typescript/service.ts
+++ b/packages/language-server/src/plugins/typescript/service.ts
@@ -186,6 +186,7 @@ export function createLanguageService(
         // svelte2tsx JSX typings and react's JSX typings.
         // This may happen if a node module has (in)directly installed/imported react's types.
         if (!configJson.compilerOptions.paths?.react) {
+            configJson.compilerOptions.baseUrl = configJson.compilerOptions.baseUrl || '.';
             configJson.compilerOptions.paths = configJson.compilerOptions.paths || {};
             configJson.compilerOptions.paths.react = [
                 ts.sys.resolvePath(resolve(__dirname, '../../../../public/sink.d.ts')),
--- a/packages/language-server/test/plugins/typescript/features/CompletionProvider.test.ts
+++ b/packages/language-server/test/plugins/typescript/features/CompletionProvider.test.ts
@@ -216,7 +216,7 @@ describe('CompletionProviderImpl', () => {
             item!,
         );
 
-        assert.strictEqual(detail, 'Auto import from ./definitions\nfunction blubb(): boolean');
+        assert.strictEqual(detail, 'Auto import from definitions\nfunction blubb(): boolean');
 
         assert.strictEqual(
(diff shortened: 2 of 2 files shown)


dummdidumm:
I tested it without the `baseUrl` in the other PR and the errors went away for me. I reproduced it like you specified in the issue. For you the error still occured?

joelmukuthu:
You're absolute right with Edit 2! I think it's the reason the tests are failing. So if there's no `baseUrl` we should probably set it to the URL of the user's current workspace? From your knowledge of the language server, is that possible? It's what I was trying to do by setting it to `.` but maybe that was a bit naive :grin: I can look into it this weekend if you agree that that's the right thing to do but you're also very welcome to take over :)

dummdidumm:
I just did a small test and confirmed that we cannot set `baseUrl` ourselves. For example rollup then does not know what this is resolved to, guesses a global variable which is wrong.

So basically we cannot fix this without breaking everyone else 😞 I'm really sorry but I think we cannot add this to `svelte-language-server` then. The only thing we can do is add a troubleshooting section and add the solution there.

joelmukuthu:
Thanks a lot for your diligence on this @dummdidumm! It all makes a ton of sense. Since we already have some guides on the svelte-vscode package README I think that would be a good place to add it. I'll send in a PR :)

dummdidumm:
I already created a PR which adds the docs and removes the code #268 - please let me know if the docs are understandable 😃

joelmukuthu:
I just saw it :grinning: I should really get in the habit of reading all my emails at once. I'll comment there :)

PR #267, opened 2020-07-03
fix svelte component type

jasonlyu123:
fix svelte component type, should be a constructor. fix typescript error of SvelteComponent is not assignable to `AConstructorTypeOf` in `bind:this`


dummdidumm:
👍

PR #268, opened 2020-07-03
(docs) add troubleshooting about conflicting react types

dummdidumm:
#228

Cannot be fixed in the language-server because it would require us to add a baseUrl to the tsconfig.json, which would break every bundler setup which does not know about this.

@joelmukuthu I'm really sorry this cannot be fixed 😞 Could you read the docs change and tell me if it is understandable?

### DIFF
--- a/packages/svelte-vscode/README.md
+++ b/packages/svelte-vscode/README.md
@@ -80,7 +80,7 @@ It's also necessary to add a `type="text/language-name"` or `lang="language-name
 
 This may be due to some library you are using having installed typings for `react`. These are picked up by the TypeScript compiler. Because we internally transform svelte to jsx, there is a clash and said error occurs.
 
-To fix this, do the following:
+The underlying [issue in TypeScript](https://github.com/microsoft/TypeScript/issues/18588) is yet to be fixed but in the meantime, one way to work around it as follows:
 
 1. Add a `sink.d.ts` with content `declare module 'react' {}`
 2. Go to your `tsconfig.json`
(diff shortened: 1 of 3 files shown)


packages/svelte-vscode/README.md
```diff
@@ -74,6 +74,40 @@ It's also necessary to add a `type="text/language-name"` or `lang="language-name
 </style>
 ```
 
+### Troubleshooting
+
+#### I get weird type errors on my html tags
+
+This may be due to some library you are using having installed typings for `react`. These are picked up by the TypeScript compiler. Because we internally transform svelte to jsx, there is a clash and said error occurs.
+
+To fix this, do the following:
```
joelmukuthu, line 83:
```suggestion
The underlying [issue in TypeScript](https://github.com/microsoft/TypeScript/issues/18588) is yet to be fixed but in the meantime, one way to work around it as follows:
```

packages/svelte-vscode/README.md
```diff
@@ -74,6 +74,40 @@ It's also necessary to add a `type="text/language-name"` or `lang="language-name
 </style>
 ```
 
+### Troubleshooting
+
+#### I get weird type errors on my html tags
+
```
joelmukuthu, line 80: Might be worthwhile adding an example to enhance understanding. I think the screenshot in #228 would suffice

dummdidumm:
Thanks for the suggestions! I adjusted the docs accordingly.

Issue #272, opened 2020-07-03
Typescript: allow not-null assertion operator in templates


dkzlv:
**Is your feature request related to a problem? Please describe.**
[Not-null assertion operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator), or simply put, the exclamation mark.

Take this code as an example:
```html
<script lang="typescript">
  export let data: { someKey: string | null };

  let show: boolean;
  $: show = !!data.someKey;
</script>

{#if show}
  {data.someKey.toUpperCase()}
{/if}
```
I have done external checks and I am pretty sure `someKey` cannot be `null`, but TS would not compile it, because `Object is possibly 'null'.`. It would be very handy to write `data.someKey!.toUpperCase()`, but currently it is not possible.

**Describe the solution you'd like**
- support the operator would be the best option
- add the possibility of ignoring the following line in Typescript (`@ts-ignore` in html comment). It may be a separate feature request as well

**Describe alternatives you've considered**
For now I see no other alternatives. The only one is to have excessive checks in `{#if}` tags.

dummdidumm:
Oh this is a tough one. The problem is that at the moment mustache tags are not preprocessed, so you cannot use anything other than basic JavaScript in there. There is an issue about this in the main repo https://github.com/sveltejs/svelte/issues/4701 . Your proposal about `@ts-ignore` could work but feels very hacky, not sure about that. Do others have opinions on this?

jasonlyu123:
tsignore would probably not work as intended sometimes. 
Also, I think there are a lot of workaround for situation that you have to use typescript in template. In many cases it could be done in script instead.

dummdidumm:
I feel the same. I'm going to close this then. It seems like template/mustache preprocessing is being worked on soon, so let's wait and use workarounds in the meantime. Sorry @dkzlv that we cannot fix this right now.

Issue #271, opened 2020-07-03
Typescript: Reactive assignments do not infer data type


dkzlv:
**Describe the bug**
Subject.

**To Reproduce**
Have these enabled in `tsconfig.json`:
```
"noImplicitAny": true,
"strictNullChecks": true,
```

and create a component with the following code:
```html
<script lang="typescript">
  export let data: { someKey: string | null };

  $: show = !!data.someKey;
</script>

{#if show}hey{/if}
```
This simple code raises two TS errors:
`Variable 'show' implicitly has type 'any' in some locations where its type cannot be determined.` for the `show` variable in the `script` tag and `Variable 'show' implicitly has an 'any' type` for the same variable in the template.

**Expected behavior**
Seems to be safe enough to infer the data type as boolean.

**System (please complete the following information):**
OS: Mac OS 10.15.3
IDE: VS Code
Plugin/Package: Svelte for VS Code (the new official one)

**Additional context**

Currently the parser does not allow neither this
```typescript
$: show = !!data.someKey as boolean;
```
nor that
```typescript
$: show: boolean = !!data.someKey;
```
The only possible (and quite simple!) solution is to duplicate the variable, which is completely fine but may not be obvious for some of svelte users.
```typescript
let show: boolean;
$: show = !!data.someKey;
// No errors after that
```

dummdidumm:
This is something we cannot change I think, but it should be documented.

jasonlyu123:
I have [some experiment with implementing]( https://github.com/jasonlyu123/language-tools/tree/reactive-assignment) infer type for reactive assignment. Maybe you can give it a try following [this step](https://github.com/sveltejs/language-tools#development) to see there is any problem.

dkzlv:
@dummdidumm Thanks for your work! Svelte has been the best for me for a long time, but now with this project I just cannot imagine what would make me leave this ecosystem ❤️ 

Back to topic.
Great! To me it is completely ok, but I also think it needs some documentation. I've converted a whole bunch of components to Typescript and encountered some quirks that were difficult to handle (the most difficult one was to find out how I should import interfaces and types in the component, took me a while to figure that out). I would be happy to help with documentation, but I do not see any in the repo. Can you point me to the place?

jasonlyu123:
I also think we should have some typescript documentation. not just like how to type a svelte event handler. There's some svelte specific syntax got transformed to do type check, like what we did in #259, for example. It could be better if we have documentation about it.

[...]

PR #275, opened 2020-07-04
(docs) restructure, more faq/troubleshooting

dummdidumm:
- The docs now live at `docs`, to prevent duplication
- Added troubleshooting/faq section to TypeScript
- Added link to docs in `svelte-check`

#271, #272 

@dkzlv could you take a look if these docs, especially the faq/troubleshooting section in TypeScript, are understandable?


docs/preprocessors/typescript.md
```diff
@@ -0,0 +1,169 @@
...
+let show: boolean; // <--- added above the reactive assignment
+$: show = !!data.someKey; // <-- `show` now has type `boolean`
+```
+
+The following is INVALID TypeScript, so DON'T do this:
+
+```ts
+$: show = !!data.someKey as boolean; // DON'T DO THIS
+$: show: boolean = !!data.someKey; // DON'T DO THIS
+```
```
dkzlv, line 126: @dummdidumm I think it is unnecessary to add this, because there's a million ways to do the same thing incorrectly, and there's no point in describing just one that is explicitly marked as and error by IDE.

dkzlv:
@dummdidumm Everything seems very understandable, thank you!

docs/preprocessors/typescript.md
dummdidumm: Good point, I'll remove it.

Issue #281, opened 2020-07-05
Docstrings are ignored by svelte2tsx


fnune:
**Edit:** made a PR to fix it here https://github.com/sveltejs/language-tools/pull/282

**Problem**

In `util.ts`:

```typescript
/** Says hello */
export function sayHello() {
  return 'hello!'
}
```

Then, in some component:

```svelte
<script type="text/typescript">
  import { sayHello } from './util'

  sayHello()
</script>
```

Hovering `sayHello` in VSCode or Vim/Neovim with coc.nvim doesn't show the docstring. ~I believe this is because these docstrings aren't added back by `svelte2tsx`?~ Edit: it was the language server provider.

**Solution**

Add docstrings that follow ~[JSDoc](https://jsdoc.app/)~ [TSDoc](https://github.com/microsoft/tsdoc) notation (which is what TypeScript uses) into the generated TSX.

**Additional context**

I found out about this shortcoming while working on a pull request for this https://github.com/sveltejs/language-tools/issues/280
I believe the fixes are separate units of work, though.

fnune:
I'm a bit confused. I believe the issue isn't actually in `svelte2tsx`, because (at least going by the tests) comments aren't removed from the produced TSX:

![image](https://user-images.githubusercontent.com/16181067/86536879-705daa00-beeb-11ea-8028-47b8c2603423.png)

fnune:
Starting to think it's the language server provider that's not making use of the comments.

Issue #280, opened 2020-07-05
Documenting components with a docstring that appears on hover documentation


fnune:
**Problem**

With React's tooling, you can comment on a component like this:

```typescript
/** Documentation that will appear on hover in other places where this is imported */
function MyComponent() { return null }
```

However, in Svelte, I haven't found a way to do this.

**Solution**

I would like to know where I can place a comment that will be taken up by the on-hover documentation of editors like VSCode and Vim/Neovim (with coc.nvim).

To illustrate, here I'm hovering `MyComponent` but I only see `import MyComponent`:

![image](https://user-images.githubusercontent.com/16181067/86531889-7b9ddf00-bec5-11ea-9a87-9459b62d40da.png)

I would like my docstring `Documentation that will appear on hover in other places where this is imported` to be included in the tooltip.

I've googled around for this for a while and tried my local setup with TypeScript and I can't find a way to do it. I found an interesting related discussion on [metadata for Svelte components](https://github.com/sveltejs/svelte/issues/2134) but the discussion is more about prop types.

**Edit:** since we're discussing whether this is better added to the compiler, I created an analogous issue here https://github.com/sveltejs/svelte/issues/5102

dummdidumm:
Since this could also be valuable for components without a script tag, I propose using the HTML comment for that. `svelte2tsx` would then look for HTML comments starting with a special tag (like `@doc`) and put the content as a jsdoc above the default export.

Opinions? Other ideas?

fnune:
I would be really interested in having a go at implementing this. However I think it needs to be designed carefully because it's a feature some devs use _very often_ every day.

fnune:
Something that comes to mind is:

```typescript
/**
 * @file Here is my documentation for this component
 */
```

At the start of the file. But I think only HTML comments work there, right?

fnune:
Also, maybe to make it more similar to the other tags, I feel like a `<svelte:documentation>` tag akin to `<svelte:head>` could be nice, but that would need to be implemented in the compiler.

fnune:
Perhaps even `<svelte:options documentation="blabla" />`?

jasonlyu123:
That will also needs some work in the compiler and the approval of the main repo. The options got verify by the compiler.

fnune:
I created an analogous issue on the main repo, let's see if we can reach some conclusion :+1:

fnune:
[...]

PR #282, opened 2020-07-05
(feat) add documentation strings to hover response

fnune:
Fixes https://github.com/sveltejs/language-tools/issues/281
Partially fixes https://github.com/sveltejs/svelte/issues/5102 and https://github.com/sveltejs/language-tools/issues/280 (still need to produce a docstring for default exports)

For a snippet:

```typescript
<script>
    /** Documentation string for a */
    const a = true
</script>
```

Improves the hover contents from:

```typescript
const a: true
```

To:

```
```typescript\nconst a: true\n```\n---\nDocumentation string for a
```

Which should print out more or less like this:

> ```typescript
> const a: true
> ```
> ---
> Documentation string for a

If there is no documentation, the result is the same as before (for example, just `const a: true` in a Markdown TypeScript snippet).

Following the documentation [here](https://microsoft.github.io/language-server-protocol/specification#textDocument_hover).

### DIFF
--- a/packages/language-server/src/plugins/typescript/TypeScriptPlugin.ts
+++ b/packages/language-server/src/plugins/typescript/TypeScriptPlugin.ts
@@ -107,10 +107,19 @@ export class TypeScriptPlugin
         if (!info) {
             return null;
         }
-        const contents = ts.displayPartsToString(info.displayParts);
+        const declaration = ts.displayPartsToString(info.displayParts);
+        const documentation = typeof info.documentation === 'string'
+            ? info.documentation
+            : ts.displayPartsToString(info.documentation);
+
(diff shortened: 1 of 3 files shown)


fnune:
I don't know how to set up my editor to use my local version of the LSP so I haven't actually tested this "live".

dummdidumm:
Thanks for the PR! To test this, first uninstall the extension and then follow the steps documented [here](https://github.com/sveltejs/language-tools#running-the-dev-language-server-against-your-app)

fnune:
Good news, it's working just fine!

Some samples.

With a simple docstring:
image

With no docstring:
image

With a complex docstring:
Screenshot from 2020-07-05 19-30-31

dummdidumm:
Great stuff!

PR #284, opened 2020-07-05
(fix) Optional props

### LINKED ISSUE #276: Svelte typescript: Props with default values cause Type error of missing props
//component code
```
<script lang="typescript">
  export let name= '';
</script>
```

when component is added without providing the Prop "name", vs code shows this error
_component = Service_

```
Type '{ service: Service; }' is not assignable to type 'IntrinsicAttributes & { service: Service; className: string; }'.
  Property 'name' is missing in type '{ service: Service; }' but required in type '{ service: Service; name: string; }'
```

Currently I'm getting rid of this by adding explicit types 
```
<script lang="typescript">
  export let name: string | undefined = '';
</script>
```

we shouldn't have to do this right, can't we infer the type from the default value assignment?
or are there any other syntax (like '?') to make this more cleaner....
or am I doing something wrong?

PatrickG:
Determinate if prop is optional based on whether the variable has an initializer.

fixes #276

### DIFF
--- a/packages/svelte2tsx/src/svelte2tsx.ts
+++ b/packages/svelte2tsx/src/svelte2tsx.ts
@@ -378,7 +378,7 @@ function processInstanceScriptContent(str: MagicString, script: Node): InstanceS
         name: ts.BindingName,
         target: ts.BindingName = null,
         type: ts.TypeNode = null,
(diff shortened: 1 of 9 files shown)


dummdidumm:
👍

PR #285, opened 2020-07-05
(svelte2tsx) Allow documenting components with a tagged HTML comment

fnune:
### Goal

Solve https://github.com/sveltejs/language-tools/issues/280

For example:

```svelte
<!--
@component
Here's some documentation for this component. It will show up
on hover for TypeScript projects using a LSP-compatible editor
such as VSCode or Vim/Neovim with coc.nvim.

- You can use markdown here.
- You can use code blocks here.
-->

<main>
  <h1>
    Hello world
  </h1>
</main>
```

Single-line comments should work too. The contents of the HTML comment can then be picked up by the language service provider.

### Progress

- [x] Use named default exports instead of anonymous, using the name of the file for the exported class, in such a way that autocompletion doesn't end up broken.
  - [x] Maybe update the test suite to expect the more realistic case where `svelte2tsx` is called with a filename.
  - [x] Transform the filename into a `PascalCase` name for the default-exported class.
- [x] Implement the documentation feature with an HTML comment.
  - [x] Parse an HTML comment that looks like `<!-- @component Documentation string -->`.
  - [x] Take the string from the tag and add it to a JSDoc on the default (now named) export.
  - [x] Document this feature somewhere.

### About the test suite

With this, the test suite is now mostly unrealistic, since all the expected strings use anonymous exports. In most (if not all?) cases, `svelte2tsx` would be called with a filename. Maybe I should update all the tests.

### To be discussed

**Edit:** discussion solved, we go for `@component`.

Is this HTML comment approach the best? And if so, what should the tag look like? Here are some ideas:

- `@documentation`
- `@doc`
- `@component`: I like this one better because `@doc` or `@documentation` sounds a bit vague. `@component` specifies what the documentation is about, and it's obvious that it's documentation.

Maybe we could support multiple, too.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -43,6 +43,35 @@ It's also necessary to add a `type="text/language-name"` or `lang="language-name
 </style>
 ```
 
+## Documenting components
+
+To add documentation on a Svelte component that will show up as a docstring in
+LSP-compatible editors, you can use an HTML comment with the `@component` tag:
+
+```html
+<!--
+ @component
+ Here's some documentation for this component. It will show up on hover for
+ JavaScript/TypeScript projects using a LSP-compatible editor such as VSCode or
+ Vim/Neovim with coc.nvim.
+
+ - You can use markdown here.
+ - You can use code blocks here.
+ - JSDoc/TSDoc will be respected by LSP-compatible editors.
+ - Indentation will be respected as much as possible.
+-->
+
+<!-- @component You can use a single line, too -->
+
+<!-- @component But only the last documentation comment will be used -->
+
+<main>
+  <h1>
+    Hello world
+  </h1>
+</main>
+```
+
 ## Troubleshooting / FAQ
 
 ### Using TypeScript? See [this section](./preprocessors/typescript.md#troubleshooting--faq)
--- a/packages/svelte2tsx/src/svelte2tsx.ts
+++ b/packages/svelte2tsx/src/svelte2tsx.ts
@@ -1,3 +1,4 @@
+import { pascalCase } from 'pascal-case';
 import MagicString from 'magic-string';
 import path from 'path';
 import { parseHtmlx } from './htmlxparser';
@@ -787,8 +788,25 @@ function addComponentExport(
     str.append(statement);
 }
 
-function classNameFromFilename(filename: string): string {
-    return path.parse(filename).name;
+/**
+ * Returns a Svelte-compatible component name from a filename. Svelte
+ * components must use capitalized tags, so we try to transform the filename.
+ *
+ * Works with most filename styles except for filenames with multiple
+ * extensions, such as my-component.extension.svelte, which will result in
+ * MyComponentExtension.
+ *
+ * https://svelte.dev/docs#Tags
+ */
+export function classNameFromFilename(filename: string): string | undefined {
+    try {
+        const withoutExtension = path.parse(filename).name;
+        const inPascalCase = pascalCase(withoutExtension);
+        return inPascalCase;
+    } catch (error) {
+        console.warn(`Failed to create a name for the component class from filename ${filename}`);
+        return undefined;
+    }
 }
 
 function isTsFile(scriptTag: Node | undefined, moduleScriptTag: Node | undefined) {
--- a/packages/svelte2tsx/src/svelte2tsx.ts
+++ b/packages/svelte2tsx/src/svelte2tsx.ts
@@ -1,4 +1,5 @@
 import MagicString from 'magic-string';
+import path from 'path';
 import { parseHtmlx } from './htmlxparser';
 import { convertHtmlxToJsx } from './htmlxtojsx';
 import { Node } from 'estree-walker';
@@ -767,6 +768,8 @@ function addComponentExport(
     uses$$propsOr$$restProps: boolean,
     strictMode: boolean,
     isTsFile: boolean,
+    /** A named export allows for TSDoc-compatible docstrings */
+    name?: string,
 ) {
     const propDef =
         // Omit partial-wrapper only if both strict mode and ts file, because
@@ -777,10 +780,11 @@ function addComponentExport(
                 ? '__sveltets_with_any(render().props)'
                 : 'render().props'
             : `__sveltets_partial${uses$$propsOr$$restProps ? '_with_any' : ''}(render().props)`;
-    str.append(
-        // eslint-disable-next-line max-len
-        `\n\nexport default class {\n    $$prop_def = ${propDef}\n    $$slot_def = render().slots\n}`,
-    );
+
+    // eslint-disable-next-line max-len
+    const statement = `\n\nexport default class ${name ? `${name} ` : ''}{\n    $$prop_def = ${propDef}\n    $$slot_def = render().slots\n}`;
+
+    str.append(statement);
 }
 
 function isTsFile(scriptTag: Node | undefined, moduleScriptTag: Node | undefined) {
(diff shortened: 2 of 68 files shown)


jasonlyu123:
Wonder if we could allow jsdoc in this documentation block. Something like `@fires`, `@see`, `@prop` would be nice. However, some of it would have a different meaning.

something like this
```
<!--
  @component PaginationToolbar
  @props list list for pagination
  @see Pagination.md
-->
```

fnune:
@jasonlyu123 Yup, TSDoc is what you're supposed to write in it. It will be picked up by editors and rendered correctly. Unless I'm missimg something.

The additional `@component` tag would be just foe svelte2tsx to be able to pick it up.

jasonlyu123:
Haven't used to refer it with the new tsdoc standard ms is pushing lol

dummdidumm:
- About `@component` - good idea, I like that.
- About the test suite issue - I'm not too worried about that. It's tested indirectly in such a way that now every test file will have `export default class Input`.

fnune:
> Haven't used to refer it with the new tsdoc standard ms is pushing lol

@jasonlyu123 I see, sorry then, I misunderstood your initial post.

The implementation I'm going to make just passes along the string to the LSP. Beyond that, I don't know if any additional work needs to be done to support what you suggested.

fnune:
>     * About `@component` - good idea, I like that.
> 
>     * About the test suite issue - I'm not too worried about that. It's tested indirectly in such a way that now every test file will have `export default class Input`.

@dummdidumm Two points arise from that statement about the test suite:

- We would need to change the `svelte2tsx/test.js` file to support that. I'll do that as part of the first commit.
- You wrote `export default class Input`. I was thinking of respecting the casing the user created in their file, in this case it would be `input`. Do you think we should capitalize it?

dummdidumm:
> * We would need to change the `svelte2tsx/test.js` file to support that. I'll do that as part of the first commit.

Sounds good!

> * You wrote `export default class Input`. I was thinking of respecting the casing the user created in their file, in this case it would be `input`. Do you think we should capitalize it?

I think so, yes, the first letter should be capitalized. To me this seems like a standard in the JS/TS world. Classes are always first letter uppercase. Same for components in svelte, at least from what I have seen.

fnune:
Yup, I've also always written component names capitalized. The problem is the user could be using snake case or something else. Maybe I could bring in some utility to transform any case to `PascalCase`.

dummdidumm:
Yes, good point, such transformations should be done.

PatrickG:
AFAIK components needs to have an uppercase first letter to be rendered anyway (except `svelte:component`).

fnune:
Alright! This is ready for review. I hope you don't find it too controversial :)

I've tested it locally and it's working just fine. Maybe you can find some edge case?

packages/svelte2tsx/src/svelte2tsx.ts
```diff
@@ -53,12 +58,20 @@ type pendingStoreResolution<T> = {
     scope: Scope;
 };
 
+/**
+ * Add this tag to a HTML comment in a Svelte component and its contents will
+ * be added as a docstring in the resulting JSX for the component class.
+ */
+const COMPONENT_DOCUMENTATION_HTML_COMMENT_TAG = '@component' as const;
```
jasonlyu123, line 65: Is it necessary to have const assertions here?

fnune: My intention was to make it read-only. Maybe it wasn't nexessary, but I don't think it hurts either. WDYT?

dummdidumm: I would remove it. A const is already `readonly` by definition and the type is also infered as `@component` I think.

packages/svelte2tsx/src/svelte2tsx.ts
```diff
@@ -777,10 +823,34 @@ function addComponentExport(
...
+
+    str.append(statement);
+}
+
+/**
+ * Returns a Svelte-compatible component name from a filename. Svelte
+ * components must use capitalized tags, so we try to transform the filename.
+ *
+ * Works with most filename styles except for filenames with multiple
+ * extensions, such as my-component.extension.svelte, which will result in
```
dummdidumm, line 840: We could fix this by doing `name.split('.').pop()`. What do you think? Does this fail in cases I don't know yet?

fnune: I'm not sure... honestly I don't think we can get it 100% right here, other than by recommending users to use `PascalCase` in their file names.

Maybe `name.split('.')[0]` is a good transformation because it will drop the extra extensions, which I doubt anybody will ever want in their class name...

packages/svelte2tsx/src/svelte2tsx.ts
fnune: I'm learning this after the fact:

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions

What this `as const` would prevent is unwanted type widening. I think in this file it doesn't really matter, and if it's causing confusion, I would rather remove it.

packages/svelte2tsx/src/svelte2tsx.ts
dummdidumm: I agree. This would come after the `pascalCase`-transformation.

packages/svelte2tsx/src/svelte2tsx.ts
dummdidumm: Note that in they docs they use `let bla = 'somestring'`. For let it is indeed typed as `string`, because you could change it, but `const` cannot be reassigned so TS should infer the exact string as a type.

packages/svelte2tsx/src/svelte2tsx.ts
fnune: `pascalCase` actually will transform `thing.thang` into `ThingThang`, so it needs to go before, I think.

dummdidumm: Since the comment is now outdated, could you update it, too?

fnune: Done, I added some docs in the last commit. Those can be reviewed too, and then this is ready to go.

packages/svelte2tsx/src/svelte2tsx.ts
fnune: I removed the superfluous assertion now.

docs/README.md
```diff
@@ -43,6 +43,35 @@ It's also necessary to add a `type="text/language-name"` or `lang="language-name
...
+## Documenting components
+
+To add documentation on a Svelte component that will show up as a docstring in
+LSP-compatible editors, you can use an HTML comment with the `@component` tag:
+
+```svelte
+<!--
+ @component
+ Here's some documentation for this component. It will show up
+ on hover for TypeScript projects using a LSP-compatible editor
```
dummdidumm, line 55:
```suggestion
 on hover for JavaScript/TypeScript projects using a LSP-compatible editor
```

docs/README.md
```diff
@@ -43,6 +43,35 @@ It's also necessary to add a `type="text/language-name"` or `lang="language-name
 </style>
 ```
 
+## Documenting components
+
+To add documentation on a Svelte component that will show up as a docstring in
+LSP-compatible editors, you can use an HTML comment with the `@component` tag:
+
+```svelte
```
dummdidumm, line 51: Would be great if Github would support svelte syntax, but it doesn't as if now ^^ 

```suggestion
```html
```

fnune: Awh. Alright, pushed it again :)

dummdidumm:
🎉

jasonlyu123:
Work like a champ with #301 lol
圖片

fnune:
Awesome :smile:

Issue #279, opened 2020-07-04
Support for array of preprocessors


kingdaro:
**Describe the bug**
`import type` syntax causes a type error. This is needed for ESM-sensitive bundlers that need to differentiate types from values for erasure during compilation

**To Reproduce**

```js
// svelte.config.js
const preprocess = require("svelte-preprocess")

module.exports = {
	preprocess: [
		preprocess.postcss(),
		preprocess.typescript({ transpileOnly: true }),
	],
}
```
```ts
// types.ts
export type Secret = string
```
```svelte
<!-- Component.svelte -->
<script lang="ts">
  import type { Secret } from './types' // type error on this line
  const secret: Secret = '42'
</script>
```

For example a code snippet that is treated in a way you don't expect.

**Expected behavior**
Type imports should be parsed without errors

**Screenshots**
![image](https://user-images.githubusercontent.com/19603573/86521865-9f0c4f80-be1b-11ea-9de5-25fb05124b9c.png)

**System (please complete the following information):**
 - OS: Windows 10
 - IDE: VSCode
 - Plugin/Package: Svelte for VSCode

dummdidumm:
I think it's because you use an array of preprocessors, that is not supported yet.
If you refactor your `svelte.config.js` to 

```js
const preprocess = require("svelte-preprocess")

module.exports = {
	preprocess: preprocess({
                postcss: true,
		typescript: { transpileOnly: true }
        }),
}
```
does it work then?

kingdaro:
That indeed fixes it, thanks 😅 maybe a note in the docs would be helpful?

dummdidumm:
Absolutely

kevmodrome:
What would be needed for this to be supported? Willing to help any way I can! Not having this feature makes it impossible to use MDSveX and sass together, for example

dummdidumm:
I think @kaisermann wanted to take a shot at this, not sure what his status on that is though.

kaisermann:
@dummdidumm I've stopped what I was doing a while ago, but will probably get back to it tomorrow or this weekend!

PR #286, opened 2020-07-05
Add note in documentation about preprocess array

kingdaro:
Adding an array causes the extension to work improperly (https://github.com/sveltejs/language-tools/issues/279), so this note should save some confusion


dummdidumm:
Thank you! 😃

PR #291, opened 2020-07-08
(feat) support default language of svelte-preprocess

dummdidumm:
`svelte-preprocess` has a feature for setting default languages so you don't have to write `lang="x"` each time. This adds support for this in style/script tags.

To accomplish this, parsing of the `svelte.config.js` was moved up to `Document`. Inside it, the `lang` attribute is set to the default language if there exists no `lang`/`type` attribute.

I deliberately did not bump the `svelte-preprocess` version yet to wait until `4.x` has received some more feedback and bugfixes.


dummdidumm:
I did not change any dependencies and now this happens ... ? oO

Issue #294, opened 2020-07-08
TypeScript Errors ts(2440) with Svelte for VS Code starting v101.1.0


vatro:
**Describe the bug**
Yesterday with v101.0.0 I finished migrating my Svelte components library to TS with almost no / no severe (correctly compiling) TS-errors. Today with **v101.1.1** (checked: starting with v101.1.0) I get a bunch of **ts(2440)** (+ a lot of resulting) errors: **Import declaration conflicts with local declaration of 'Foo'**, while everything is still compiling / building correctly as with v101.0.0. I'm not sure though if this is a bug or a feature which was not functioning correctly in v101.0.0.

Running `svelte-check` reports same errors.

**To Reproduce**
I'm  going to describe as simple as possible how I'm building the mentioned library.

It's a library of Svelte components 'target_lib' wrapping modules of library 'source_lib'. Svelte wrapper-components and their corresponding modules from 'source_lib' have identical names:

```javascript
// Svelte wrapper-component 'Foo.svelte'
<script lang="typescript">
   import { Foo } from 'source_lib'
</script>

// 'target_lib.ts'
export { default as Foo } from './components/Foo.svelte'
export * from 'source_lib'
```

When building 'target_lib' Rollup is renaming duplicates like this in the 'target_lib.mjs':

`Foo` from 'source_lib' keep name `Foo`
Svelte wrapper-`Foo`-components are being renamed to `Foo_1` (+ _1)

This might seem like bad practice but is exactly how I want it to be, it works as intended: Svelte 'target_lib'-'Foo' components override (and utilize) corresponding 'source_lib'-'Foo' components when e.g.

```javascript
//Test.svelte
<script>
   import { Foo } from 'target_lib'
</script>

<Foo />
```

**Expected behavior**
I don't know what to expect, because I don't know if ts(2440) errors are a bug or a feature. Rollup builds o.k. as expected / as before despite the "new" errors.

Actually it feels like correct behavior that these conflicts are being reported, but on the other hand it makes Rollup's ability to manage such cases obsolete and then it feels kind of wrong again. I would just like to get a clear response if this is intended / correct behavior, then I would know if I have to think of another solution / fix / find workaround or wait for an update. Thanks!

**System (please complete the following information):**
 - OS: Windows 10
 - IDE: VSCode
 - Plugin/Package:
Svelte for VSCode
```json
 "devDependencies": {
    "@rollup/plugin-alias": "^3.1.1",
[...]

PR #297, opened 2020-07-09
(fix) change svelte2tsx' default export

dummdidumm:
Append `__SvelteComponent_` to make it unlikely that name clashes occur.
#294

Due to this most of the simplifications in #293 had to be reverted.


vatro:
cool! 😎 Thank you for the super-fast fix! 🚀🎉

fnune:
Just curious, does this also make it so that the autocompletion for `MyComponent` would end up importing `import MyComponent_Svelte... from 'files/MyComponent'`?

fnune:
Oook nevermind, I just saw you're removing it here:

https://github.com/sveltejs/language-tools/pull/297/files#diff-fa2252ad7bea78e1b88fecce4db7af08R294

PR #303, opened 2020-07-12
(feat) (svelte2tsx) Type check event forward from element on component

jasonlyu123:
Aims to be able to type-check events bubbled/forwarded from an element. https://svelte.dev/docs#on_element_event
Part of #304 

Component.svelte
```svelte
<button on:click></button>
```

Component2.svelte
```svelte
<scirpt>
import Component2 from './Component2.svelte
</script>

<Component2 on:click={e => console.log(e)} /> <!--- type check here -->
```

- [x] Component event type check spec
- [x] Definition of element event forward in the component class
- [x] New Component event type check method
- [x] Component forward event again


jasonlyu123:
Can't get it to extract event type from `$on` method, Seems like we might still need to add a def property like `$$props_def` and `$$slot_def`

packages/svelte2tsx/svelte-shims.d.ts
```diff
@@ -49,8 +58,23 @@ declare function __sveltets_restPropsType(): SvelteRestProps
 declare function __sveltets_partial<T>(obj: T): Partial<T>;
 declare function __sveltets_partial_with_any<T>(obj: T): Partial<T> & SvelteAllProps
 declare function __sveltets_with_any<T>(obj: T): T & SvelteAllProps
-declare function __sveltets_store_get<T=any>(store: SvelteStore<T>): T
+declare function __sveltets_store_get<T = any>(store: SvelteStore<T>): T
 declare function __sveltets_any(dummy: any): any;
 declare function __sveltets_empty(dummy: any): {};
 declare function __sveltets_componentType(): AConstructorTypeOf<SvelteComponent>
 declare function __sveltets_invalidate<T>(getValue: () => T): T
+declare function __sveltets_eventDef<T extends SvelteEventRecord>(def: T): SvelteOnEvent<T>
```
dummdidumm, line 66: I get a ts error here: `Generic type 'SvelteOnEvent' requires 2 type argument(s)`

```suggestion
declare function __sveltets_eventDef<T extends SvelteEventRecord, K extends keyof T>(def: T): SvelteOnEvent<T, K>
```

jasonlyu123: miss this during rename. Should be `SvelteOnAllEvent<T>`

dummdidumm: Does this also fix the "cannot use custom events"-bug I described?

jasonlyu123: yeah :grin:

Issue #310, opened 2020-07-15
Regression: reactive variables inside brackets raise an error


dkzlv:
**Describe the bug**
Reactive variables appear to be unassigned to the compiler, if you define them in a reactive block with curly braces, both in template and in script tag.

**To Reproduce**
```html
<script lang="ts">
  export let check: boolean;

  let derived: string;
  $: {
    derived = check.toString();
  }

  // Variable 'derived' is used before being assigned.
  console.log(derived.toUpperCase())
</script>

<!-- Variable 'derived' is used before being assigned. -->
{derived.toUpperCase()}
```

**Expected behavior**
Shouldn't report any errors here.

**System (please complete the following information):**
OS: Mac OS 10.15.3
IDE: VS Code
Plugin/Package: Svelte for VS Code (the new official one)

**Additional context**
Used to work a week ago. It works fine if you change the syntax to `$: derived = ...` without the brackets.
Also, `svelte-check` doesn't report any errors here.

jasonlyu123:
the one in the script is correct. I'll see what we can do about the one in the template

dkzlv:
Same goes for other cases like this:

```javascript
let sliceStart: number, sliceEnd: number;
$: [sliceStart, sliceEnd] = [page * limit, page * limit + limit];
```

It also says these two variables are used before assigning.

jasonlyu123:
This raises another problem
```
let d = 1;
$: a = d;
console.log(a)
```
The `console.log(a)` is also used before assigning this should also show an error

https://svelte.dev/repl/1001568f2498473bbfb665129876259b?version=3.24.0

jasonlyu123:
Let's track the used before declare in a separate issue

PR #314, opened 2020-07-16
(svelte2tsx) wrap template with callback

jasonlyu123:
fix for #310 and better union type for template
```svelte
<script>
let a: 'a' | 'b' = 'a';

const handler = () => {
  a = 'b'
}
</script>

{ a } <!-- a will have type 'a' | 'b' now -->
```


jasonlyu123:
Since it'll affect the "most-discussed-problem", the reactive reactive-declarations, This will be blocked until a better solution for both problems.

jasonlyu123:
Umm... Why are there so many merging conflicts when the edit is on the different line lol

PR #318, opened 2020-07-17
added TSX style color to svelte component tags.

BlackFenix2:
i modified the grammar file using support.class.component.tsx to append any tag starting with a capital letter with TSX style colors to separate svelte components from html tags. the yarn test script passes without issues.

granted, i might want to double-check the intellisense in case i added tsx commands by accident. if i can narrow the scope to just affect the color alone this will be perfect.

image


jasonlyu123:
Great!🤩

PR #335, opened 2020-07-23
type check for reactive declaration

gfreezy:
```js
let a = 1;
$: b = a
```

generates

```javascript
let a = 1;

;let b = a; $: b = __sveltets_invalidate(() => a);

```

so typescript can infer the type of `b` to be `number` instead of `any`.

### DIFF
--- a/packages/svelte2tsx/src/svelte2tsx.ts
+++ b/packages/svelte2tsx/src/svelte2tsx.ts
@@ -392,7 +392,7 @@ function processInstanceScriptContent(str: MagicString, script: Node): InstanceS
     const astOffset = script.content.start;
     const exportedNames: ExportedNames = new Map();
 
-    const implicitTopLevelNames: Map<string, number> = new Map();
+    const implicitTopLevelNames: Map<string, {pos: number; expr: ts.Expression}> = new Map();
     let uses$$props = false;
     let uses$$restProps = false;
 
@@ -760,7 +760,10 @@ function processInstanceScriptContent(str: MagicString, script: Node): InstanceS
             ) {
                 const name = node.statement.expression.left.text;
                 if (!implicitTopLevelNames.has(name)) {
-                    implicitTopLevelNames.set(name, node.label.getStart());
+                    implicitTopLevelNames.set(name, {
+                        pos: node.label.getStart(),
+                        expr: node.statement.expression.right
+                    });
                 }
 
                 wrapExpressionWithInvalidate(node.statement.expression.right);
@@ -786,9 +789,10 @@ function processInstanceScriptContent(str: MagicString, script: Node): InstanceS
     pendingStoreResolutions.map(resolveStore);
 
     // declare implicit reactive variables we found in the script
-    for (const [name, pos] of implicitTopLevelNames.entries()) {
+    for (const [name, {pos, expr}] of implicitTopLevelNames.entries()) {
         if (!rootScope.declared.has(name)) {
-            str.prependRight(pos + astOffset, `;let ${name}; `);
+            const exprStr = str.slice(astOffset + expr.getStart(), astOffset + expr.getEnd());
+            str.prependRight(pos + astOffset, `;let ${name} = ${exprStr}; `);
         }
     }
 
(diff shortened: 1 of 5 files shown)


jasonlyu123:
We originally considering just replace `$:` label to `let`. the generated code would be smaller than copying the initializer. 

Another problem with the copying initializer approach would be the invalidation of the typescript control flow.
```ts
declare function __sveltets_invalidate<T>(getValue: () => T): T
let a: 1 | 2 = 1;

let b = a; b = __sveltets_invalidate(() => a);
```

https://www.typescriptlang.org/play/?ssl=4&ssc=47&pln=1&pc=1#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXwH0CBnANxAgxA2IK1VKgi2CioB4AVAPgAoBzagDUmyEAC54vAJTwAvN3idpkzgCgI1eFEkBGeAB94AJnnxdAbjUatAIzNQL8e3MIlylarXqNmrKrwy8opQ0lZAA

this would result in some unwanted typescript errors

Nonetheless, I do agree with it infer the type from the initializer, It just needed some adjustment. But since this is the most discussed problem with svelte tooling and typescript I'll let it stay here a few days for more discussion.

gfreezy:
```typescript
declare function __sveltets_invalidate<T>(getValue: () => T): T
let a: 1 | 2 = 1;

let b = __sveltets_invalidate(() => a);
```

This seems to infer the right type.

https://www.typescriptlang.org/play/?ssl=1&ssc=1&pln=5&pc=1#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXwH0CBnANxAgxA2IK1VKgi2CioB4AVAPgAoBzagDUmyEAC54vAJTwAvN3idpkzgCgI1eFEkBGeAB94AJnnxdAbjUatAIzNEyFKjToMmLNiF4z5iqNJWQA

gfreezy:
```typescript
declare function __sveltets_invalidate<T>(getValue: () => T): T
let a: 1 | 2 = 1;

let b = __sveltets_invalidate(() => a);
```

This seems to infer the right type.

https://www.typescriptlang.org/play/?ssl=1&ssc=1&pln=5&pc=1#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXwH0CBnANxAgxA2IK1VKgi2CioB4AVAPgAoBzagDUmyEAC54vAJTwAvN3idpkzgCgI1eFEkBGeAB94AJnnxdAbjUatAIzNEyFKjToMmLNiF4z5iqNJWQA

dummdidumm:
I second the comment to change the output to

```ts
let b = __sveltets_invalidate(() => a);
```

Other than that I do not see any problems with this approach.

gfreezy:
@dummdidumm @jasonlyu123 I've updated the code to 

```ts
let b = __sveltets_invalidate(() => a);
```

packages/svelte2tsx/src/svelte2tsx.ts
```diff
@@ -789,10 +786,17 @@ function processInstanceScriptContent(str: MagicString, script: Node): InstanceS
...
     // declare implicit reactive variables we found in the script
-    for (const [name, {pos, expr}] of implicitTopLevelNames.entries()) {
+    for (const [name, pos] of implicitTopLevelNames.entries()) {
         if (!rootScope.declared.has(name)) {
-            const exprStr = str.slice(astOffset + expr.getStart(), astOffset + expr.getEnd());
-            str.prependRight(pos + astOffset, `;let ${name} = ${exprStr}; `);
+            // remove '$:' label
+            str.remove(pos + astOffset, pos + astOffset+2);
+            const followingChar = str.slice(pos + astOffset+2, pos + astOffset+3);
+            // if there exists whitespace after '$:', we do not insert more space.
```
dummdidumm, line 794: Not sure if we need this. Is there any advantage over just doing it either way?

gfreezy: we dont care about the format of the generated code？

dummdidumm: We don't need to, we can do all the bad formatting we want :-P

dummdidumm:
Awesome!

Issue #339, opened 2020-07-23
Custom Directive Support in VSCode: Partial Hydration for SSG


nickreese:
Hi, I'm working on a full fledged SSG that is designed to be Svelte's answer to Gatsby. It is designed for SEO focused, mainly static sites that have some highly interactive components.

We're already using this SSG in production at https://elderguide.com but are working to make it more user friendly before releasing it.

The key feature that has been blocking me from using Svelte 100% for templating is partial hydration.

Thanks to the help from @kev on Discord, we've come up with a pretty elegant solution, but I can't get VS code to accept the custom directive.

### Example of usage:

```
<!-- Parent -->
<script>
  import HydrateMe from './HydrateMe.svelte';

  const lotsOfData = {
    stuff: true,
    name: 'Nick',
    email: 'yep@yep.com',
    notThisOne: [1, 2, 3, 4],
  };
</script>

<!-- problemsome syntax -->
<HydrateMe interactive:data={{ name: lotsOfData.name, email: lotsOfData.email }} />

<!-- Hydrate Me -->
<script>
  export let name;
  export let email;
</script>

<div>{name} {email}</div>
```

### Custom Directive:

`<HydrateMe interactive:data={{ name: lotsOfData.name, email: lotsOfData.email }} />`

### svelte.config.js:

```
const partialHydration = require('./partialHydration');

module.exports = {
  preprocess: partialHydration,
};
```

### partialHydration.js

```
module.exports = {
  markup: async ({ content, filename }) => {
    const matches = content.matchAll(/<([a-zA-Z]+)\s+interactive:data={(.*)}/gim);

    for (const match of matches) {
      console.log('match', match);
      const componentName = match[1];
      const dataFunction = match[2];

      const replacement = `<div class="needs-hydration" data-component="${componentName}" data-data={JSON.stringify(${dataFunction})}`;
      content = content.replace(match[0], replacement);
    }

    return { code: content };
  },
};
```

## Under the Hood

To be clear, how this is working is the SSR version of the component is built with the preprocessor. Then when we parse what is returned from the SSR component, we have a small wrapper which adds a new root component when it finds `<div class="needs-hydration" ...`

The data found in `data-compontent` is added as the props on the new root component. 

## Questions

1 -- Is there a way I can get VS code to accept this custom directive? Here is the error I'm seeing after restarting VS code to make sure the language server had the latest svelte.config. 

[...]

PR #340, opened 2020-07-23
(Svelte2txs) custom:directive to custom_directive

nickreese:
So this is my first time playing with a language tool. 

The main problem is that the TS checker is choking on any sort of custom:directive as outlined in #339.

This PR simply replaces colons in attribute names with an underscore to get TS to validate them.


nickreese:
The more I'm looking at this, the more I don't believe this is a good solution. Svelte natively offers preprocessing, the language tools just need to support it and I'm not seeing where the markup is being preprocessed in `htmlxparser.ts`

PR #344, opened 2020-07-24
(svelte-vscode) Add initial support for multiple preprocessors

### LINKED ISSUE #279: Support for array of preprocessors

kaisermann:
An initial stab at supporting multiple preprocessors in the `svelte.config.js` file.

Closes #279

### DIFF
--- a/packages/language-server/src/plugins/svelte/SvelteDocument.ts
+++ b/packages/language-server/src/plugins/svelte/SvelteDocument.ts
@@ -1,5 +1,5 @@
+import { InternalPreprocessorGroup } from './../../lib/documents/configLoader';
 import { SourceMapConsumer } from 'source-map';
-import { PreprocessorGroup } from 'svelte/types/compiler/preprocess';
 import type { compile } from 'svelte/compiler';
 import { CompileOptions } from 'svelte/types/compiler/interfaces';
 import { Processed } from 'svelte/types/compiler/preprocess';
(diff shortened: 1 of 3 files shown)


packages/language-server/src/plugins/svelte/SvelteDocument.ts
```diff
@@ -1,5 +1,5 @@
+import { InternalPreprocessorGroup } from './../../lib/documents/configLoader';
```
kaisermann, line 1: I'm not exactly liking this 🤔

packages/language-server/src/plugins/svelte/SvelteDocument.ts
```diff
@@ -262,46 +265,59 @@ export class SvelteFragmentMapper {
...
+            // we don't want to modify the original preprocessor
+            const wrapperPreprocessor = { ...originalPreprocessor };
+
+            if (originalPreprocessor.script) {
+                wrapperPreprocessor.script = async (args: any) => {
+                    try {
+                        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+                        const res = await originalPreprocessor.script!(args);
+                        if (res.map) {
+                            processedScript = res;
```
dummdidumm, line 288: This (and `processedStyle` as well) needs to be an array to which the map or undefined is pushed, so that later we can create an array of source mappers.

dummdidumm:
Finished in #384

PR #346, opened 2020-07-24
fix style svg attribute

### LINKED ISSUE #345: svelte-check complaints against type of correct style attribute on SVG elements
**Describe the bug**

svelte-check complaints against type of correct style attribute on SVG elements.
Example:

```
<svg
    width={size}
    height={size}
    viewBox="0 0 {iconSvg.viewbox} {iconSvg.viewbox}"
    class={`icon ${status} ${$$props.class || ''}`}
    style={$$props.style || null}>
    <g>
      {#each iconSvg.paths as p}
         <path style={p.color || color ? `fill:${p.color || color}` : undefined}
               d={p.path} />
      {/each}
    </g>
</svg>
```
Here `string` is the actual type of `style` attribute of `path` element.
We have the following `svelte-check` error:

```
Error: Type 'string | undefined' is not assignable to type 'CSSProperties | undefined'.
  Type 'string' is not assignable to type 'CSSProperties | undefined'. (ts)
    <path style={p.color 
```

I suppose this caused by incorrect definition of `SVGAttributes` (svelte2tsx/svelte-jsx.d.ts)
```
    interface SVGAttributes<T> extends DOMAttributes<T> {
      // Attributes which also defined in HTMLAttributes
      ....
      name?: string;
      style?: CSSProperties;  <---
      target?: string;
      type?: string;
      width?: number | string;
```

galkatz373:
Fixes #345 
I changed the style attribute for svg element from CSSProperties to string | CSSProperties, as HTMLAttributes interface.

### DIFF
--- a/packages/svelte2tsx/svelte-jsx.d.ts
+++ b/packages/svelte2tsx/svelte-jsx.d.ts
@@ -2147,7 +2147,7 @@
       method?: string;
       min?: number | string;
       name?: string;
-      style?: CSSProperties;
+      style?: CSSProperties | string;
       target?: string;
       type?: string;
       width?: number | string;


jasonlyu123:
I think we can remove `CSSProperties`. It doesn't work. But we can leave the interface there, doesn't hurt.
https://svelte.dev/repl/84c05f0e49104782ae8998dde51b906a?version=3.24.0

PR #350, opened 2020-07-24
Fix style svg

galkatz373:
#345 
Remove CSSProperties from the style of the svg, and replace it with a string.


jasonlyu123:
Can you also remove the CSSProperties from element?

galkatz373:
Do we even need the interface? i'm not sure the purpose of it

jasonlyu123:
I'm ok with remove the interface

dummdidumm:
Yeah we should remove it, this is react-specific stuff.

dummdidumm:
Thanks! The shim is a whole lot smaller now, I like!

PR #351, opened 2020-07-25
Toggle JSX namespace based on svelte vs svelte-native

halfnelson:
This brings Svelte Native support to language tools.
Since svelte and svelte-native have different intrinsic elements, it was essential to namespace them (JSX -> svelte.JSX)
The namespace is chosen by typescript based on the jsxFactory compiler option. I have changed this to be either svelte.createElement or svelteNative.createElement to select either svelte.JSX or svelteNative.JSX namespace.

The code tries to detect svelte-native by resolving the package. The namespace can be forced by providing jsxFactory compiler option in your tsconfig.

I had to change a couple of tests to handle the new test folder I added, and to handle the namespace change.
I think the language server will work with old versions of svelte2tsx (as the namespace falls back to JSX) but an old language-server release against a new svelte2tsx will probably show unwanted TS errors

### DIFF
--- a/packages/language-server/src/plugins/typescript/features/RenameProvider.ts
+++ b/packages/language-server/src/plugins/typescript/features/RenameProvider.ts
@@ -127,7 +127,7 @@ export class RenameProviderImpl implements RenameProvider {
         if (
             !renameInfo.canRename ||
             renameInfo.kind === ts.ScriptElementKind.jsxAttribute ||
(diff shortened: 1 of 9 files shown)


dummdidumm:
Great stuff! Nice side effect: This also fixes #228, no more clashes with react types because it's now regarded as a different factory.

PR #359, opened 2020-07-25
Rename `ondoubleclick` to `ondblclick` to match the MDN ref and not React

### LINKED ISSUE #357: Rename s/doubleclick/dblclick ?
**Describe the bug**

`doubleclick` is not the standard name for "double click" events, [MDN ref](https://developer.mozilla.org/en-US/docs/Web/API/Element/dblclick_event).

Though React uses `doubleclick`, the entire layer of synthetic event-ing allows them to do those kind of renames. AFAIK, there is no such thing in Svelte so we should stick with the ref and update those 2 occurrences:

https://github.com/sveltejs/language-tools/blob/6a457a1db05faf43ea4f374fe573ef454a4de6c5/packages/svelte2tsx/src/knownevents.ts#L98

https://github.com/sveltejs/language-tools/blob/5cd71dc6d62b41adf07e436d8e0a54cffdc34133/packages/svelte2tsx/svelte-jsx.d.ts#L1906

**To Reproduce**
https://svelte.dev/repl/af7e3e3f854d40fe998bd7c016bd3f8e?version=3.24.0

zowesiouff:
Fixes #357 

https://developer.mozilla.org/en-US/docs/Web/API/Element/dblclick_event
https://github.com/sveltejs/language-tools/issues/357


dummdidumm:
Thanks! 😃

PR #364, opened 2020-07-26
(svelte-vscode) Fix trigger character on '#'

galkatz373:
The '#' character  was missing in the wordPattern regex in extension.ts.
So now the autocomplete on '#' works.


dummdidumm:
Thank you! A regression we did not catch.

PR #365, opened 2020-07-27
fixes #5192: SvelteAction should allow void as return value - https:/…

opensas:
refers to https://github.com/sveltejs/svelte/issues/5192

This change allows to specify an action like this (just like you can do using js):

```typescript
const focusOnInit = (node) => node.focus()
```

instead of:

```typescript
const focusOnInit = (node) => { node.focus(); return {} }
```


dummdidumm:
Thanks!

PR #368, opened 2020-07-27
Improve Completions

galkatz373:
Related to my previous pull #364.
This pull further improve completions for 'if', 'each', 'await' blocks.
Added an optional insertText to completions, in order to bring more completions flexibility.
The insertText could be typed as a plain text or as a snippet format.

### DIFF
--- a/packages/language-server/src/plugins/svelte/features/getCompletions.ts
+++ b/packages/language-server/src/plugins/svelte/features/getCompletions.ts
@@ -58,8 +58,13 @@ function getCompletionsWithRegardToTriggerCharacter(
             { tag: 'each', label: 'each', insertText: 'each $1 as $2}\n\t$3\n{/each' },
             {
                 tag: 'await',
-                label: 'await',
+                label: 'await :then',
                 insertText: 'await $1}\n\t$2\n{:then $3} \n\t$4\n{/await',
+            },
+            {
+                tag: 'await',
+                label: 'await then',
+                insertText: 'await $1 then $2}\n\t\t$3\n{/await',
             },
         ]);
     }
--- a/packages/language-server/src/plugins/svelte/features/getCompletions.ts
+++ b/packages/language-server/src/plugins/svelte/features/getCompletions.ts
@@ -4,10 +4,11 @@ import {
     CompletionList,
     CompletionItemKind,
     CompletionItem,
+    InsertTextFormat,
 } from 'vscode-languageserver';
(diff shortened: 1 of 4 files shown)


packages/language-server/src/plugins/svelte/features/getCompletions.ts
```diff
@@ -54,9 +54,13 @@ function getCompletionsWithRegardToTriggerCharacter(
...
         return createCompletionItems([
-            { tag: 'if', label: 'if' },
-            { tag: 'each', label: 'each' },
-            { tag: 'await', label: 'await' },
+            { tag: 'if', label: 'if', insertText: 'if $1}\n\t$2\n{/if' },
+            { tag: 'each', label: 'each', insertText: 'each $1 as $2}\n\t$3\n{/each' },
+            {
+                tag: 'await',
+                label: 'await',
+                insertText: 'await $1}\n\t$2\n{:then $3} \n\t$4\n{/await',
```
dummdidumm, line 62: Altough this is super nice, there is another possibility to write the `await` construct: `{#await expression then name}...{/await}`. Maybe add another `await` completion with that other `insertText`?

packages/language-server/test/plugins/svelte/features/getHoverInfo.test.ts
```diff
@@ -66,7 +66,7 @@ describe('SveltePlugin#getHoverInfo', () => {
         });
 
     describe('should return hover for #', () => {
-        (['if', 'each', 'await'] as const).forEach((tag) => {
+        (['if', 'each', 'await :then', 'await then'] as const).forEach((tag) => {
```
dummdidumm, line 69: I would leave the test as-is, I think it's unrelated and your changes should not affect this. Instead you need to adjust the method signature in `getCompletions`.

galkatz373: what's the alternative?

galkatz373: Putting two different completions with the same label would be a bad idea.

dummdidumm: These tests are for hover info, not completions.

galkatz373: oh right, I'm stupid

dummdidumm:
Awesome!

Issue #347, opened 2020-07-24
Optional props: Type is not assignable to type 'IntrinsicAttributes & ...'.


dominic-simplan:
**Describe the bug**
I have a component like this:
```
<script lang="ts">
    export let width = "100%";
    export let height = "100px";

    export let responsive = false;
</script>

<div id="categorySelect" style="width: {width}; height: {height};" />
```

If I use this component from another component but **do not define the width**: `<Abc height="100%" responsive={true} />`, VSCode shows the following error:
```
Type '{ height: string; responsive: true; }' is not assignable to type 'IntrinsicAttributes & { width: string; height: string; responsive: boolean; }'.
  Property 'width' is missing in type '{ height: string; responsive: true; }' but required in type '{ width: string; height: string; responsive: boolean; }
```

Therefore I tried explicitly defining width as string | undefined:

```
<script lang="ts">
    export let width: string | undefined = "100%";
    export let height = "100px";

    export let responsive = false;
</script>

<div id="categorySelect" style="width: {width}; height: {height};" />
```

Now I get this error:
`Type 'true' is not assignable to type 'false'`
It complains because I am setting another value than the default value.

Most strangely, **adding an event handler** for an event which doesn't even exist **fixes the errors**:
`<Abc height="100%" on:bla={e => {}} responsive={true} />`

**Expected behavior**
It should be possible to skip properties with default value or use other values than the default value.

**System (please complete the following information):**
 - OS: Windows 10
 - IDE: VSCode 1.47.2
 - Plugin/Package: "Svelte for VSCode 101.3.0"
```
"svelte": "^3.24.0",
"svelte-check": "^0.1.55",
"svelte-loader": "^2.13.6",
"svelte-preprocess": "^4.0.8",
"@tsconfig/svelte": "^1.0.3",
"typescript": "^3.9.7",
```

tsconfig:
```
{
    "extends": "@tsconfig/svelte/tsconfig.json",
    "include": [
        "src/**/*",
        "src/node_modules"
    ],
    "exclude": [
        "node_modules/*",
        "dist/*"
    ],
    "compilerOptions": {
        "strict": true,
        "resolveJsonModule": true,
        "sourceMap": true
    }
}
```

**Additional context**
This is so strange, maybe something with my setup is wrong?
Related: #276

jasonlyu123:
Don't worry you just encounter a ton of problems at once.
1. props without type annotation don't got annotate as optional in strict mode.
[...]

PR #370, opened 2020-07-27
(fix) mark props as optional in ts when no typings given

dummdidumm:
Before this, `export let a = ''` would not be marked as optional because it had no type.
#347

Also fixes a bug where file type was incorrectly inferred as jsx which occurs when user uses the default languages feature of svelte-preprocess. Instead of computing the file type inside svelte2tsx, it is now passed as an option.


packages/svelte2tsx/src/svelte2tsx.ts
```diff
@@ -614,11 +614,7 @@ function processInstanceScriptContent(str: MagicString, script: Node): InstanceS
         ts.forEachChild(list, (node) => {
             if (ts.isVariableDeclaration(node)) {
                 if (ts.isIdentifier(node.name)) {
-                    if (node.type) {
```
dummdidumm, line 617: @halfnelson do you remember why you made this distinction? I removed it to enable optional props, but I fear I may have overlooked something else.

PR #384, opened 2020-07-30
(feat) Support for multiple preprocessors

dummdidumm:
#279

Built from the initial work in #344

### DIFF
--- a/packages/language-server/src/lib/documents/DocumentMapper.ts
+++ b/packages/language-server/src/lib/documents/DocumentMapper.ts
@@ -82,9 +82,7 @@ export class IdentityMapper implements DocumentMapper {
     }
 
     destroy() {
-        if (this.parent?.destroy) {
-            this.parent.destroy();
-        }
+        this.parent?.destroy?.();
     }
 }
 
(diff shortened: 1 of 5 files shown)


packages/language-server/src/lib/documents/DocumentMapper.ts
```diff
@@ -51,23 +51,41 @@
...
         return true;
     }
 
     getURL(): string {
         return this.url;
     }
+
+    destroy() {
+        if (this.parent?.destroy) {
+            this.parent.destroy();
```
jasonlyu123, line 86: Maybe this?
```
this.parent?.destroy?.();
```

packages/language-server/src/plugins/svelte/SvelteDocument.ts
```diff
@@ -208,11 +210,20 @@ export class SvelteFragmentMapper {
...
-        return new SvelteFragmentMapper(
-            { end: -1, diff: 0 },
-            new IdentityMapper(originalDoc.uri),
-            new IdentityMapper(originalDoc.uri),
-            sourceMapper,
+        return null;
+    }
+
+    private static async createSourceMapper(processed: Processed[], originalDoc: Document) {
+        return await processed.reduce(
```
kaisermann, line 217: `return await` in an `async` method is usually redundant

```suggestion
        return processed.reduce(
```

PR #388, opened 2020-07-30
fix: Added missing event handlers

bfanger:
Event handlers that are working, like `<svelte:window on:hashchange={handler} />` are triggering typescript errors:

 <img width="759" alt="is not assignable to type 'HTMLProps<Window> & SvelteWindowProps'." src="https://user-images.githubusercontent.com/207248/88932046-dab10280-d27d-11ea-84cd-fdfd5580ae78.png">

I've extracted the available events from Chrome with: `JSON.stringify(Object.keys(window).filter(k=>k.substr(0,2) ==='on'))` and removed events that didn't have a type like `Window['onsearch']`or are deprecated/non-standard like `onmousewheel`


packages/svelte2tsx/svelte-jsx.d.ts
```diff
@@ -250,7 +291,65 @@
...
+      onpagehide?: PageTransitionEventHandler<T>;
+      onpagehidecapture?: PageTransitionEventHandler<T>;
+      onpageshow?: PageTransitionEventHandler<T>;
+      onpageshowcapture?: PageTransitionEventHandler<T>;
+
+      // Storage Events
+      onstorage?: StorageEventHandler<T>;
+      onstoragecapture?: StorageEventHandler<T>;
+
+      // Global / Window Events
```
jasonlyu123, line 333: This part is also covered in #394. But his generic parameter is correct. Maybe you can remove this part and I'll merge both.

packages/svelte2tsx/svelte-jsx.d.ts
```diff
@@ -250,7 +291,65 @@
       // Transition Events
       ontransitionend?: TransitionEventHandler<T>;
       ontransitionendcapture?: TransitionEventHandler<T>;
-    }
+
+      // HashChange Events
```
jasonlyu123, line 295: Can you add a new interface for the window specific events? like `WindowEventHandlersAttributes` and make the `SvelteWindowProps` `extends` it, Or you can just move these events To `SvelteWindowProps` interface. These events don't apply to `HTMLElement`. 
Also, I think we don't need capture-suffix events. The existing are adopted from the react's definition.

bfanger: I've removed the capture suffix and and moved the events into the SvelteWindowProps interface

bfanger:
Hmm, my editor formatted causing conficts, i'll clean up the commits, one moment..

bfanger:
Done, removed (and tested) the unnecessary on*capture event handlers, also synced the knownevents.ts and added onfullscreenchange & onfullscreenerror (I checked all events on div element)

packages/svelte2tsx/svelte-jsx.d.ts
```diff
@@ -700,6 +663,25 @@
       scrollX?: Window['scrollX'];
       scrollY?: Window['scrollY'];
       readonly online?: Window['navigator']['onLine'];
+
+
+      onbeforeprint?: EventHandler<Window>;
```
dummdidumm, line 668: I think all `EventHandler<Window>` should be `EventHandler<Event, Window>` instead.

jasonlyu123: Yeah, that's what I means about the problem with generic parameter

bfanger: Fixed

PR #401, opened 2020-07-31
Update typescript.md to include vscode absolute path import info

elialbert:
some advice about getting vscode to allow absolute imports and not just relative ones. This little gotcha took me a while.


dummdidumm:
👍

Issue #398, opened 2020-07-31
[Docs] Update docs to indicate that the Svelte Language Server will respect defaults and not require explicit "lang" or "type" tags.


shirakaba:
On each of these documentation pages, it is implied that hints such as `lang="ts"` or `lang="scss"` are required to infer the language inside a `<script>` or `<style>` tag.

* https://github.com/sveltejs/language-tools/tree/master/docs#using-with-preprocessors
* https://github.com/sveltejs/language-tools/blob/master/docs/preprocessors/typescript.md
* https://github.com/sveltejs/language-tools/blob/master/docs/preprocessors/scss-less.md

It is now clear (as confirmed by @dummdidumm on Discord) that the Svelte Language Server will respect any defaults specified in `svelte.config.js`, so it should be clarified that this is no longer a necessity as long as those defaults are defined.

dummdidumm:
While this is true for the LSP, it is not 100% true for syntax highlighting. For this the `lang` attribute is still needed. I updated the docs accordingly.

Issue #397, opened 2020-07-31
[Docs] Document how to set up svelte.config.js with respect to rollup.config.js


shirakaba:
Trying to get the Svelte Language Server set up in VS Code for the first time, onboarding could have been smoother. I wanted to use TS and SCSS, so I followed the docs:

* https://github.com/sveltejs/language-tools/tree/master/docs#using-with-preprocessors
* https://github.com/sveltejs/language-tools/blob/master/docs/preprocessors/typescript.md
* https://github.com/sveltejs/language-tools/blob/master/docs/preprocessors/scss-less.md

Each of these pages mentioned a need for a `svelte.config.js`. It wasn't clear whether the `svelte.config.js` config needed to be written any differently from that of the Svelte Rollup Plugin in `rollup.config.js` (especially with respect to the `production` variable).

My understanding from @dummdidumm on Discord is that the Svelte Language Server is only (currently) interested in the `preprocess` export from `svelte.config.js`, so there's no need to write the whole config for the Svelte Rollup Plugin in there.

So I've gone with the following inheritance pattern. `svelte.config.js` exports both an instance of `sveltePreprocess` for consumption by Svelte Language Server, and also the base config for it (`preprocessOptions`) so that we can extend it when we create another `sveltePreprocess` instance in `rollup.config.js`:

## `svelte.config.js`

```javascript
const sveltePreprocess = require('svelte-preprocess');

const preprocessOptions = {
    sourceMap: true, // "you would always want sourcemaps for the IDE" – dummdidumm
    defaults: {
        script: "typescript",
        style: "scss",
    },
    scss: {
        prependData: `@import 'src/styles/_variables.scss';`
    },
    postcss: {
        plugins: [require('autoprefixer')()]
    }
};

module.exports = {
    preprocess: sveltePreprocess(preprocessOptions),

    // Export this to allow rollup.config.js to inherit the same preprocess options.
    preprocessOptions,
}
```

## `rollup.config.js `

```javascript
// ...

const preprocessOptions = require("./svelte.config").preprocessOptions;
const production = !process.env.ROLLUP_WATCH;

export default {
    // ...

    plugins: [
        // ...
        svelte({
            // enable run-time checks when not in production
            dev: !production,

            // we'll extract any component CSS out into
            // a separate file - better for performance
            css: css => {
                css.write('public/build/bundle.css');
            },
            preprocess: sveltePreprocess({
                ...preprocessOptions,
                sourceMap: !production,
            }),
        }),
[...]

PR #405, opened 2020-08-01
(docs) better preprocessor docs

dummdidumm:
#397 , #398 

@shirakaba mind taking a look if that matches your expectations?

### DIFF
--- a/docs/preprocessors/in-general.md
+++ b/docs/preprocessors/in-general.md
@@ -1,0 +1,102 @@
+# Using Preprocessors
+
+## Generic setup
+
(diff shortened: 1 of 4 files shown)


docs/preprocessors/in-general.md
```diff
@@ -0,0 +1,100 @@
...
+```js
+// svelte.config.js - NOTE: you cannot use the new "import x from y" and "export const" syntax in here.
+const sveltePreprocess = require('svelte-preprocess');
+
+module.exports = {
+    preprocess: sveltePreprocess(),
+};
+```
+
+It's also necessary to add a `type="text/language-name"` or `lang="language-name"` to your `style` and `script` tags, which defines how that code should be interpreted by the extension.
```
shirakaba, line 16: Maybe worth mentioning in this line that default languages can also be configured (with details in a section below), so that readers don’t jump to conclusions upon reading “It’s also necessary to add ...”.

jasonlyu123:
My understanding is that the syntax highlight for specific language would only work when lang/type is specified. It that's the case maybe we should also mention that.

dummdidumm:
You are totally right, this would break syntax highlighting in some cases. We should mention that.

docs/preprocessors/in-general.md
dummdidumm: I totally forgot about syntax highlighting which @jasonlyu123 pointed out. Because of this, I'll keep it as is to not "advertise" this as something this would work for everyone, which it will not (at least syntax highlighting wise).

PR #406, opened 2020-08-01
(feat) support for message events

caroso1222:
PR https://github.com/sveltejs/language-tools/pull/388 misses the [`onmessage` event](https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onmessage) for window and lacks good types for the `onbeforeunload` event. This PR takes care of that.


dummdidumm:
Thanks!

PR #411, opened 2020-08-02
(fix) allow empty initializationOptions

endenwer:
Allow to start the language server with empty `initializationOptions`. Right now it throws an error 
`Cannot read property 'dontFilterIncompleteCompletions' of null (Internal Error)`


dummdidumm:
Nice catch, thanks!

Issue #404, opened 2020-07-31
Language server does not start properly with --stdio flag of svelteserver


yaegassy:
## Describe the bug

I mainly use `Vim` and am trying to use `svelteserver` using the LSP client plugin called `coc.nvim`.

coc.nvim has a very nice dedicated extension the `coc-svelte`.

coc.nvim can also be used to set up a language server without the need for a dedicated extension.

I found this when I set up svelteserver without a dedicated extension, and it starts fine with the `--node-ipc` flag, but `--stdio` can't start the server.

By the way, coc.nvim itself has a proven track record with many language servers, so there should be no problems.

## Screenshots

Sorry for the "coc.nvim example", but I've attached a screen capture.

**[NG] DEMO**

coc.nvim settings (coc-settings.json):

```json
    "svelte-language-server": {
      "command": "svelteserver",
      "args": ["--stdio"],
      "filetypes": ["svelte"]
    },
```

![svelte-language-server-issue-1](https://user-images.githubusercontent.com/188642/89078736-ea386480-d3bf-11ea-93fd-6706eed7cb5b.gif)

## Misc

Do you know of any other editors or IDEs that work properly with the `--stdio` flag?

If you have one, I'd appreciate it if you could let me know so I can reference it.

If the `--stdio` flag is not working properly, it may be difficult to integrate with other editors/IDEs.

yaegassy:
@jasonlyu123 @dummdidumm 

This [PR](https://github.com/sveltejs/language-tools/pull/420) here confirmed that it works properly with `--stdio`. Thanks for the fix.

Please publish the latest version of `svelte-language-server` to npm so that installation is easy.

jasonlyu123:
@yaegassy it'll be automatically update at around 4: 10 utc that about 30mins from now

yaegassy:
@jasonlyu123 Thanks publish to npm!

PR #420, opened 2020-08-04
Disable console.log since it break content-length

mattn:
#404

### DIFF
--- a/packages/language-server/src/server.ts
+++ b/packages/language-server/src/server.ts
@@ -61,7 +61,9 @@ export interface LSOptions {
 export function startServer(options?: LSOptions) {
     let connection = options?.connection;
     if (!connection) {
-        console.log = function() {};
+        console.log = (...args: any[]) => {
+            console.warn(...args);
+        };
         connection = process.argv.includes('--stdio')
             ? createConnection(process.stdin, process.stdout)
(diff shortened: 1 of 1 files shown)


dummdidumm:
Oh, this is the cause of the issue? Glad you found that!
Could you elaborate a bit on what exactly you mean by "it breaks content-length"? Is there maybe another way than to just disable logging?

For the fix itself we would also need to disable `console.error` / `console.warning` I guess? Also I'd rather not just patch the console in every environment, just for `stdio`. A long-term / better alternative would be to find something that also works in stdio.

mattn:
`--stdio` read constent from stdin and write content to stdout with content-length headers. And `console.log` also write to stdout. If you keep console.log, wrap console.log to write the content to stderr.

```
console.log = function() { console.warn.apply(console, arguments) }
```

dummdidumm:
Hints:
- You can run `yarn lint` in the workspace to see the lint errors before pushing
- Please only do the `console.log` patching if the `stdio` command is set:
```ts
if (process.argv.includes('--stdio')) {
   console.log = ...
}
```

mattn:
Sorry about noisy.

dummdidumm:
No problem 😄 I'll just sqash it when merging.

dummdidumm:
Thank you very much!

Issue #416, opened 2020-08-03
Allow for actions to be typed specific to the node they're applied to


yagebu:
**Is your feature request related to a problem? Please describe.**
I use actions a lot and have many actions that are specific for some node types (they might for example only make sense for an SVGGElement). However, currently the following fails since the SvelteAction expects the first parameter to be just an `HTMLElement`:

```
<script lang="ts">
    function action(node: HTMLInputElement): {} {
        console.log(node.size)  // just an example of something that wouldn't type-check on HTMLElement
        return {}
    }
</script>
<input use:action />
```
This is the error:
```
Argument of type '(node: HTMLInputElement) => {}' is not assignable to parameter of type 'SvelteAction<[]>'.
  Types of parameters 'node' and 'node' are incompatible.
    Type 'HTMLElement' is missing the following properties from type 'HTMLInputElement': accept, align, alt, autocomplete, and 49 more.
```

**Describe the solution you'd like**
It would be nice if the typing for `use:...` would take the type of the node that the action is applied to into account.

**Describe alternatives you've considered**
Of course I could type-check using `instanceof` inside my action to have everything typecheck, but that would be more verbose and would incur a minimal but unnecessary runtime cost.

yagebu:
That was quick :) Thank you very much @dummdidumm!

PR #421, opened 2020-08-04
(feat) better typings for actions

dummdidumm:
#416

### DIFF
--- a/packages/svelte2tsx/src/htmlxtojsx.ts
+++ b/packages/svelte2tsx/src/htmlxtojsx.ts
@@ -96,11 +96,11 @@ export function convertHtmlxToJsx(
         }
     };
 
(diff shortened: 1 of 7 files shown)


packages/svelte2tsx/src/htmlxtojsx.ts
```diff
@@ -103,7 +97,11 @@ export function convertHtmlxToJsx(
     };
 
     const handleActionDirective = (attr: Node) => {
```
jasonlyu123, line 99: `(attr: Node, parent: Node) =>`
I think we can pass the parent of the action as parameter it would be the element.

https://astexplorer.net/#/gist/4c18623d74b7906a0126287db9d78162/1e59c75a386605eaa9ca3c54fbbdbd3cbbb75aca

PR #433, opened 2020-08-07
(fix) bubble events

dummdidumm:
Added new property `$$events_def` to bubble events
Also removed some obsolete types

Now correctly infers the bubbled event type and falls back to any if it cannot be inferred.

@jasonlyu123 while removing the obsolete types I noticed that you typed the event record as `type SvelteEventRecord = Record<string, Event | Event[]>`. Is there a reason why you typed this as "event or array of events"?

### DIFF
--- a/packages/svelte2tsx/src/nodes/event-handler.ts
+++ b/packages/svelte2tsx/src/nodes/event-handler.ts
@@ -8,7 +8,8 @@ export function createEventHandlerTransformer() {
 
         const handleEventHandlerBubble = () => {
             const componentEventDef = `__sveltets_instanceOf(${parent.name})`;
-            const exp = `__sveltets_bubbleEventDef(${componentEventDef}.$on, '${eventName}')`;
+            // eslint-disable-next-line max-len
+            const exp = `__sveltets_bubbleEventDef(${componentEventDef}.$$events_def, '${eventName}')`;
 
             const exist = events.get(eventName);
             events.set(eventName, exist ? [].concat(exist, exp) : exp);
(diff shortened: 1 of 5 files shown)


jasonlyu123:
That's for if there are multiple sources of the event. Paving the path for the union type of component event bubble. For now, it should all be the same type.

dummdidumm:
I don't quite get the idea yet. Could you post a small code sample?

jasonlyu123:
```
<Child1 on:whatever />
<Child2 on:whatever />
```

if somehow these event has different type

dummdidumm:
Oooh okay, I get it now.
I think the current solution has the possibility to be enhanced to support something like this, so I would leave it as it is now - or do you have a different opinion?

packages/svelte2tsx/src/nodes/event-handler.ts
```diff
@@ -47,17 +45,9 @@ function getEventDefExpressionForNonCompoent(eventName: string, ele: Node) {
...
 }
 
-function eventMapEntryToString([eventName, expression]: [
-    string,
-    string | string[]
-]) {
-    return `'${eventName}':${
-        Array.isArray(expression) ? `[${expression}]` : expression
-        }`;
+function eventMapEntryToString([eventName, expression]: [string, string | string[]]) {
```
jasonlyu123, line 51: I think we can add back the `SvelteExtractEvent` type alias because of this is still possibly return array

dummdidumm: Good catch! I solved it in another way by adding a `unionType` method.

PR #435, opened 2020-08-09
Fix paths on Windows

mattn:
On Windows, definition does not work when svelte-language-server is started with `--stdio`. This is because that vscode-uri always use lower-case drive letter. vscode is also use vscode-uri. So it always works correctly. But Language Server Client may send request  with upper-case drive letter.

This change add normalizeUrl to fix URLs.

### DIFF
--- a/packages/language-server/src/server.ts
+++ b/packages/language-server/src/server.ts
@@ -29,7 +29,7 @@ import {
     SveltePlugin,
     TypeScriptPlugin,
 } from './plugins';
-import { urlToPath } from './utils';
+import { urlToPath, normalizeUrl } from './utils';
 
 namespace TagCloseRequest {
     export const type: RequestType<
@@ -77,7 +77,7 @@ export function startServer(options?: LSOptions) {
     }
 
     const docManager = new DocumentManager(
-        (textDocument) => new Document(textDocument.uri, textDocument.text),
+        (textDocument) => new Document(normalizeUrl(textDocument.uri), textDocument.text),
     );
     const configManager = new LSConfigManager();
     const pluginHost = new PluginHost(docManager, configManager);
@@ -191,10 +191,10 @@ export function startServer(options?: LSOptions) {
 
     connection.onDidOpenTextDocument((evt) => {
         docManager.openDocument(evt.textDocument);
-        docManager.markAsOpenedInClient(evt.textDocument.uri);
+        docManager.markAsOpenedInClient(normalizeUrl(evt.textDocument.uri));
     });
 
-    connection.onDidCloseTextDocument((evt) => docManager.closeDocument(evt.textDocument.uri));
+    connection.onDidCloseTextDocument((evt) => docManager.closeDocument(normalizeUrl(evt.textDocument.uri)));
     connection.onDidChangeTextDocument((evt) =>
         docManager.updateDocument(evt.textDocument, evt.contentChanges),
     );
--- a/packages/language-server/src/utils.ts
+++ b/packages/language-server/src/utils.ts
@@ -1,6 +1,8 @@
 import { URI } from 'vscode-uri';
 import { Position, Range } from 'vscode-languageserver';
 
+const isWindows = process.platform === 'win32';
+
 export function clamp(num: number, min: number, max: number): number {
     return Math.max(min, Math.min(max, num));
 }
@@ -14,7 +16,15 @@ export function urlToPath(stringUrl: string): string | null {
 }
 
 export function pathToUrl(path: string) {
-    return URI.file(path).toString();
+    const url = URI.file(path).toString();
+    // On Windows, first %3A should be part of drive name.
+    return isWindows ? url.replace(/%3A/, ':') : url;
+}
+
+export function normalizeUrl(stringUrl: string) {
+    // vscode-uri use lower-case drive letter
+    // https://github.com/microsoft/vscode-uri/blob/95e03c06f87d38f25eda1ae3c343fe5b7eec3f52/src/index.ts#L1017
+    return isWindows ? stringUrl.replace(/file:\/\/\/[A-Z]:/, (s) => s.toLowerCase()) : stringUrl;
 }
 
 export function flatten<T>(arr: T[][]): T[] {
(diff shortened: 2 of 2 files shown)


dummdidumm:
Thanks for bringing this up! Could you elaborate a bit more where this is a problem? I did not quite understand it yet.

I'm not sure about the solution though. In my tests with the extension I got an URI from VSCode like `file:///c%3A/...`, so your regex-replace with `:` will likely not work. I think the better solution would be to just use `URI.parse(uri).toString()` -> this will lowercase the letter, too, while making sure we use the same implementation as VSCode does.
Another thing is that you did not catch all cases where this is used, I think it's better to do the transformation inside `DocumentManager`.

mattn:
When using non-vscode editor, it may send request with upper-case drive letter. So the path does not match paths generated by server. Also drive letter in URL should not escaped. For example, gopls does not escape it.

mattn:
I'm okay to fix that translate in DocumentManaager. but I will do it tonight.

dummdidumm:
Thanks for the explanation.

About the drive letter `:` not getting escaped: It seems like VSCode-URI does this. When I do `URI.parse('file:///C:/...').toString()` I get back `file:///c%3A/...`. Haven't tested it, but if we change that behavior I fear it may break the existing extension.

mattn:
But some Language Server Client send request with `file:///c:/foo/bar` .

dummdidumm:
But is it really a problem for them if they get back `file:///c%3A/foo/bar`? Since this seems to be the correct encoding (I guess VSCode-URI conforms to the standard), I would expect clients to be able to handle this.

mattn:
svelte-language-server have to normalize URLs with one of two ways.

* translate `file:///D:/` from client to `file:///d%3A/` for server
* translate `file:///d%3A/` in server for `file:///d:/` and translate `file:///D:/` from client  to `file:///d:/` for server.

dummdidumm:
I would prefer the first option since we can then use `URI.parse('file:///C:/...').toString()`

mattn:
So the patch for utils.js should be:
```diff
diff --git a/packages/language-server/src/utils.ts b/packages/language-server/src/utils.ts
index 561520d..0e17c66 100644
--- a/packages/language-server/src/utils.ts
+++ b/packages/language-server/src/utils.ts
@@ -1,6 +1,8 @@
 import { URI } from 'vscode-uri';
 import { Position, Range } from 'vscode-languageserver';
 
+const isWindows = process.platform === 'win32';
+
 export function clamp(num: number, min: number, max: number): number {
     return Math.max(min, Math.min(max, num));
 }
@@ -17,6 +19,15 @@ export function pathToUrl(path: string) {
     return URI.file(path).toString();
 }
 
+export function normalizeUrl(stringUrl: string) {
+    // vscode-uri use lower-case drive letter
+    // https://github.com/microsoft/vscode-uri/blob/95e03c06f87d38f25eda1ae3c343fe5b7eec3f52/src/index.ts#L1017
+    if (!isWindows) {
+        return stringUrl;
+    }
+    return stringUrl.replace(/^file:\/\/\/([a-zA-z]):/, (s: string, m: string) => 'file:///' + m.toLowerCase() + '%3A');
+}
+
 export function flatten<T>(arr: T[][]): T[] {
     return arr.reduce((all, item) => [...all, ...item], []);
 }
```

mattn:
I seems the normalizeUrl will be called in Document.ts and DocumentManager.ts . Right?

mattn:
`url` in constructor of Document is not used?

mattn:
Ah, sorry. It is property. I was confused.

dummdidumm:
What about doing `URI.parse(uri).toString()` instead of your patch? It feels a little too low-level for me and it seems to me that vscode-uri would handle both the lowercase and the `:`->`%3A` transformation.

mattn:
You still prefer the transformation inside DocumentManager and Document?

packages/language-server/src/server.ts
```diff
@@ -191,10 +191,10 @@ export function startServer(options?: LSOptions) {
...
     connection.onDidOpenTextDocument((evt) => {
         docManager.openDocument(evt.textDocument);
-        docManager.markAsOpenedInClient(evt.textDocument.uri);
+        docManager.markAsOpenedInClient(normalizeUrl(evt.textDocument.uri));
     });
 
-    connection.onDidCloseTextDocument((evt) => docManager.closeDocument(evt.textDocument.uri));
+    connection.onDidCloseTextDocument((evt) => docManager.closeDocument(normalizeUrl(evt.textDocument.uri)));
     connection.onDidChangeTextDocument((evt) =>
         docManager.updateDocument(evt.textDocument, evt.contentChanges),
```
dummdidumm, line 199: This one also contains an URI

packages/language-server/src/server.ts
```diff
@@ -191,10 +191,10 @@ export function startServer(options?: LSOptions) {
 
     connection.onDidOpenTextDocument((evt) => {
         docManager.openDocument(evt.textDocument);
```
dummdidumm, line 193: This one also contains an URI

dummdidumm:
Not so sure anymore about doing it inside `Document(Manager)`. I added two comments about other possible URI occurences. There might also be more in the rename request and so on. Generally speaking this needs some thorough search in `server.ts` which of all these events could have an URI inside.

On a side note, I wonder why this is not handled inside the `vscode-languageserver` package.

mattn:
Most easy way to reproduce is to send requests with lower-case drive letters. But vscode-uri is also used in Visual Studio Code, you can't reproduce this with vscode.

I use Vim and install vim-lsp, vim-lsp-settings. If you don't have know-how of Vim or vim plugins, It will be hard to setup.

dummdidumm:
I reproduced it by adding a `normalizeUri` function but doing the opposite, uppercasing the drive letter. I get now what your problem is. Basically nothing is working because the URIs don't match those of the TypeScript language service.

During this I concluded that it's probably best to do the URI normalization inside the `DocumentManager`. This is where the connection to the "outside world"/client happens.

mattn:
I don't mind you will close this and fixes this issue by your-self (If it is best). My best hope is only that this issue will be fixed. :)

dummdidumm:
Alright, having a look at fixing this tomorrow 😃

dummdidumm:
Fixed through #443

PR #437, opened 2020-08-10
(feat) be able to type props/events/slots

dummdidumm:
Through a new reserved interface `ComponentDef`

See new tests for examples.

### Generics 
Problem to be solved: Generics. We cannot use `instanceOf(ComponentName)` anymore, because this way we cannot pass in the information about the type of the generic.

If we have 

Child:
```html
<script lang="ts">
   interface ComponentDef<T> {
        props: {items: T[]},
        events: {itemClick: CustomEvent<T>;}
        slots: {}
   }
</script>
// ...
```
Parent:
```html
<Child items="{['a', 'stringarray']}" on:itemClick="{iShouldBeOfTypeString => ...}" />
```

We need a transformation that makes sure that `iShouldBeOfTypeString` actually is inferred as `string`. My current idea is to transform this to `new Child({props: {...allPropsAndItsValuesThatAreDefinedOnTheChild}}).$on` instead of the old `instanceOf(Child).$on`. Thoughts?

### DIFF
--- a/packages/svelte2tsx/src/interfaces.ts
+++ b/packages/svelte2tsx/src/interfaces.ts
@@ -1,12 +1,14 @@
 import MagicString from 'magic-string';
 import { Node } from 'estree-walker';
 import { ExportedNames } from './nodes/ExportedNames';
+import ts from 'typescript';
 
 export interface InstanceScriptProcessResult {
     exportedNames: ExportedNames;
     uses$$props: boolean;
     uses$$restProps: boolean;
     getters: Set<string>;
+    componentDef: ts.InterfaceDeclaration | undefined;
 }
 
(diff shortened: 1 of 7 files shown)


jasonlyu123:
Maybe we can add a dummy helper function like `__sveltets_norename` or `__sveltets_generated` to retain the no-rename of old `instanceOf`

Evertt:
Hey guys, any idea when this will be completed? 👼 
I'm running up against this very issue right now and for now I'm just using `any` instead of a generic type, but you know, a generic type would be nice. 😁

dummdidumm:
This implementation will very likely not land in this form. [This RFC](https://github.com/sveltejs/rfcs/pull/38) is where the discussion is happening and the details are specified. After that the implementation will follow.

PR #438, opened 2020-08-10
support sapper-noscroll

jasonlyu123:
[see discord](https://discordapp.com/channels/457912077277855764/689494103380983930/742379516701442140)

In the sapper runtime check, as long as there is the attribute it would affect so I type it as `true` instead of `boolean`
https://github.com/sveltejs/sapper/blob/6918586aaf7f6e8e525e7869335f2186ec75df2b/runtime/src/app/start/index.ts#L113

### DIFF
--- a/packages/svelte2tsx/src/htmlxtojsx/index.ts
+++ b/packages/svelte2tsx/src/htmlxtojsx/index.ts
@@ -355,13 +355,14 @@ export function convertHtmlxToJsx(
         //if we are on an "element" we are case insensitive, lowercase to match our JSX
         if (parent.type == 'Element') {
             //skip Attribute shorthand, that is handled below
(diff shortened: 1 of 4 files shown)


packages/svelte2tsx/src/htmlxtojsx/index.ts
```diff
@@ -355,13 +355,14 @@ export function convertHtmlxToJsx(
         //if we are on an "element" we are case insensitive, lowercase to match our JSX
         if (parent.type == 'Element') {
             //skip Attribute shorthand, that is handled below
+            const sapperNosSroll = attr.name === 'sapper:noscroll';
```
dummdidumm, line 358: minor typo, it should be `sapperNoScroll` instead of `sapperNosSroll`

PR #439, opened 2020-08-10
(fix) awaited type of promise or plain value

yagebu:
When awaiting the value of a type like Promise<string> | null, the type
wouldn't match PromiseLike. Hence the awaited value would still be
Promise<string> | null and not string | null. Reducing it to one type
declaration and using a conditional type to infer the awaited type
should do the trick.


dummdidumm:
Makes sense 👍  Thank you!

PR #443, opened 2020-08-11
(fix) normalize uris coming from client

dummdidumm:
#435


dummdidumm:
@mattn after the next deployment of the `svelte-language-server` npm package in about 20 hours you can check if the fix worked. Let me know!

mattn:
I confirmed your fix works fine on my enviroment. Thank you.

Issue #444, opened 2020-08-11
Unexpected Error when Destructing $store Values 


JoeDailey:
**Describe the bug**
The statement `$: ({ value } = $store);` is valid syntax when `value` is a property of `$store` and `$store` is a Svelte store. It behaves like normal destructuring and is reactive to store updates. This produces an error when `lang="ts"`.
```
# App.svelte
------------
<script lang="ts">
  import data from "./store";
  $: ({ count } = $data);
</script>
```
```
jd:/svelte-playground $ yarn svelte-check                                                                                                                                                                                                                      [0:30]
yarn run v1.22.4
warning package.json: No license field
$ /home/jd/Projects/svelte-playground/node_modules/.bin/svelte-check

Loading svelte-check in workspace: /home/jd/Projects/svelte-playground
Getting Svelte diagnostics...
====================================

/home/jd/Projects/svelte-playground/src/App.svelte:3:9
Error: No value exists in scope for the shorthand property 'count'. Either declare one or provide an initializer. (ts)
;
  $: ({ count } = $data

====================================
svelte-check found 1 error and 0 warnings
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

**To Reproduce**
1) Start with this REPL: [link](https://svelte.dev/repl/1b09be146b1a44b396495187f1ba0bc5?version=3.24.1)
2) Download it and add TS support to the project: [link](https://svelte.dev/blog/svelte-and-typescript)
3) Modify `App.svelte` by adding `lang="ts"` to the opening `<script>` tag. (see the above snippet)
4) run `svelte-check`

**Expected behavior**
This is valid Svelte/JS syntax and his code runs fine. It should not produce typechecker errors. 

**Screenshots**
![image](https://user-images.githubusercontent.com/3579057/89870646-016b1500-db6b-11ea-9eb2-a1b525a339fc.png)

**System (please complete the following information):**
 - OS: Windows
 - IDE: VSCode (Windows-Linux Subsystem)
 - Plugin/Package:
   - Svelte for VSCode
   - svelte-check
   - @tsconfig/svelte

JoeDailey:
**Workaround**
Don't use destructuring on Svelte stores in Typescript. This is not ideal but not a blocker. It is just less concise when destructuring several properties at once.
REPL: [link](https://svelte.dev/repl/57d3c037cc984d6d91d80b495fc22150?version=3.24.1))
```
# App.svelte
------------
[...]

PR #445, opened 2020-08-11
(fix) destructuring inside $

dummdidumm:
#444

TODOs:
- [x] don't remove the `( )` around the expression if there is no transformation from `$: ...` to `let ...`
- [x] remove the `( )` in such a way that the inner transformations are preserved and not removed as well

### DIFF
--- a/packages/svelte2tsx/src/nodes/ImplicitTopLevelNames.ts
+++ b/packages/svelte2tsx/src/nodes/ImplicitTopLevelNames.ts
@@ -4,7 +4,34 @@ import MagicString from 'magic-string';
 export class ImplicitTopLevelNames {
     private map = new Map<string, ts.LabeledStatement>();
 
-    add(identifier: ts.Identifier, node: ts.LabeledStatement, astOffset: number, str: MagicString) {
+    add(
+        binaryExpr: ts.BinaryExpression,
+        node: ts.LabeledStatement,
+        rootVariables: Set<string>,
+        astOffset: number,
+        str: MagicString,
+    ) {
+        if (ts.isIdentifier(binaryExpr.left)) {
+            this.addSingle(binaryExpr.left, node);
+        } else if (ts.isObjectLiteralExpression(binaryExpr.left)) {
+            this.getPropsOfObjectLiteral(binaryExpr.left).map((prop) =>
+                this.addSingle(prop.name, node),
+            );
+
+            if (
+                ts.isExpressionStatement(node.statement) &&
+                ts.isParenthesizedExpression(node.statement.expression) &&
+                this.objectLiteralContainsNoTopLevelNames(binaryExpr.left, rootVariables)
+            ) {
+                const start = node.statement.expression.getStart() + astOffset;
+                str.remove(start, start + 1);
+                const end = node.statement.expression.getEnd() + astOffset - 1;
+                str.remove(end, end + 1);
+            }
+        }
+    }
+
+    private addSingle(identifier: ts.Identifier, node: ts.LabeledStatement) {
         const name = identifier.text;
 
         // svelte won't let you create a variable with $ prefix anyway
@@ -12,17 +39,6 @@ export class ImplicitTopLevelNames {
 
         if (!this.map.has(name) && !isPotentialStore) {
             this.map.set(name, node);
-        }
-
-        // TODO don't do this if the variable inside the expression is defined explicetly
-        if (
-            ts.isExpressionStatement(node.statement) &&
-            ts.isParenthesizedExpression(node.statement.expression)
-        ) {
-            const start = node.statement.expression.getStart() + astOffset;
-            str.remove(start, start + 1);
-            const end = node.statement.expression.getEnd() + astOffset - 1;
-            str.remove(end, end + 1);
         }
     }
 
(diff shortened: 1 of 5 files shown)


jasonlyu123:
Also
```
let a = [1, 2, 3]
$: [ first ] = a
```

it would be nice if we can find the equivalent of this line in typescript
https://github.com/sveltejs/svelte/blob/f739b4772a4ad32c0fd337afb9d5594ebd776073/src/compiler/compile/Component.ts#L607

I looked into `typescript.d.ts` 'MemberExpression' seems to be the base interface of four other expressions in typescript

dummdidumm:
Good point, forgot about them. If you like you can take over, not going to work in this today anymore.

tanhauhau:
you can use periscopic#extract_names https://github.com/Rich-Harris/periscopic#extract_identifiers-and-extract_names to get all the identifier names within the assignment

dummdidumm:
Oh that's a nice way of getting it, but we are using the typescript AST here so we cannot use this.

jasonlyu123:
I think we could just recreate that function for typescript AST. Another option would be using typescript-eslint parser to transform AST to estree compliant but that seems a little bit too overkill

jasonlyu123:
When looking into the solution, I found another issue. Because we're transforming this:
```
$: a = b
```
to 
```
let a = b
```
to infer type for a. So when one of the destructured identifiers is already defined but others don't, how do we transform it to do an assignment as well as declaration.

like this:
```
let a: number;
$: ({ a, c } = b);
```

dummdidumm:
This situation came to my mind, too, and my current solution for it is "that's such a corner case, just don't try to infer the type" 😄 
So I would just transform it to
```
let a; // <- no type
$: ({a, c} = b);
```

jasonlyu123:
Oh! I just thought like "what about this situation you wrote?" 😂

dummdidumm:
Oooh you mean a different situation, I get it now. What if the user has defined one of them ... well yes, we somehow would have to add the other `let`. Maybe like this:

User:
```
let a: number;
$: ({a,b} = c);
```
Gets transformed to:
```
let a: number;
let b;
$: ({a,b} = c);
```

Sorry for the confusion 😄

dummdidumm:
So to summarize the cases:

**No let x; by user**

Input
```ts
$: ({a,b} = c);
```
Output:
```ts
let {a,b} = c;
```

**At least one let x; defined by user**

Input:
```ts
let a: number;
$: ({a,b} = c);
```
Output
```ts
let a: number;
let b; // <- we don't infer the type, too much work for such an edge case
$: ({a,b} = c);
```


Same goes for `$: ([a, b] = c);`

jasonlyu123:
I added a ts version of Rich's `extract_identifiers` in my branch
https://github.com/jasonlyu123/language-tools/tree/fix-destructuring-inside-himBHsj

dummdidumm:
Oh that's great 👍 I'll incorporate it into my changes.

dummdidumm:
Thanks a lot for that adapted method @jasonlyu123 ! Tests are passing now for both object and array destructuring. Could you take one more look at the code if everything is okay?

packages/svelte2tsx/src/nodes/ImplicitTopLevelNames.ts
```diff
@@ -0,0 +1,79 @@
...
+        str: MagicString,
+    ) {
+        this.map.add(node);
+
+        if (
+            ts.isExpressionStatement(node.statement) &&
+            ts.isParenthesizedExpression(node.statement.expression) &&
+            (ts.isObjectLiteralExpression(binaryExpr.left) ||
+                ts.isArrayLiteralExpression(binaryExpr.left)) &&
+            this.containsNoTopLevelNames(binaryExpr.left, rootVariables)
```
jasonlyu123, line 22: I think this part should execute later. At this point, not all the top-level names are found. if the variable is declared later in the code than the $: statement the parenthesize will be removed but the $; label won't be replaced by `let`.

packages/svelte2tsx/src/svelte2tsx.ts
```diff
@@ -655,10 +655,8 @@ function processInstanceScriptContent(str: MagicString, script: Node): InstanceS
 
         str.prependLeft(start, '__sveltets_invalidate(() => ');
         str.appendRight(end, ')');
-
-        if (!semiRegex.test(htmlx.substring(end))) {
```
jasonlyu123, line 659: This is added by me to prevent this
```
$: [ first ]  = a
['a', 'b'].forEach(item => something(item))
```
because the invalidate function call the last character might be a ')' so the semi might sometimes be needed

packages/svelte2tsx/src/nodes/ImplicitTopLevelNames.ts
dummdidumm: True, we should either do it all right here or do it all in the end.

I think we have to do it right here; as stated in the comment afterwards: Because we use `remove` which would remove everything that we appended/prepended previously in that range. This includes things like the transformation of stores.
In TS it will also throw an error if you use the variable before declaring it. I doubt many people do it, and if so, they just need to move the declaration up.

packages/svelte2tsx/src/svelte2tsx.ts
jasonlyu123: Nevermind I am just having examples that don't works 😅

dummdidumm: Oh ok, I did not think of this situation. The problem with that added semicolon is that I did not find a way to to make it work in combination with the generated output of `$: ({x} = y)`.

jasonlyu123: Nevermind the example doesn't work. it requires a semi before transformation anyways. I must have set that semi for other reasons but I forgot 😂

packages/svelte2tsx/src/nodes/ImplicitTopLevelNames.ts
dummdidumm: I found a way to do the transformation without removing the previously added strings by doing `str.overwrite(start, start + 1, '', { contentOnly: true })` instead of `str.remove(start, start + 1)`

PR #446, opened 2020-08-11
(fix) add missing html attributes

dummdidumm:
@shirakaba FYI


shirakaba:
@dummdidumm Brilliant! Thanks very much for the fix 😊 this’ll help me out immensely next week.

Issue #442, opened 2020-08-11
Typing Svelte Component Props/Events/Slots [Feedback Thread]


dummdidumm:
## [this is now the feedback thread]

This was implemented according to [this RFC](https://github.com/sveltejs/rfcs/pull/38). Please provide feedback about using it in this thread.

## Original Post

There are several issues about this already (#424, sveltejs/svelte#304, sveltejs/svelte#273, sveltejs/svelte#263), but I wanted to make a consolidated one to have a discussion about the approaches we can take.
Related PR: sveltejs/svelte#437 

Note that this issue is NOT about typing a `d.ts` file, this will come afterwards as a separate issue.

**Is your feature request related to a problem? Please describe.**
At the moment it is not possible to type the inputs/outputs of a component under certain circumstances:
- It is not possible to define a generic relationship between props and events/slots (generics)
- It is not possible to define the events and their types
- It is not possible to define the slots and their types in a specific way

**Describe the solution you'd like**
A way to type props/events/slots explicitly.

Proposal: A new reserved interface `ComponentDef` which, when defined, is used as the public API of the component instead of infering the things from the code.

Example (with comments about what likely is not possible):
```html
<script lang="ts"
   interface ComponentDef<T> { // <-- note that we can use generics as long as they are "defined" through props
      props: {  items: T[]  }
      events: {  itemClick: CustomEvent<T>  }
      slots: { default: { item: T } }
  }

   // generic type T is not usable here. Is that a good solution? Should it be possible?
   // Also: How do we make sure the types match the component definition?
  export items: any[];

   // ...

   // We cannot make sure that the dispatched event and its type are correct. Is that a problem?
   // Maybe enhance the type definitions in the core repo so createEventDispatcher accepts a record of eventname->type?
   dispatch('itemClick', item);
</script>
<!-- ... -->
<slot item={item}> <!-- again, we cannot make sure this actually matches the type definition -->
```
When someone now uses the component, he will get the correct types from the props/events/slots and also error diagnostics when something is wrong:
```html
<Child
     items="{['a', 'string']}"
     on:itemClick="{event => event.detail === 1 /* ERROR, number not comparable to type string */}"
/>
```

If this works and is tested a little, we could enhance it with other reserved interfaces like `ComponentEvents` to only type one of the things.

I'd like to get as much feedback on this as possible because this is likely a big change which might be used broadly. Also, I'd like to have some of the core team members to have a look at this if this looks good to them or introduces something they don't agree with.

[...]

Issue #424, opened 2020-08-04
Discussion: How can we type the events supported by a component


AlexGalays:
The goal is to easily catch typos and help the discovery of all possible events dispatched by a component. After all, there is no reason why the inputs (props) can have a strong contract but the events (output) don't.
Right now I can type `on:qoidfoqidjoiqsjd` just fine.

Flex, which also had its compiler used custom annotations for this: https://github.com/apache/flex-sdk/blob/master/frameworks/projects/mx/src/mx/core/Container.as#L97

Perhaps we could have a special `export type Events = {name: 'eventA', data: number} | ...`

dummdidumm:
The reason why this cannot be implemented easily is that it's a lot harder to collect all possible events than to collect the props. That's also the reason why autocompletion does not work because these are for props. 

Looking under the hood it's that props are implemented through jsx props. Events are not because of the limitations of collecting them. If the user would explicitly type them this could be changed. We cannot mark the dispatched events as "this does not conform to the definition" though because of the challenges mentioned.

Having a reserved interface is likely the way forward. We have thought about this before in the context of generic props. Reserved interface names could be `ComponentEvents`, `ComponentProps`, `ComponentSlots` and `ComponentDef` for typing all three.

dummdidumm:
Getting autocompletion with `on:<event>` has to be done differently because under the hood we use JSX to get these autocompletions. To get autocompletion "out of the box", we would need to have `on:<event>` as part of the props. But this is not possible because JSX does not allow characters like `:` in props.
Another problem is that we somehow would need to convert the events type definition to prepend the `on:`, which cannot be done with TypeScript at the moment ([related TS issue](https://github.com/microsoft/TypeScript/issues/12754)).

Two solutions arise from these constraints:
- Tell devs to type their events like `__on__<eventName>: ..` and we would in the language server then replace `__on__` with `on:` during autocomplete. This feels suboptimal and brittle.
- Implement our "own" autocompletion for this. This is possibly a hard task, but would lead to a clean solution.

Getting "`on:XXX` does not exist" errors is easily possible after #386 is merged through a variation of the `$on` method definition not falling back to `CustomEvent<any>` if users type their events explicitely.

dummdidumm:
Once https://github.com/sveltejs/svelte/pull/5260 is released we can try to update `svelte2tsx` so that it searches for `createEventDispatcher`, looks if it has a type annotation, and if yes, uses that to get proper types.

[...]

Issue #304, opened 2020-07-12
Component event type check


jasonlyu123:
**Is your feature request related to a problem? Please describe.**
Current component type check can only check if it is `(e: Event) => any`

**Describe the solution you'd like**
1. Event forward like `<button on:click />` can have type check in consumer component
2. Can type check for event dispatched by event-dispatcher 

**Describe alternatives you've considered**
if the event dispatched by event-dispatcher can't be type-check or it is dispatched by an imported function provides a way to manually define the event's type. For example:
```ts
interface $$ComponentEventDef {
    'some-event': CustomEvent<number>
}
```

dummdidumm:
Closing in favor of #424

PR #459, opened 2020-08-17
(feat) ComponentEvents interface

dummdidumm:
This adds the possibility to use a reserved interface name `ComponentEvents` and define all possible events within it.
Also adds autocompletion for these events.
Also disables autocompletions from HTMLPlugin on component tags.
NO support for generics.

#424 #304 #442 

TODO:
- [x] tests
- [x] when interface is defined, make events strict so that events that do not exist are marked as error
- [x] docs


packages/svelte2tsx/src/nodes/ComponentEvents.ts
```diff
@@ -0,0 +1,82 @@
...
+        const map = new Map<string, { type: string; doc?: string }>();
+
+        node.members.filter(ts.isPropertySignature).forEach((member) => {
+            map.set(member.name.getText(), {
+                type: member.type?.getText() || 'Event',
+                doc: this.getDoc(node, member),
+            });
+        });
+
+        return map;
```
jasonlyu123, line 41: The name here is possibly a string literal or a computed property. i.e.
```
interface ComponentEvents {
    'kebab-case': CustomEvent<number>
    [SOME_CONSTANT]: CustomEvent<number>
}
```
`getText` in either situation would be wrong. if it's a string literal the text can be extracted from `member.name.text`.
About the computed property, from my testing, computed event handler like <Component on:[SOME_CONSTANT] /> doesn't work. Maybe we could make it works here but filter out in the completion. Or simply doesn't support it as it doesn't make much sense when you can't use it in markup.

jasonlyu123 (changes requested):
Mostly great, only an issue about string literal or computed property name needs more works

packages/language-server/src/plugins/typescript/features/CompletionProvider.ts
```diff
@@ -71,10 +72,11 @@ export class CompletionsProviderImpl implements CompletionsProvider<CompletionEn
             ? triggerCharacter
             : undefined;
         const isCustomTriggerCharacter = triggerKind === CompletionTriggerKind.TriggerCharacter;
+        const isInvoked = triggerKind === CompletionTriggerKind.Invoked;
 
         // ignore any custom trigger character specified in server capabilities
         //  and is not allow by ts
-        if (isCustomTriggerCharacter && !validTriggerCharacter) {
+        if (isCustomTriggerCharacter && !validTriggerCharacter && !isInvoked) {
```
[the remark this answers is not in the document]
dummdidumm: No, the added check makes it possible to get completions when doing Ctrl+Space (in my keybindings). I'll update the comment above.

packages/svelte2tsx/src/nodes/ComponentEvents.ts
dummdidumm: Good catch! We should definitely support string literals. About the constant, we could go to the root and check if it's defined there and get the value from there - so we at least support it a little. I will also add docs about this.

packages/language-server/src/plugins/typescript/features/CompletionProvider.ts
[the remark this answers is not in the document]
jasonlyu123: What did I mean it that if `isCustomTriggerCharacter` is true isn't `isInvoked` always be false?

packages/svelte2tsx/src/svelte2tsx.ts
```diff
@@ -1016,6 +1032,7 @@ export function svelte2tsx(
         str,
         uses$$props || uses$$restProps,
         !!options?.strictMode,
+        events instanceof ComponentEventsFromInterface,
```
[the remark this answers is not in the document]
dummdidumm: Yeah makes sense, `addComponentExport` has way too many parameters right now 😄

packages/language-server/src/plugins/typescript/features/CompletionProvider.ts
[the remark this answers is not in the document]
dummdidumm: Oh yes, you are completely right ... why did I do this? 😄 I'll remove it.

packages/svelte2tsx/test/svelte2tsx/index.js
```diff
@@ -16,15 +16,20 @@ describe('svelte2tsx', () => {
...
-            const { map, code} = svelte2tsx(input, {
+            const output = svelte2tsx(input, {
 				strictMode: dir.includes('strictMode'),
 				isTsFile: dir.startsWith('ts-'),
 				filename: 'input.svelte'
 			});
-            assert.equal(code, expectedOutput);
+			assert.equal(output.code, expectedOutput);
+			if (expecterOtherOutput) {
+				expecterOtherOutput(output);
```
dummdidumm, line 31: I enhanced the test suite so you can test other outputs of `svelte2tsx`, see `component-events-interface-constant/expected.js` for an example.

packages/svelte2tsx/src/nodes/ComponentEvents.ts
```diff
@@ -0,0 +1,129 @@
...
+    private throwError(prop: ts.PropertyName) {
+        const error: any = new Error(
+            'The ComponentEvents interface can only have properties of type ' +
+                'Identifier, StringLiteral or ComputedPropertyName. ' +
+                'In case of ComputedPropertyName, ' +
+                'it must be a const declared within the component and initialized with a string.',
+        );
+        error.start = toLineColumn(prop.getStart());
+        error.end = toLineColumn(prop.getEnd());
+        throw error;
```
dummdidumm, line 80: This throws if the interface definition does not match our expextations. I hope this makes it clear to the user what is possible and what is not.

packages/svelte2tsx/src/nodes/ComponentEvents.ts
```diff
@@ -41,6 +42,52 @@ export class ComponentEventsFromInterface extends ComponentEvents {
         return map;
     }
 
+    private getName(prop: ts.PropertyName) {
+        if (ts.isIdentifier(prop) || ts.isStringLiteral(prop)) {
```
jasonlyu123, line 46: Also NumberLiteral

dummdidumm: According to the Svelte typings, only `string` is possible for the events: https://github.com/sveltejs/svelte/blob/master/src/runtime/internal/lifecycle.ts#L32

jasonlyu123: If it's official now then is fine. I just tried it in REPL and find it possible as a number XD.

dummdidumm: I think it will be stringified, so the correct typing would be to only allow `string`.

dummdidumm:
Thanks for the feedback! I added the requested changes. Let me know if it's all good now for you.

jasonlyu123 (approved):

nateshmbhat:
can someone please provide on how to define type defs to include events too ?

dummdidumm:
This is not official yet until [this RFC](https://github.com/sveltejs/rfcs/pull/38) is finalized and merged. You can type `createEventDispatcher` though.

PR #464, opened 2020-08-20
support $$slots

tanhauhau:
$$slots get introduced here: https://github.com/sveltejs/svelte/pull/4602


dummdidumm:
Thanks a lot! I reworked the `$$slots` generation a little. We cannot use type annotations in the transformation because the user could be using JS only and would get "type annotations can only be used in TS files" errors.

Issue #462, opened 2020-08-20
Easy to accidentally add imports for same component


brunnerh:
**Describe the bug**
If i have a file re-exporting components and then import from that, the import to the original file gets inserted in the script automatically when auto-completing the component name and using the first result.

**To Reproduce**

Example:

```ts
// Barrel file: /utility/index.ts
import Button from './button.svelte';
import TextField from './text-field.svelte';
import Dialog from './dialog.svelte';
...

export {
	Button,
	TextField,
	Dialog,
	...
}
```

```svelte
<!-- Component /components/edit-dialog.svelte -->
<script type="text/typescript">
	import { TextField, Button, Dialog } from "../utility";
	import Dialog from "../utility/dialog.svelte"; // Gets inserted on auto-completing to Dialog tag in HTML below
</script>

<Dialog>...
```

**Expected behavior**
The existing named import is recognized and no additional import is added.

**Screenshots**

![image](https://user-images.githubusercontent.com/834235/90765661-9fd54400-e2ea-11ea-8920-229cecf1b2ce.png)

There are two `Dialog` entries, the first one adds the redundant import, the other one is for the existing named import. Ideally the first suggestion would not appear if the named import is already there.

**System (please complete the following information):**
 - OS: Windows 10, 64bit
 - IDE: VSCode
 - Plugin/Package: Svelte for VSCode

brunnerh:
(I only noticed that both suggestions exist at the same time just now, so this might not necessarily be considered a bug.)

dummdidumm:
All Svelte component under the hood are postfixed with `__SvelteComponent` so that you can import something called `Component` into `Component.svelte` and not have name clashes from typescript. The problem now likely is that we leave module resolution to typescript, which then sees that `X__SvelteComponent` is not imported yet. A fix would be to sort the results differently or to filter out the completion.

PR #466, opened 2020-08-20
(fix) filter out component import if already one present

dummdidumm:
#462
Drawback: Also will mark commented out imports as already imported - but this seems like a bigger edge case than having double imports when doing barrel exports. @jasonlyu123  what's your opinion on this?


jasonlyu123:
I am ok with the commented-out-import.

PR #469, opened 2020-08-21
do not suggest lang when editing attributes

tanhauhau:
Before

Screenshot 2020-08-21 at 8 25 22 AM

After

Screenshot 2020-08-21 at 8 24 51 AM


tanhauhau:
how do i add test for this?

jasonlyu123:
You can have a test similar to this one
https://github.com/sveltejs/language-tools/blob/9f9fef99e3618ef85eec3c310b4f56a3066b8fde/packages/language-server/test/plugins/html/HTMLPlugin.test.ts#L41

setup('<div sty') and filter the return list see if lang completion exists

PR #480, opened 2020-08-26
(fix) multi folders in workspace

g-plane:
Currently, Svelte Language Server doesn't work well when there're multiple folders in a workspace. This PR fixes it.

Hope to be merged.

### DIFF
--- a/packages/language-server/src/server.ts
+++ b/packages/language-server/src/server.ts
@@ -86,10 +86,10 @@ export function startServer(options?: LSOptions) {
     let sveltePlugin: SveltePlugin = undefined as any;
 
     connection.onInitialize((evt) => {
-        const workspacePaths = evt.workspaceFolders?.map(folder => folder.uri.toString())
+        const workspaceUris = evt.workspaceFolders?.map(folder => folder.uri.toString())
             ?? [evt.rootUri ?? ''];
-        Logger.log('Initialize language server at ', workspacePaths.join(', '));
-        if (workspacePaths.length === 0) {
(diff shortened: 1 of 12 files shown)


packages/language-server/src/server.ts
```diff
@@ -86,9 +86,9 @@ export function startServer(options?: LSOptions) {
     let sveltePlugin: SveltePlugin = undefined as any;
 
     connection.onInitialize((evt) => {
-        const workspacePath = urlToPath(evt.rootUri || '') || '';
-        Logger.log('Initialize language server at ', workspacePath);
-        if (!workspacePath) {
+        const workspacePaths = evt.workspaceFolders?.map(folder => folder.uri.toString()) ?? [];
```
jasonlyu123, line 89: For potential backward compatible and for LSP clients that don't support multi-root-workspaces can you make it fallback to the old behavior?

jasonlyu123 (changes requested):
If I understand correctly you are trying to support [multi-root-workspaces](https://code.visualstudio.com/docs/editor/multi-root-workspaces)?

packages/language-server/src/server.ts
g-plane: Fallback to `evt.rootUri`?

jasonlyu123: Yes.

g-plane:
Are there any other problems which should be addressed?

jasonlyu123 (approved):

dummdidumm (approved):
👍

Issue #478, opened 2020-08-25
Syntax highlighting breaks in multi-line attributes


brunnerh:
**Describe the bug**
If an attribute value contains multiple lines and the first line has no interpolation, the syntax highlighting of interpolations in later lines will not be applied.

**To Reproduce**

Example component:

```svelte
<script>
	export let question = 'Really';
	export let flag = true;
</script>

<!-- Second line highlighted. -->
<div title="{question}?
	{flag ? 'Yes' : 'No'}"/>

<!-- Second line not highlighted. -->
<div title="Really?
	{flag ? 'Yes' : 'No'}"/>
```

(The same seems to be true for the highlighting on GitHub.)

**Expected behavior**
Syntax highlighting on all interpolations.

**Screenshots**

![image](https://user-images.githubusercontent.com/834235/91206229-1c12c180-e707-11ea-9aae-8d8cdd91cd5f.png)

**System (please complete the following information):**
 - OS: Windows 10, 64bit
 - IDE: VSCode 1.48.1
 - Plugin/Package: Svelte for VSCode (v101.10.0)

dummdidumm:
If this is fixed, the syntax highlighting part of #381 is probably fixed, too.

Issue #381, opened 2020-07-29
Autocomplete and syntax highlighting for style attributes


techniq:
Coming from React/JSX, you could get the list of CSS properties within the `style` property/attribute and have syntax highlighting.

It uses object notation instead of a string, although this surprising this still shows up within a Svelte component (albeit not supported).  I figure due to the use of `svelte2tsx`

![image](https://user-images.githubusercontent.com/177476/88824053-d9a3a680-d193-11ea-86c0-cf16e51f7aa1.png)

![image](https://user-images.githubusercontent.com/177476/88824002-c0025f00-d193-11ea-96e9-80a5ff551c53.png)

It would be useful if the same was supported within Svelte when using the `style` attribute (string).

![image](https://user-images.githubusercontent.com/177476/88824098-e627ff00-d193-11ea-994f-bba202546817.png)

![image](https://user-images.githubusercontent.com/177476/88824140-f4761b00-d193-11ea-882d-88a6e05205da.png)

It would also be helpful if the expressions within the values were highlighted as such (and not just shown as one big string).

![image](https://user-images.githubusercontent.com/177476/88824274-27201380-d194-11ea-831f-ebba0c42ef2f.png)

tanhauhau:
Reading through how this works in VSCode HTML language server, it seemed like they are using tokenizer to map out different language regions in the HTML document.

https://github.com/vscode-langservers/vscode-html-languageserver/blob/master/src/modes/embeddedSupport.ts#L26

seeing how we map out script, context script, style from the Document, it seemed like we are doing at a different approach.

i want to work on this, but i am wondering what is the recommended approach here to figure out the language mode?

jasonlyu123:
We're currently using [vscode-html-languageserver's parser](https://github.com/microsoft/vscode-html-languageservice/blob/master/src/parser/htmlParser.ts) to do style/script tag extraction. The parser also uses the same token scanner to parse the file. so we could probably use the parsing result to get style attributes. And if there's an advantage using the token scanner maybe we could migrate to that.
After the style attributes have been extracted, the next problem is how do we handle the interpolation. With the mustache, the code is not valid CSS so the problem is how do we transform it to let CSS language service to understand it?
As for the language modes, the plugin in our codebase is similar concept. Maybe we can create another plugin for the style attributes and shared some logic with the CSS plugin.

dummdidumm:
The syntax highlighting part will be fixed with the next release.

jasonlyu123:
[...]

PR #483, opened 2020-08-26
(fix) moustache highlighting multiline

dummdidumm:
Highlights moustache tags properly when the tag spans multiple lines
#478, #381

This fix looks suspicously simple, so it would be great if you could have a look at this @jasonlyu123  😄

### DIFF
--- a/packages/svelte-vscode/syntaxes/svelte.tmLanguage.json
+++ b/packages/svelte-vscode/syntaxes/svelte.tmLanguage.json
@@ -790,7 +790,7 @@
             ]
         },
         "tag-moustaches": {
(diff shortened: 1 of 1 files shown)


jasonlyu123:
Ummm...  I didn't found any downside to this either. Seems like that part of the regex is not needed.

dummdidumm:
Alright, then it really must be that simple 😄  Thanks for looking into it!

jasonlyu123:
Feel dumb to have spent 40 minutes trying to make it match `\\b([a-zA-Z\\-_:]+)=(\"|')(?=(.|\\s)*{`

Issue #489, opened 2020-08-26
Unused export property warning is on the wrong line when using the TypeScript pre-processor.


brunnerh:
**Describe the bug**
`export let` statements of an unused property cause a warning. The warning is displayed on the line above the actual property for me.

**To Reproduce**

Set the script language to typescript (`lang="ts"` or `type="text/typescript"`).
Declare some export property, do not use it anywhere.

**Expected behavior**

The warning is on the property not being used.

**Screenshots**

![image](https://user-images.githubusercontent.com/834235/91361338-5e67fb80-e7f8-11ea-8436-282065c2a241.png)

If the property is in the first line of the script:

![image](https://user-images.githubusercontent.com/834235/91361408-8192ab00-e7f8-11ea-96bd-eeca55a8de72.png)

**System (please complete the following information):**
 - OS: Windows 10, 64bit
 - IDE: VSCode
 - Plugin/Package: Svelte for VSCode

jasonlyu123:
Try update your @tsconfig/svelte to lastest

brunnerh:
I am not using that.
Added it temporarily, did nothing.

jasonlyu123:
Oh I thought you were using the official template. The package is a base config for svelte. You can update the  `extends` of your `tsconfig.json` to match this.
https://github.com/sveltejs/template/blob/8194bf8f6452f8117195bf304ac99b15d329fbb3/scripts/setupTypeScript.js#L87

jasonlyu123:
Duplicate of #375

brunnerh:
@jasonlyu123 As i said, i added the base config temporarily but nothing changed. I still get the errors on the wrong line.

jasonlyu123:
Then can you try run `tsc` to compile a ts file in your workspace and see if the output file has sourcemap?

jasonlyu123:
The problem is with sourcemap. We use the svelte compiler to get svelte diagnostics. if you have a svelte.config.js and specify your preprocessor in it we would use that to preprocess, other than that we'll use the official svelte-preprocess. And if the preprocessor doesn't return a sourcemap the warning would be in a wrong position.

brunnerh:
It does not always generate source maps. If i call `tsc` on a single file it apparently does not load the tsconfig; if i use `tsc -p .` or `tsc --build tsconfig.json` it will output the source map as expected.

jasonlyu123:
[...]

PR #495, opened 2020-08-28
always include source map in ts-preprocess of internal svelte compile

jasonlyu123:
svelte-preprocess merge the `compilerOption` with the parsed config file
https://github.com/sveltejs/svelte-preprocess/blob/36d6421db847f8886ebad4c0e8c77671a9a4e3f9/src/transformers/typescript.ts#L69
So it won't override other user-defined configs. Although the problem mentioned here https://github.com/sveltejs/language-tools/issues/489#issuecomment-682281963 here still exists, this would at least make the set up a little bit easier.


dummdidumm:
This will only fix errors for people who do not have a `svelte.config.js` inside their workspace, because it's only a fallback. But as you say, in these cases it will make setup easier 😄

Issue #494, opened 2020-08-27
Svelte innerHTML is not supported in typescript


cortopy:
**Describe the bug**

The following code from the [svelte docs](https://svelte.dev/docs#Binding_select_value):

```svelte
<div contenteditable="true" bind:innerHTML={html}></div>
```

generates the following error:

```
(JSX attribute) innerHTML: string | undefined
Type '{ contenteditable: "true"; innerHTML: string | undefined; }' is not assignable to type 'HTMLProps<HTMLDivElement>'.
  Property 'innerHTML' does not exist on type 'HTMLProps<HTMLDivElement>'.ts(2322)
```

**To Reproduce**

I'm using code similar to this

```
<script>
const html = "<p>hello</p>"
</script>

<div contenteditable="true" bind:innerHTML={html}></div>

```

**Expected behavior**
The tools recognise the HTML code as valid

**Screenshots**
![Screenshot_2020-08-27_23-46-19](https://user-images.githubusercontent.com/8725825/91502371-caf9fd80-e8bf-11ea-8655-6bc63c88e028.png)

**System (please complete the following information):**
 - OS: Linux
 - IDE: VSCode
 - Plugin/Package:

PR #496, opened 2020-08-28
(fix) innerHTML and textContent binding

jasonlyu123:
#494


dummdidumm:
I think they need to be all lowercase because `svelte2tsx` lowercases attributes

Issue #497, opened 2020-08-28
Sapper: Support for preload function


dummdidumm:
**Is your feature request related to a problem? Please describe.**
Using sapper with the `preload` function inside module script gives a bad DX.

```ts
export async function preload(page, session) { // <------- not typed, need to add typings yourself
    const res = await this.fetch(`blog/${slug}.json`); // <--- error, what it `this`?
}
```

**Describe the solution you'd like**
Support typing for the preload function and for this, but in such a way that it will not also provide these for non-sapper-projects. 

**Notes / open questions**
- `this` probably needs to be transformed to some ambient global which has `fetch` defined on it.
- `preload` is tricky because it's not a _usage_ of something we can provide in the ambient context, but a _definition_ . Not sure how to implement this.
- How to only provide these typings/transformations in a Sapper project? How can we determine this? Look into the `package.json`?

dummdidumm:
After some more thought and hearing that the context is possibly user-modifiable in the future, I'm deciding against baking this into the language-tools.

- getting arguments ambiently typed is not possible
- typing `this` is possible to do by the user, and since it may be modified, our typings may be wrong anyway in the future

### How to type this `this` (hah) then?

You can type the `this` context of a function explicitly in TypeScript ([see docs](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#specifying-the-type-of-this-for-functions)):

```ts
     // Better move that into some file and do `import type { PreloadContext } from ...` to not type it every time
    interface PreloadContext {
        fetch: ...,
        ...
    }

    // won't work with export const preload = (...
    export async function preload(this: PreloadContext, ...page/session, if you need it) {
       ...
       this.fetch(...) // <-- works without type errors
    }
```

dummdidumm:
Fixed via https://github.com/sveltejs/sapper/pull/1468

PR #499, opened 2020-08-29
(feat) Sapper: shim `this` inside preload function

dummdidumm:
#497 

Question:
- Don't care about detection and just do the transformation either way? --> Yes, don't care for now, too much of an edge case to care for it

Side note:
- Found a source map bug during this when only `<script context="module">` is present?

### DIFF
--- a/packages/language-server/src/plugins/typescript/service.ts
+++ b/packages/language-server/src/plugins/typescript/service.ts
@@ -65,12 +65,7 @@ export function createLanguageService(
 
     const svelteModuleLoader = createSvelteModuleLoader(getSnapshot, compilerOptions);
 
-    const svelteTsPath = dirname(require.resolve('svelte2tsx'));
-    const svelteTsxFiles = [
-        './svelte-shims.d.ts',
-        './svelte-jsx.d.ts',
-        './svelte-native-jsx.d.ts',
-    ].map((f) => ts.sys.resolvePath(resolve(svelteTsPath, f)));
+    const svelte2TsxFiles = getSvelte2TsxFiles();
 
     const host: ts.LanguageServiceHost = {
         getCompilationSettings: () => compilerOptions,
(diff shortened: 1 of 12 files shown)


jasonlyu123:
I think we can add some check if the user has already type `this` in the preload function
```ts
interface PreloadContext {
    fetch: (url: string, options?: any) => Promise<any>;
    error: (statusCode: number, error: any) => void;
    redirect: (statusCode: number, location: string) => Promise<any>;
}
export function preload(this: PreloadContext) {})
```

dummdidumm:
Good point, did not think about that. We could also make this the proposed solution to get proper typing without baking it into the language tools.

@benmccann what do you think about this? Is it feasible/better to add typings for the `this`-context and the other `preload` arguments to the Sapper typings and instruct people in the docs to use those when they want to use `preload` in TypeScript projects?

benmccann:
One thing to note is that the Sapper preload context may become user-modifiable along the lines of https://github.com/sveltejs/sapper/pull/1336

It does seem like language-tools shouldn't need to know about Sapper specifically. Though the current way for the user to do this is really ugly: https://github.com/babichjacob/sapper-typescript-graphql-template/issues/19#issuecomment-683362969. I'm not quite sure what the best way to improve this would be.

dummdidumm:
@benmccann thanks for the issue links. This makes me think even more that we should not know about this `this` (hah).

As mentioned in my comment above, the best way IMO would be to provide typings for the standard context and the preload/session parameters:

````html
<script context="module">
    import type { PreloadContext, PreloadPage, PreloadSession } from "@sapper/preload";

    export async function preload(this: PreloadContext, page: PreloadPage, session: PreloadSession) {
       ...
       this.fetch(...) // <-- works without type errors
    }
</script>
```

If you tell me where to put them I'm happy to open a PR with those types over at the Sapper repo.

Issue #500, opened 2020-08-29
Property 'playbackRate' does not exist on type 'HTMLProps<HTMLVideoElement>'


martingraf4:
**Describe the bug**

Binding `playbackRate` on a `video` element results in the following error (when running `svelte-check`): `Error: Type '{ playbackRate: number; }' is not assignable to type 'HTMLProps<HTMLVideoElement>'. Property 'playbackRate' does not exist on type 'HTMLProps<HTMLVideoElement>'. (ts)`

**To Reproduce**

```svelte
<script lang="ts">
  let playbackRate: number;
</script>

<video bind:playbackRate>
  <track kind="captions" />
</video>
```

**Expected behavior**

No error

**System:**

 - OS: Windows
 - IDE: VSCode
 - Plugin/Package: `svelte-check`

PR #502, opened 2020-08-30
svelte media bindings

jasonlyu123:
#500

### DIFF
--- a/packages/svelte2tsx/svelte-jsx.d.ts
+++ b/packages/svelte2tsx/svelte-jsx.d.ts
@@ -251,6 +251,14 @@
       /**
        * Elements with the contenteditable attribute support innerHTML and textContent bindings.
        */
+      innerHTML?: string;
+      /**
+       * Elements with the contenteditable attribute support innerHTML and textContent bindings.
+       */
+      textContent?: string;
+            /**
+       * Elements with the contenteditable attribute support innerHTML and textContent bindings.
+       */
       innerhtml?: string;
       /**
        * Elements with the contenteditable attribute support innerHTML and textContent bindings.
@@ -358,6 +366,9 @@
       type?: string;
       usemap?: string;
       value?: string | string[] | number;
+      /**
+       * a value between 0 and 1
+      */
       volume?: number;
       width?: number | string;
       wmode?: string;
@@ -713,7 +724,7 @@
     }
 
     interface SapperAnchorProps {
-        sapperNoscroll?: true;
+        sappernoscroll?: true;
     }
 
     interface SvelteMediaTimeRange {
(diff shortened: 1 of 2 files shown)


jasonlyu123:
The `readonly` modifier isn't needed. It's just for documentation purposes. And it could be easiler for future works if someday we can typecheck readonly props.

packages/svelte2tsx/svelte-jsx.d.ts
```diff
@@ -716,6 +716,40 @@
...
+        readonly buffered?: SvelteMediaTimeRange[];
+        readonly played?: SvelteMediaTimeRange[];
+        readonly seekable?: SvelteMediaTimeRange[];
+        readonly seeking?: boolean;
+        readonly ended?: boolean;
+
+        /**
+         * the current playback time in the video, in seconds
+         */
+        currentTime?: number;
```
dummdidumm, line 735:
```suggestion
        currenttime?: number;
```

packages/svelte2tsx/svelte-jsx.d.ts
```diff
@@ -716,6 +716,40 @@
...
+        /**
+         * how fast or slow to play the video, where 1 is 'normal'
+         */
+        playbackRate?: number;
```
dummdidumm, line 739:
```suggestion
        playbackrate?: number;
```

packages/svelte2tsx/svelte-jsx.d.ts
```diff
@@ -716,6 +716,40 @@
...
+        paused?: boolean;
+        /**
+         * a value between 0 and 1
+         */
+        volume?: number;
+        muted?: boolean;
+    }
+
+    interface SvelteVideoProps extends SvelteMediaProps {
+        readonly videoWidth?: number;
```
dummdidumm, line 749:
```suggestion
        readonly videowidth?: number;
```

packages/svelte2tsx/svelte-jsx.d.ts
```diff
@@ -716,6 +716,40 @@
...
+        readonly videoHeight?: number;
```
dummdidumm, line 750:
```suggestion
        readonly videoheight?: number;
```

dummdidumm (changes requested):
Nice idea of not having it as part of the big `HTMLAttributes` list 👍  I suggested some lowercase changes because all props are lowercased by `svelte2tsx`. Also some properties which are on `HTMLAttributes`, like `defaultplaybackrate`, should be moved into this new interface I think.

jasonlyu123:
I have to look into it. all lowercase would have typescript errors on two way binding but camel case won't have error in both props and two way binding.

jasonlyu123:
two-way binding is not being transformed to all lowercase. Another problem is most of these camel case props are not HTML attributes but svelte props. The compiler would complain when you use all lower cases which means the autocomplete is wrong.

dummdidumm:
Damn that's tricky ... About the autocompletion: We could try to turn of autocompletion for all JSX intrinsic elements, they should be shown anyway through the html language service. This would have the additional benefit of not everything showing up twice. If that works, you could just have two variants inside the JSX typings - one all lowercase for the attribute, one camelcased for the binding.

dummdidumm:
I just noticed that the fix in #496 for #494 has the same problem (so #494 is not fixed yet). My bad.

jasonlyu123:
What do you think about also transform two-way binding to lowercase?  We can let the svelte compiler validate the casing on the binding on the svelte-specific props.

dummdidumm:
Good idea, I think this will work. It's transformed to a simple input property, so we can lowercase it like we do with the real input properties. But we should only do this for html elements, not component elements.

jasonlyu123:
It seems like only binding is case-sensitive that's just put two variant as you suggest earlier

jasonlyu123:
Some of it doesn't work when used as HTML attribute. It seems like they are svelte spefic props. So I removed the lowercase variant 
```svelte
<script>let playbackRate = 1</script>
<video playbackRate={1} /> <!--doesn't work -->
<video bind:playbackRate /> <!--work -->
```

dummdidumm (approved):
👍

PR #507, opened 2020-08-30
fix failure to start with emacs/lsp-mode

martinxyz:
initializationOptions was null (it's optional but the vscode client
always sets it), which resulted in this traceback:

TypeError: Cannot read property 'svelte' of undefined
+    at LSConfigManager.update (.../dist/src/ls-config.js:61:46)
+    at PluginHost.updateConfig (.../PluginHost.js:29:21)
...snip...

Fixes regression from #411

### DIFF
--- a/packages/language-server/src/server.ts
+++ b/packages/language-server/src/server.ts
@@ -94,7 +94,7 @@ export function startServer(options?: LSOptions) {
         }
 
         pluginHost.initialize(!!evt.initializationOptions?.dontFilterIncompleteCompletions);
-        pluginHost.updateConfig(evt.initializationOptions?.config);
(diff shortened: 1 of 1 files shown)


martinxyz:
Note: I'm not sure why typescript didn't catch this bug. It certainly does complain if I write `pluginHost.updateConfig(undefined)`.

dummdidumm:
Thanks! Error probably got introduced in #476

martinxyz:
For the record: with some help, I figured out why typescript didn't complain about the bug.

`evt.initializationOptions` is of type `any`. The `.?` operator will add a free `| undefined`, but `any | undefined` is still `any`, and will happily cast to `LSConfig`.

PR #510, opened 2020-08-31
enhance diagnostics

dummdidumm:
Make it more clear what the `cannot be used as a JSX component` error is about.

@jasonlyu123 could you have a look and tell me if the error message is understandable?


jasonlyu123:
Maybe a simple example of $$props_def?

jasonlyu123 (approved):
👍

PR #511, opened 2020-09-01
(fix) (svelte2tsx) also transform element true attribute to lowercase

jasonlyu123:
ex: `<input READONLY>`

### DIFF
--- a/packages/svelte2tsx/src/htmlxtojsx/index.ts
+++ b/packages/svelte2tsx/src/htmlxtojsx/index.ts
@@ -354,9 +354,11 @@ export function convertHtmlxToJsx(
     };
 
     const handleAttribute = (attr: Node, parent: Node) => {
-        const sapperNoScroll = attr.name === 'sapper:noscroll';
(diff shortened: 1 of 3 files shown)


packages/svelte2tsx/src/htmlxtojsx/index.ts
```diff
@@ -384,7 +384,13 @@ export function convertHtmlxToJsx(
         }
 
         //we are a bare attribute
-        if (attr.value === true) return;
+        if (attr.value === true) {
+            if (parent.type === 'Element' && !sapperNoScroll && parent.name !== '!DOCTYPE') {
+                console.log(parent);
```
dummdidumm, line 389: A wild `console.log` appears!

dummdidumm: ... okay not anymore 😄

jasonlyu123: Fail to remove it before you see it lol

dummdidumm (approved):
👍

PR #528, opened 2020-09-07
warn if old extension is installed

dummdidumm:
This still comes up and causes confusion. Add a warning message which shows up at the start if you have the old extension installed.


jasonlyu123 (approved):
Tested 👍

Issue #539, opened 2020-09-14
Svelte for VS Code is inserting 2 spaces at document formating, instead tab (with size 4), as is set in VS Code.


Mlocik97:
**To Reproduce**
install Svelte for VS Code
npx degit svelte/sapper-template#rollup
open for example _error.svelte
set "intend using tab size 4" and convert
ctrl+shift+i -> get two spaces.

```html
<svelte:head>
  <title>{status}</title>
</svelte:head>

<h1>{status}</h1>

<p>{error.message}</p>

{#if dev && error.stack}
  <pre>{error.stack}</pre>
{/if}
```

**Expected behavior**

**Screenshots**
![Snímka obrazovky z 2020-09-14 16-07-53](https://user-images.githubusercontent.com/34808350/93096278-79fe4d80-f6a4-11ea-925d-911227c3baf7.png)

**System (please complete the following information):**
 - OS: Ubuntu 20.04
 - IDE: VS Code
 - Plugin/Package: [Svelte for Code]

dummdidumm:
Formatting is done with Prettier, but we do not fall back to the user settings if there is no prettier config. We should do something [similar to the Prettier VSCode Plugin](https://github.com/prettier/prettier-vscode/blob/main/src/ConfigResolver.ts).

dummdidumm:
You will not be able to indent on a per-file level, so doing the indentation like that will not work even after we fix the settings bug. But the indentation in your user settings should be adhered to after the fix.

dummdidumm:
Looks like we can use the `FormattingOptions` parameter of the LSP request to get the desired behavior - but we will only use it if no prettier config is found. I'll also have to check with the other maintainers if this is a change we want to make, because it's a breaking change.

dummdidumm:
This will now work IF you don't have any prettier configuration in either your VSCode settings or through a config file.

Mlocik97:
@dummdidumm 

https://www.youtube.com/watch?v=eHazE93BClo&feature=youtu.be&ab_channel=GabrielMlocik

Mlocik97:
thanks so much, it's working now.

PR #554, opened 2020-09-21
(fix) fall back to FormattingOptions

dummdidumm:
In case no prettier configs are found
#539

@jasonlyu123 what do you think? Is this change a change we want to make? It's a breaking change and _might_ be confusing. On the other hand it's also confusing right now that your user settings are not used if you don't know of prettier. So I'm in favor of it.

Situation: 
User does not have any prettier config (either through `.prettierrc` or through VSCode settings)

Before:
Prettier defaults of tabWidth and useTabs are used

After:
The FormattingOptions in the format request are used. These are usually the tabSize/insertSpaces editor settings.

### DIFF
--- a/packages/language-server/src/plugins/svelte/SveltePlugin.ts
+++ b/packages/language-server/src/plugins/svelte/SveltePlugin.ts
@@ -36,7 +36,7 @@ export class SveltePlugin
         CodeActionsProvider {
     private docManager = new Map<Document, SvelteDocument>();
 
-    constructor(private configManager: LSConfigManager, private prettierConfig: any) {}
+    constructor(private configManager: LSConfigManager, private prettierConfig?: any) {}
 
     async getDiagnostics(document: Document): Promise<Diagnostic[]> {
         if (!this.featureEnabled('diagnostics')) {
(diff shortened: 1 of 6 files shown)


jasonlyu123:
I'm also in favor of it. By default `tabSize` and `insertSpaces` might be overridden by indentation detection. So maybe it would not affect many people on existing files. On new files, it would use editor settings for indentation so it probably won't affect much either.

packages/language-server/src/plugins/svelte/SveltePlugin.ts
```diff
@@ -74,9 +71,9 @@ export class SveltePlugin
             (await prettier.resolveConfig(filePath, { editorconfig: true })) ||
             this.prettierConfig ||
             // Be defensive here because IDEs other than VSCode might not have these settings
-            (this.editorConfig && this.editorConfig.tabSize && {
-                tabWidth: this.editorConfig.tabSize,
-                useTabs: !this.editorConfig.insertSpaces,
+            (options && {
```
jasonlyu123, line 74: We fall back to an empty object in the instantiation of `SveltePlugin` in the `server.ts`. When I tried disabled the prettier extension it doesn't fall back to the editor options.

dummdidumm: Oh right 😄 this must be changed.

jasonlyu123 (approved):

PR #557, opened 2020-09-23
(feat) directive comments snippet

jasonlyu123:
1. component documentation 
圖片
2. typescript directives comment like `// @ts-check`
圖片


packages/language-server/src/plugins/svelte/features/getCompletions.ts
```diff
@@ -8,9 +9,21 @@ import {
...
- * Get completions for special svelte tags within moustache tags.
- */
+
+const HTML_COMMENT_START = '<!--';
+
+const componentDocumentationCompletion: CompletionItem = {
+    label: '@component',
+    insertText: `component${EOL}$1${EOL}`,
+    documentation: 'documentation for this component. ' +
+        'It will show up on hover. You can use markdown or code block here',
```
dummdidumm, line 19:
```suggestion
    documentation: 'Documentation for this component. ' +
        'It will show up on hover. You can use markdown and code blocks here',
```

packages/language-server/src/plugins/svelte/features/getCompletions.ts
```diff
@@ -28,13 +41,40 @@ export function getCompletions(
...
+            return null;
+        }
+
+        const commentStartIndex = lastCharactersBeforePosition.lastIndexOf(HTML_COMMENT_START);
+        const text = lastCharactersBeforePosition.substring(
+            commentStartIndex + HTML_COMMENT_START.length
+        ).trimLeft();
+
+        if (componentDocumentationCompletion.label.includes(text)) {
+            return CompletionList.create([componentDocumentationCompletion], true);
```
dummdidumm, line 74: Is there a reason why you set the completion to `incomplete`? Are there cases where other completions would no longer show up if we set it to `false`?

packages/language-server/src/plugins/svelte/features/getCompletions.ts
```diff
@@ -28,13 +41,40 @@ export function getCompletions(
...
+
+    function getComponentDocumentationCompletions() {
+        if (!lastCharactersBeforePosition.includes(HTML_COMMENT_START)) {
+            return null;
+        }
+
+        const commentStartIndex = lastCharactersBeforePosition.lastIndexOf(HTML_COMMENT_START);
+        const text = lastCharactersBeforePosition.substring(
+            commentStartIndex + HTML_COMMENT_START.length
+        ).trimLeft();
```
dummdidumm, line 71: Maybe also newlines could also be trimmed here, so that 
```
<!--
@ 
```
will get the autocompletion, too

packages/language-server/src/plugins/typescript/features/getDirectiveCommentCompletions.ts
```diff
@@ -0,0 +1,79 @@
...
+            textEdit: TextEdit.replace(
+                Range.create(start, Position.create(start.line, start.character + value.length)),
+                value,
+            ),
+        }));
+
+        return CompletionList.create(items, false);
+    }
+
+    return null;
```
dummdidumm, line 78: Minor code style suggestion: Change it to

```ts
if (!match) {
   return null;
}

...
```

Then it's in line with the "return early" style of the rest of the function.

packages/language-server/src/plugins/svelte/features/getCompletions.ts
jasonlyu123: You're right, don't need that. I got what I want to achieve with `filterText` and didn't find out I don't need `isIncomplete` anymore

packages/language-server/src/plugins/svelte/features/getCompletions.ts
jasonlyu123: Seems like it does trim new line. Doesn't know that.

dummdidumm: Oh nice ^^

dummdidumm (approved):
👍

PR #562, opened 2020-09-23
Fix reactive break #559

### LINKED ISSUE #559: Syntax Error when using "break $"
**Describe the bug**

In svelte, the syntax `break $` is used to escape a reactive block (https://github.com/sveltejs/svelte/issues/2828).
However, in VSCode, using this syntax shows an error "Expression expected // ; Expected".

"break $" is transformed by svelte2tsx to :
"break __sveltets_store_get();" which is not a valid syntax

**To Reproduce**

```
<script>
  $: { break $; }
</script>
```

**Expected behavior**

The previous code should be valid in vscode.

 - OS: Windows 10
 - IDE: VSCode
 - Plugin/Package:

j3rem1e:
Fix #559 

Stores are ignored in Break node


dummdidumm (approved):
👍

PR #582, opened 2020-09-28
feat: CLI argument --fail-on-hints

bfanger:
**Feature**
`svelte-check --fail-on-hints` will exit with error code when there are hints, similar to `--fail-on-warnings`

**Use case**
I want my CI and commit hooks to fail when there are hints, that way I can prevent committing code with hints such as: 

```
/path/to/Component.svelte:10:3
Hint: 'someVariable' is declared but its value is never read. (ts)
```

### DIFF
--- a/packages/svelte2tsx/svelte-jsx.d.ts
+++ b/packages/svelte2tsx/svelte-jsx.d.ts
@@ -39,7 +39,8 @@
     //
     // Event Handler Types
     // ----------------------------------------------------------------------
-    type EventHandler<E = Event, T = HTMLElement> = (event: E & { currentTarget: EventTarget & T}) => any;
+    type EventHandler<E = Event, T = HTMLElement> =
+      (event: E & { currentTarget: EventTarget & T}) => any;
 
     type ClipboardEventHandler<T> = EventHandler<ClipboardEvent, T>;
     type CompositionEventHandler<T> = EventHandler<CompositionEvent, T>;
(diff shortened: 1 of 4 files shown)


dummdidumm:
Thanks for the PR! Could you also update the writer's `completion` method so that hints also appear in the summary? Else it might feel odd that the summary doesn't mention hints but still fails.

bfanger:
Also in the machine output?

dummdidumm:
Yes

bfanger:
I've refactored the completion method, going from 4 to 8 variations:
```
X error(s) and X warning(s)
X error(s)
X warning(s)
no errors and no warning
```
to:
```
X error(s), X warning(s) and X hint(s)
X error(s) and X warning(s)
X error(s) and X hint(s)
X error(s)
X warning(s) and X hint(s)
X warning(s)
X hint(s)
no errors, no warnings  and no hints
```

dummdidumm:
I think at this point it might make sense - also from a maintenance perspective - to simplify this to `X error(s), X warning(s) and X hint(s)` where `X >= 0`. I think this also makes it easier to read because it's consistent, all three levels are always listed.

packages/svelte2tsx/svelte-jsx.d.ts
```diff
@@ -39,7 +39,8 @@
     //
     // Event Handler Types
     // ----------------------------------------------------------------------
-    type EventHandler<E = Event, T = HTMLElement> = (event: E & { currentTarget: EventTarget & T}) => any;
+    type EventHandler<E = Event, T = HTMLElement> =
```
bfanger, line 42: This line was 102 characters, creating a linting error.

bfanger:
Simplified

dummdidumm (approved):
👍

Issue #556, opened 2020-09-23
Expand selection in VS Code doesn’t work correctly while the language mode is svelte.


KocsisGabor:
[Expand selection](https://code.visualstudio.com/docs/editor/codebasics#_shrinkexpand-selection) is `Shift+Alt+Right` by default.  

The expected behavior is demonstrated in the second half of the gif below.   
It should expand the selection by tag/scope/bracket, etc…  but works only inline and as soon as the selection should grow into multiple lines it selects the whole document.  

No alternative solutions work with any of the extensions either (e.g.: expand-region).   

![This is what happens](https://user-images.githubusercontent.com/3774059/93962975-4c786a80-fd5c-11ea-9854-50034b19158e.gif)

**System:**
 - OS: Windows
 - IDE: VSCode
 - Plugin/Package: Svelte for VSCode

jasonlyu123:
This is a feature request so I removed the bug label.

jasonlyu123:
Wondering if we should do this with svelte AST or HTML. 

Downside of the former is it won't work when there's syntax errors. Or maybe somehow we could make it fallback to HTML.

KocsisGabor:
Personally I don’t mind at all if it only works when there is no syntax error. 

This is not a real solution of course but until this issue is solved, we can use the “Emmet: Balance (outward)” command or a keyboard shortcut like this one:

```json
{
  "key": "shift+alt+right",
  "command": "editor.emmet.action.balanceOut",
  "when": "editorTextFocus && editorLangId == svelte"
}
```

This is not equivalent to “expand selection” but it is better than nothing.

MartinBspheroid:
do we have any update on this? I used to use this a lot with React and Typescript and I guess some other people as well.

PR #583, opened 2020-10-01
Selection range

jasonlyu123:
#556 
The selection range seems to support multiple cursors, so the execution should be done in an `array.map` manner. Because of that, I didn't make the `PluginHost` implementing the same interface. Haven done the typescript part. Probably need to done a similar filtering like `getDocumentSymbols`


packages/language-server/src/plugins/svelte/features/getSelectionRanges.ts
```diff
@@ -0,0 +1,68 @@
...
+    const { ast: { html } } = await svelteDoc.getCompiled();
+    const transpiled = await svelteDoc.getTranspiled();
+    const content = transpiled.getText();
+    const offset = svelteDoc.offsetAt(position);
+
+
+    const embedded = [script, style, moduleScript];
+    for (const info of embedded) {
+        if (isInTag(position, info)) {
+            // let other plugin done it
```
dummdidumm, line 26:
```suggestion
            // let other plugins do it
```

packages/language-server/src/plugins/PluginHost.ts
```diff
@@ -334,6 +335,32 @@ export class PluginHost implements LSProvider, OnWatchFileChanges {
         );
     }
 
+    async getSelectionRanges(
```
dummdidumm, line 338: Could you add a small comment above the function briefly explaining why this is a special case and the ranges have to be accumulated like this?

packages/language-server/src/plugins/svelte/features/getSelectionRanges.ts
```diff
@@ -0,0 +1,68 @@
...
+
+export async function getSelectionRange(
+    svelteDoc: SvelteDocument,
+    position: Position,
+) {
+    const { script, style, moduleScript } = svelteDoc;
+    const { ast: { html } } = await svelteDoc.getCompiled();
+    const transpiled = await svelteDoc.getTranspiled();
+    const content = transpiled.getText();
+    const offset = svelteDoc.offsetAt(position);
```
dummdidumm, line 20: The offset here is wrong. The position has to be converted to a generated position first. For that you need to enhance the mappers inside `SvelteDocument` to support `getGeneratedPosition`.

dummdidumm (changes requested):
Looks good to me. There is one bug though in `getSelectionRange` of the `SveltePlugin`: The position has to be mapped to the generated position first.

dummdidumm:
Should I do another review or are there some things missing you want to do first?

jasonlyu123:
Ok! It's ready for review again.

dummdidumm (approved):
👍

Issue #584, opened 2020-10-01
Invalid TS diagnostic mapping with context="module"


GrzegorzKazana:
Hi all 👋 

**Describe the bug**
When using typescript in components with `context="module"` diagnostics are hinted on wrong line/character.

**To Reproduce**
Create a component using TS and `module="context"` that contains type error. For example
```html
<script context="module" lang="typescript">
    const asd: string = true;
</script>
```

**Expected behavior**
Typescript errors should be placed under correct lines/characters for `context="module"` components.

**Screenshots**
Actual behavior:
![image](https://user-images.githubusercontent.com/22738928/94859352-ac45d400-0434-11eb-9cd4-27b781edd20c.png)

**System (please complete the following information):**
 - OS: MacOS
 - IDE: VSCode
 - Package: `svelte-language-server@0.10.131`
 - Plugin version: `v102.1.0`

**Additional context**
Working on a solution, might open a PR soon.

PR #585, opened 2020-10-01
fix: TS diagnostic mapping for context="module"

GrzegorzKazana:
Addresses #584 

Fixes typescript diagnostic info range for components with `context="module"`.

Before:
image

After:
image

Changes:
Take `moduleScriptInfo` into account when creating `SvelteDocumentSnapshot` mapper.


dummdidumm (approved):
Thank you very much!

PR #587, opened 2020-10-02
Fix svelte plugin diagnostic mapping for context="module"

GrzegorzKazana:
Hi again 👋 

After merging fix for the TS diagnostics misplacement within `context="module"` - #585 , I noticed this issue also occurs for svelte errors/warnings.

Before:
image

After:
image

**Changes**
Again, `moduleScriptInfo` was not taken into account when creating `SvelteFragmentMapper`. Now it is.

**Note**
In order to prevent similar errors similar to that one, I have scanned the `language-server` for references to `.scriptInfo` that are not accompanied by `|| doc.moduleScriptInfo`, and added the missing fallback.  While I tried to do my best, these changes were only based on my limited understanding of the codebase, and I would appreciate if someone more experienced could have a look into that.
To keep things clean, the first commit just solves the issue presented on the screen, and the second has the prevention measures (found via grep check, not by any observed issue). If we decide to discard the second commit, I am also fine with that.

**P.S.**
Going further with diagnostics for `context="module"`, I see that we are not handling correctly cases when component contains both `<script>...</script>` and `<script context="module">...</script>`. What are your thoughts on that? Should we wait with merging this, and try to get full handling of that as well at once, or is it fine if we go about it incrementally?

### DIFF
--- a/packages/language-server/src/plugins/svelte/SvelteDocument.ts
+++ b/packages/language-server/src/plugins/svelte/SvelteDocument.ts
@@ -175,11 +175,16 @@ export class SvelteFragmentMapper {
     }
 
     static async createScript(originalDoc: Document, transpiled: string, processed: Processed[]) {
+        const scriptInfo = originalDoc.scriptInfo || originalDoc.moduleScriptInfo;
+        const maybeScriptTag = extractScriptTags(transpiled);
+        const maybeScriptTagInfo =
+            maybeScriptTag && (maybeScriptTag.script || maybeScriptTag.moduleScript);
+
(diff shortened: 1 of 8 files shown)


dummdidumm:
Thanks a bunch for the fixes! All are legit and will prevent some future inconcistencies.
About your P.S.: Yes I noticed this a while ago, too, but did not get into looking at this more closely. My hope is that this becomes obsolete once preprocessor mapping is handled in svelte core (PR: https://github.com/sveltejs/svelte/pull/5428), so for now I just wait.

PR #594, opened 2020-10-05
Introduced the common Svelte eslint-config

ehrencrona:
This project had its own eslint config. Adding the `eslint-config` that `sapper` and `svelte` use. 

Bumped the `@typescript-eslint` version.

Keeping the rules that are stricter in this project than in the common config.

Turning off two rules (`@typescript-eslint/no-namespace`, `@typescript-eslint/no-non-null-assertion`)  from the central config that are widely violated and are non-trivial to fix.

Fixed warnings. Almost all the changes are the removed trailing commas. (the 400 new lines are from the `yarn.lock`)


ehrencrona:
Btw, we could make the trailing comma change in a separate PR if we want to get this one down to a more manageable size?

dummdidumm:
I'm okay with having all in one. About the turned off validations: We could make them level `warn` instead, so we know that's not right but the build still passes. That way new violations of this kind are less likely.

ehrencrona:
@dummdidumm Using warnings makes sense; I'll update!

Issue #493, opened 2020-08-27
Typescript control flow inside if/{#if} not working when using $store 


Evertt:
When you have a store defined as follows:

```typescript
import { readable, Readable } from "svelte/store"

const post: Readable<Post|null> = readable(null, /* some code that loads the post asynchronously */)
```

And then in your template you write:

```svelte
<h1>{$post.title}</h1>
```

Then the typescript checker rightly complains `Object is possibly 'null'`. Then I was hoping that I could make that error disappear by doing this:

```svelte
{#if $post !== null}
    <h1>{$post.title}</h1>
{/if}
```

But that has no effect whatsoever. Which kind of makes sense. I understand that typescript may not understand that that is an if-statement.

I can imagine that this would actually not be trivial to implement, but I don't know, it would be really nice.

dummdidumm:
The problem is that

```html
{#if $post !== null}
    <h1>{$post.title}</h1>
{/if}
```

gets (by us, through `svelte2tsx`, for all the intellisense goodness) translated to

```jsx
if (__sveltets_store_get(post) !== null){<>
    <h1>{__sveltets_store_get(post).title}</h1>
</>}
``` 

As you can see the store access gets translated to a `__sveltets_store_get` function, and TS's control flow therefore cannot determine this is not undefined/null. To fix this, we would need to translate this in a way that the first `__sveltets_store_get` is assigned to a variable, which is then used within the if-scope.

Evertt:
Okay, so this sounds doable? 😇

dummdidumm:
I think so

delhanty:
Just a comment to say I ran into the same issue when trying to add Typescript support to the standard Sapper template:

From [my post](https://discordapp.com/channels/457912077277855764/473466028106579978/750871546993573971) on the Discord #sapper channel:


I have a question:

Does Svelte Typescript support extend to HTML blocks like `{#each cat as cats}`?

The screenshot is from svelte-check after I turned on Typescript for the standard Sapper template.

<img width="1018" alt="Screen Shot 2020-09-03 at 9 03 34" src="https://user-images.githubusercontent.com/44455/92059602-aa083f80-edcc-11ea-8532-3170dc5ca289.png">

Here is the relevant code-snippet from `src/routes/blog/index.svelte`:

```
<ul>
    {#each posts as post}
        <!-- we're using the non-standard `rel=prefetch` attribute to
                tell Sapper to load the data for the page as soon as
[...]

PR #598, opened 2020-10-10
TS template control flow inside {#if} blocks

GrzegorzKazana:
Addresses #493 

**Before**, `{#if $store}` was not working as a flow control statement, which meant that if store value was nullable, `{#if` was not narrowing down the type.

```svelte
<script lang="ts">
  import { writable } from 'svelte/store';
  const store = writable<{ val: number } | null>(null);
</script>

{#if $store}
  <p>{$store.val}</p>
<!--  ^^^^^^^ 'Object is possibly 'null'' -->
{/if}
```

**After**, `{#if` acts as flow control statement.

```svelte
<script lang="ts">
  import { writable } from 'svelte/store';
  const store = writable<{ val: number } | null>(null);
</script>

{#if $store}
  <p>{$store.val}</p>
<!--  ^^^^^^^ No errors here -->
{/if}
```

This also works with custom type guards:
```svelte
<script lang="ts">
  import {writable} from 'svelte/store'
  const store = writable<number | string | { val: number }>('');
  const isString = (x: unknown): x is string => typeof x === 'string';
  const isNumber = (x: unknown): x is number => typeof x === 'number';
</script>

{#if isString($store)}
  <p>{$store.split('')}</p>
{:else if isNumber($store)}
  <p>{$store.toPrecision()}</p>
{:else}
  <p>{$store.val}</p>
{/if}
```

**Approach**
Previously, `svelte2tsx` was transforming 
```svelte
{#if $store}
  <p>{$store.val}</p>
{/if}
```
into roughly: 
```tsx
if (__sveltets_store_get(store)){<>
    <h1>{__sveltets_store_get(store).val}</h1>
</>}
```
Since `ts` can not safely assume that `__sveltets_store_get` is pure, `if` will not narrow down the type. In order to achieve the desired result, `__sveltets_store_get(store)` must be first assigned to a variable.
```tsx
const storeVal = __sveltets_store_get(store);
if (storeVal){<>
    <h1>{storeVal.val}</h1>
</>}
```

**Implementation**
References to store values are now hoisted to a variable `__svelte_store_get_values__` which is a record containing actual store values.
*E.g.*
```ts
const __svelte_store_get_values__ = {storeA:__sveltets_store_get(storeA),storeB:__sveltets_store_get(storeB),}
```
This declaration is injected between _script_ and _template_ section of `tsx` component.
*E.g.*
```tsx
<></>;
import {writable} from 'svelte/store'
function render() {
const store = writable<{ val: number } | null>(null);
;
() => (<>
{() => {if (__sveltets_store_get(store)){<>
    <p>{__sveltets_store_get(store).val}</p>
</>}}}</>);
```
became:
```tsx
<></>;
import {writable} from 'svelte/store'
function render() {
const store = writable<{ val: number } | null>(null);
;const __svelte_store_get_values__ = {store:__sveltets_store_get(store),};
() => (<>
{() => {if (__svelte_store_get_values__['store']){<>
    <p>{__svelte_store_get_values__['store'].val}</p>
</>}}}</>);
```

**Edge case**
A special case of accessing stores is when they are accessed inside `BlockStatements` inside event handlers. In current svelte, declaring stores outside script tag is forbidden, however based on the error message this may change in future versions.
```svelte
<button on:click={() => {
  const store = writable();
  if ($store) { /* forbidden for now, but who knows */ }
}}/>
```
The problem with such case is that we cannot hoist `__sveltets_store_get(store)` in `__svelte_store_get_values__`, because `store` is not defined in script tag scope. I had two ways of handling this:
1. Disregard the future possibility of declaring store in handler being valid, and do not worry about it (this however would potentially cause ts errors if that feature was allowed, and require to remove/update this test case `packages/svelte2tsx/test/svelte2tsx/samples/nested-$-variables-template`)
2. Do not hoist store accesses inside block statements inside component template. (Downside is that user will have to assign the store value him/herself to a contant to achieve type narrowing. This is however safer in case declaring stores would be allowed inside nested scopes, because it would not begin throwing errors).

I took the latter option, but option 1. has an advantage of being simpler (no checks if hoisting is allowed). I can _degrade_ to option 1. if you consider this additional complexity yucky 😅

### DIFF
--- a/packages/language-server/test/plugins/typescript/features/DiagnosticsProvider.test.ts
+++ b/packages/language-server/test/plugins/typescript/features/DiagnosticsProvider.test.ts
@@ -178,4 +178,11 @@ describe('DiagnosticsProvider', () => {
             },
         ]);
     });
+
+    it('treats svelte {#if ...} as typescript flow control', async () => {
+        const { plugin, document } = setup('ts-template-flow-control.svelte');
+        const diagnostics = await plugin.getDiagnostics(document);
+
+        assert.deepStrictEqual(diagnostics, []);
+    });
 });
--- a/packages/language-server/test/plugins/typescript/testfiles/ts-template-flow-control.svelte
+++ b/packages/language-server/test/plugins/typescript/testfiles/ts-template-flow-control.svelte
@@ -1,0 +1,7 @@
+<script lang="ts">
+    import {writable} from 'svelte/store'
+    const store = writable<{ val: number } | null>(null);
+</script>
+
+{#if $store}<p>{$store.val}</p>{/if}
+{#if $store !== null}<p>{$store.val}</p>{/if}
--- a/packages/svelte2tsx/src/svelte2tsx/index.ts
+++ b/packages/svelte2tsx/src/svelte2tsx/index.ts
@@ -31,6 +31,7 @@ interface CreateRenderFunctionPara extends InstanceScriptProcessResult {
     slots: Map<string, Map<string, string>>;
     events: ComponentEvents;
     isTsFile: boolean;
+    hoistedStoreDeclarations: string[];
 }
 
 interface AddComponentExportPara {
@@ -58,6 +59,7 @@ type TemplateProcessResult = {
     /** To be added later as a comment on the default class export */
     componentDocumentation: ComponentDocumentation;
     events: ComponentEvents;
+    hoistedStoreDeclarations: string[];
 };
 
 /**
@@ -77,6 +79,11 @@ function processSvelteTemplate(str: MagicString): TemplateProcessResult {
 
     //track if we are in a declaration scope
     const isDeclaration = { value: false };
+
+    // track if we are inside block statement inside template
+    // for example in multiline event handler
+    // required for determining if getting store value can be hoisted
+    let isInBlockStatementContext = 0;
 
     //track $store variables since we are only supposed to give top level scopes special treatment, and users can declare $blah variables at higher scopes
     //which prevents us just changing all instances of Identity that start with $
@@ -174,7 +181,7 @@ function processSvelteTemplate(str: MagicString): TemplateProcessResult {
                 break;
             case 'Identifier':
                 handleIdentifier(node);
-                stores.handleIdentifier(node, parent, prop);
+                stores.handleIdentifier(node, parent, prop, !isInBlockStatementContext);
                 eventHandler.handleIdentifier(node, parent, prop);
                 break;
             case 'Slot':
@@ -187,6 +194,7 @@ function processSvelteTemplate(str: MagicString): TemplateProcessResult {
                 scripts.handleScriptTag(node, parent);
                 break;
             case 'BlockStatement':
+                isInBlockStatementContext++;
                 scopeStack.push();
                 break;
             case 'FunctionDeclaration':
@@ -230,6 +238,7 @@ function processSvelteTemplate(str: MagicString): TemplateProcessResult {
 
         switch (node.type) {
             case 'BlockStatement':
+                isInBlockStatementContext--;
                 scopeStack.pop();
                 break;
             case 'FunctionDeclaration':
@@ -257,7 +266,7 @@ function processSvelteTemplate(str: MagicString): TemplateProcessResult {
     scripts.blankOtherScriptTags(str);
 
     //resolve stores
-    stores.resolveStores();
+    const hoistedStoreDeclarations = stores.resolveStores();
 
     return {
         moduleScriptTag,
@@ -268,6 +277,7 @@ function processSvelteTemplate(str: MagicString): TemplateProcessResult {
         uses$$restProps,
         uses$$slots,
         componentDocumentation,
+        hoistedStoreDeclarations,
     };
 }
 
@@ -334,6 +344,7 @@ function createRenderFunction({
     uses$$props,
     uses$$restProps,
     uses$$slots,
+    hoistedStoreDeclarations,
 }: CreateRenderFunctionPara) {
     const htmlx = str.original;
     let propsDecl = '';
@@ -354,6 +365,14 @@ function createRenderFunction({
             '});';
     }
 
+    const hoistedStoreDefinitions = hoistedStoreDeclarations
+        .map((storeName) => `${storeName}:__sveltets_store_get(${storeName}),`)
+        .join('');
+
+    const templateStoreDeclarations = hoistedStoreDefinitions
+        ? `const __svelte_store_get_values__ = {${hoistedStoreDefinitions}};\n`
+        : '';
+
     if (scriptTag) {
         //I couldn't get magicstring to let me put the script before the <> we prepend during conversion of the template to jsx, so we just close it instead
         const scriptTagEnd = htmlx.lastIndexOf('>', scriptTag.content.start) + 1;
@@ -362,11 +381,17 @@ function createRenderFunction({
 
         const scriptEndTagStart = htmlx.lastIndexOf('<', scriptTag.end - 1);
         // wrap template with callback
-        str.overwrite(scriptEndTagStart, scriptTag.end, ';\n() => (<>', {
-            contentOnly: true,
-        });
+        str.overwrite(
+            scriptEndTagStart,
+            scriptTag.end,
+            `;${templateStoreDeclarations}\n() => (<>`,
+            { contentOnly: true },
+        );
     } else {
-        str.prependRight(scriptDestination, `</>;function render() {${propsDecl}\n<>`);
+        str.prependRight(
+            scriptDestination,
+            `</>;function render() {${propsDecl}\n${templateStoreDeclarations}<>`,
+        );
     }
 
     const slotsAsDef =
@@ -410,6 +435,7 @@ export function svelte2tsx(
         uses$$restProps,
         events,
         componentDocumentation,
+        hoistedStoreDeclarations,
     } = processSvelteTemplate(str);
 
     /* Rearrange the script tags so that module is first, and instance second followed finally by the template
@@ -458,6 +484,7 @@ export function svelte2tsx(
         uses$$props,
         uses$$restProps,
         uses$$slots,
+        hoistedStoreDeclarations,
     });
 
     // we need to process the module script after the instance script has moved otherwise we get warnings about moving edited items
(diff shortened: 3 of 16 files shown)


dummdidumm (changes requested):
Thank you for the PR! Unfortunately there are some things that no longer work with this solution:

- Renaming the store will not update all its usages
- "Go to definition" does not work anymore
- Error diagnostic "X is not a store" does not work anymore

Furthermore, the solution only works for control flow statements inside the template, but the control flow problem also exists for `if ($a) { ... }` blocks inside the `script`.

To fix the first three problems, changes to the `svelte-language-server` are likely needed. To fix the last problem, adding the store map between the script and template is too late. It might be better to not use a store map and instead give each store its own variable which is defined directly after the store definition. So `let a = writable(0)` would become `let a = writable(0);let $a = svelte_ts_store_get(a);`.

GrzegorzKazana:
Thanks for catching these issues! Will try to come up with a better approach

dummdidumm:
I added some tests to have something to test against when tackling this issue.
Note that the ranges in these tests are wrong (they end one too soon) (this is a known limitation currently) so if the tests fail due to ranges being correct, the tests should be adjusted accordingly.

GrzegorzKazana:
👋 
I have played around with your suggestion - creating a variable containing store value directly after store declaration.

I managed to achieve sth like this:
```svelte
// input.svelte
<script lang="ts">
    import { writable } from 'svelte/store';
    const store = writable(0);

    if($store) {
        console.log($store)
    }
</script>

{#if $store}
    <p>{$store}</p>
{/if}
```
```ts
// output.ts
<></>;
import { writable } from 'svelte/store';
function render() {
    const store = writable(0);let $store = __sveltets_store_get(store);
    if($store) {
        console.log($store)
    }
;
() => (<>
{() => {if ($store){<>
    <p>{$store}</p>
</>}}}</>);
return { props: {}, slots: {}, getters: {}, events: {} }}

export default class Input__SvelteComponent_ extends createSvelte2TsxComponent(__sveltets_partial(__sveltets_with_any_event(render))) {
}
```

and unfortunately I struggle with the following problem:
- store diagnostic mappings broke due to the fact `__sveltets_store_get` is called in not source mapped spot. I.e.
```svelte
<script lang="ts">
let store = "not a store";
                         ^ sadly errors out only here (where let $store = __sveltets_store_get(store); is defined)
$store;
 ^^^^^ should error out here
if ($store) {}
     ^^^^^ should also error out here
</script>
```
I tried to force the errors by wrapping all references to `$store` with sth like 
```ts
$store as InferStoreValue<typeof store>
```
where `InferStoreValue` would be
```ts
type InferStoreValue<S extends SvelteStore<any>> = S extends SvelteStore<infer V> ? V : never;
```
This caused the errors to show up in desired places (well, to be honest they were one character off), and was looking very promising both in regard to the diagnostics and renaming features. But, to my great surprise, this defeated the whole purpose of this changes, because as it turns out type casted values do not work as guards.
```ts
type Thing = {val: number} | null;
const asd = null as Thing;

if (asd as Thing) {
    // asd is still nullable here
    asd.val
}
```

I am a bit running out of ideas, and start to doubt it is actually possible. The only thing that comes to my mind would be manually editing tsx sourcemaps, to map all `$store` references to position where `__sveltets_store_get(store)` is actually called. This however seems like a nasty/difficult solution (unless MagicString has a suitable feature that I am not aware of).  If you have some suggestions I would be very grateful 😅

dummdidumm:
I think there is not much you can do inside `svelte2tsx` to solve this - I'm out of ideas too. You have to go over to `language-server` and remap the positions "by hand". So you'd need to check "is this position inside a `store_get` and if so 
- run the rename one more time with the position of the original variable
- find all store usages and manually apply diagnostic messages to their ranges 
- find the original variable for find references, or all store usages if you click on the original one
So it would be custom work for each of these features.

GrzegorzKazana:
Yeah, I agree, but I am a bit hesitant to hand-crafting that logic myself inside `language-server` (I would like to avoid the need to handle such special cases here, especially I presume this would be quite complex). 

Hopefully, I just realised there is one more niche js feature that I could _(ab)use_ to go forward with the above approach - *Comma expressions*.

I could compile `$store` references into `(__sveltets_store_get(store), $store)` - which:
- will report the error as we desire, since we are still using `__sveltets_store_get`
- will enable narrowing type, since last expression is used as a value of whole comma expression - fortunately ts acknowledges that (!)
```ts
type Thing = {val: number} | null;
const asd = null as Thing;
const doSth = (a: unknown) => {};

if ((doSth(asd), asd)) {
    // no errors here
    asd.val
}
```

Will see if it sticks.

dummdidumm:
I tested this and at the first level it works, but nested checks it doesn't narrow as expected when doing it like this:

```ts
const a: SvelteStore<{h: string | number } | string> = null as any;
const $a: {h: string | number } | string =  null as any;

if (typeof (__sveltets_store_get(a), $a) !== "string") {
    const b = (__sveltets_store_get(a), $a); // {h: string | number } <- good
    if (typeof (__sveltets_store_get(a), $a).h === "string") {
        const c = (__sveltets_store_get(a), $a).h; // still string | number <- bad
    }
}
```

However, if we tweak it and move the `.h` access into `$a`, it seems to work:

```ts
if (typeof (__sveltets_store_get(a), $a) !== "string") {
    const b = (__sveltets_store_get(a), $a); // {h: string | number } <- good
    if (typeof (__sveltets_store_get(a), $a.h) === "string") {
        const c1 = $a.h;
        const c = (__sveltets_store_get(a), $a.h); // string <- good
    }
}
```

So the transformation would need to be

```
                     $a     .h
(__sveltets_store_get(a), $a.h)
```

GrzegorzKazana:
I found a dealbreaker - comma expression hack will not work correctly with custom type guards
```ts
type Thing = {val: number} | null;
const asd = null as Thing;
const doSth = (a: unknown) => {};
const isNotNull = <T extends {}>(a: T | null): a is T => a !== null;

if ((doSth(asd), asd)) {
    // no errors here
    asd.val
}

if (isNotNull((doSth(asd), asd))) {
    // Object is possibly null - but should not be.
    asd.val
}
```

Now I really think it is not possible without modifications to `language-server`. I guess I will have to go that route. This was an interesting discourse non-the-less.

dummdidumm:
FYI I opened [this bug report](https://github.com/microsoft/TypeScript/issues/41183) over at the TS repo about that type narrowing fail. Maybe wait until we have an answer and then decide how to proceed.

dummdidumm:
I also just had another idea how we could do this regardless the TS control flow comma issue: Duplicate the whole `if`-expression.

So this
`{#if $bla}` / `if ($bla)`
becomes this
`if (svelte_ts_get(bla) && $bla)`
(possibly for ternary operators, too)
Everywhere else we use the `(svelte_ts_get(bla), $bla)`-transformation.

GrzegorzKazana:
I have a vague idea why ts may not want to narrow that down - but this does not explain why in the first case it does, but not in the second one. I will explain what I mean in the ts issue.

And regarding proceeding here - I was planning to go forward with the strategy of changes to `svelte-language`, just having troubles finding enough time to get into it 😅 

About the `if` duplication, I am not sure if it will work with custom type-guards:
```svelte
{#if isString($bla)}
```
which, as I understand, would be compiled to `if (isString(svelte_ts_get(bla) && $bla))`. This will not be narrowed down I think. But I might got the idea wrong

dummdidumm:
> About the `if` duplication, I am not sure if it will work with custom type-guards:
> 
> ```svelte
> {#if isString($bla)}
> ```
> 
> which, as I understand, would be compiled to `if (isString(svelte_ts_get(bla) && $bla))`. This will not be narrowed down I think. But I might got the idea wrong

It would get transformed to `if (isString(svelte_ts_get(bla)) && isString($bla))`

dummdidumm:
The issue I opened in the TS repo was declared as a bug and a PR to fix it also already exists - so maybe we can use your comma-expression-solution afterwards 😃

dummdidumm:
Opened another issue about property access https://github.com/microsoft/TypeScript/issues/41625 . If that gets fixed, too, I think we should be able to go ahead with this comma operator solution.

dummdidumm:
Mhm we might need to go down the "do renames twice" / "modify language server" road after all. If I have 

```ts
const value = writable<{inner: number | string}>(null as any);
const $value : {inner: number | string} = null as any;

const b = (get(value), $value).inner;
```

clicking on `inner` or renaming it will lead me to/rename the `inner` of `$value`, but we need it to be `value`.

GrzegorzKazana:
Yup, I also got a bit stuck when experimenting with nightly TS builds (prior to you opening the aforementioned issue). 

Namely, I failed to handle cases when we index store value with another store value.
```js
// samples/$store-index
$storeA[$storeB]
```
ideally this would be converted to
```js
(__get(storeA), $storeA[(__get(storeB), $storeB)])
```
but ~simple~ string manipulations based on AST nodes might not do it. Moving the property access into the comma expression alone is fine, but resolving store access inside of it is not 😅 


I have checked on my partially broken local wip branch, and it seems that renaming is broken the same way you described. So yeah, I agree this may be a dead-end for this approach 🤷

dummdidumm:
Alright, I think we have consensus that this approach unfortunately will not work, we need to go the more cumbersome `const store = writable(0);let $store = __sveltets_store_get(store);` route. I will close this PR since it's about the old approach.

dummdidumm:
> 👋
> I have played around with your suggestion - creating a variable containing store value directly after store declaration.
> 
> I managed to achieve sth like this:
> 
> ```svelte
> // input.svelte
> <script lang="ts">
>     import { writable } from 'svelte/store';
>     const store = writable(0);
> 
>     if($store) {
>         console.log($store)
>     }
> </script>
> 
> {#if $store}
>     <p>{$store}</p>
> {/if}
> ```
> 
> ```ts
> // output.ts
> ///<reference types="svelte" />
> <></>;
> import { writable } from 'svelte/store';
> function render() {
>     const store = writable(0);let $store = __sveltets_store_get(store);
>     if($store) {
>         console.log($store)
>     }
> ;
> () => (<>
> {() => {if ($store){<>
>     <p>{$store}</p>
> </>}}}</>);
> return { props: {}, slots: {}, getters: {}, events: {} }}
> 
> export default class Input__SvelteComponent_ extends createSvelte2TsxComponent(__sveltets_partial(__sveltets_with_any_event(render))) {
> }
> ```
> 

@GrzegorzKazana do you have the code for that lying around somewhere? I'm interested in that because given our discussion above that's the only solution that will work for all cases. (of course we then need to handle all the other intellisense stuff, but it's a start).

GrzegorzKazana:
Sure, I will push and link my wip branch this evening 👍

GrzegorzKazana:
Here it is.
https://github.com/GrzegorzKazana/language-tools/commits/ts-flow-control-2

Basically, second commit since branching off is most important, next one ignores some new false positive diagnostics, and the last one tried to resolve nested property access when using the comma expression, so it is quite hectic 😅 

I wasn't very verbose when leaving comments, so feel free to ask 🤷

Issue #572, opened 2020-09-25
support JSDoc in .svelte files


subz390:
Svelte for VS Code `102.0.0` in VS Code for Windows 10
Would it be easy enough to implement or support JSDoc handling in `.svelte` files like it is in `.js` files in VS Code?
I hope the pictures demonstrate the issue, if not I can help explain some more, thanks.

When creating JSDoc comment blocks;
![200925-155427 no autocomplete](https://user-images.githubusercontent.com/11546815/94283862-7b9fff00-ff49-11ea-979e-7b56420f3834.png)
![200925-155601 autocomplete](https://user-images.githubusercontent.com/11546815/94283868-7e025900-ff49-11ea-80d5-809f4073982a.png)

When using JSDoc;
![200925-160345 js file all of the jsdoc](https://user-images.githubusercontent.com/11546815/94283874-7fcc1c80-ff49-11ea-9828-58b4454db73e.png)
![200925-160620 svelte only has summary](https://user-images.githubusercontent.com/11546815/94283879-8195e000-ff49-11ea-9f72-fc76bab45632.png)

jasonlyu123:
We probably have to migrate [this one](https://github.com/microsoft/vscode/blob/master/extensions/typescript-language-features/src/utils/previewer.ts#L103) to make the JSDoc hover happen, probably with the license comment. Vetur also has [it](https://github.com/vuejs/vetur/blob/master/server/src/modes/script/previewer.ts), making me a bit more comfortable copying the entire file lol.

jasonlyu123:
Quick update. I have the jsdoc comment template completion working. but not the param type, it works like typescript file.
![圖片](https://user-images.githubusercontent.com/36730922/95803839-b2408c80-0d33-11eb-83cb-d1deb18ad522.png)
It seems like typescript's language service only provides the type for jsDoc template like `@param {}` when the file extension is javascript. So until they change that behavoiur we can only provide it like typescript.

subz390:
Sounds like you have made good progress.  Something already reads the parameter type in .svelte files, see my screenshot below.
![201013-033622 the parameter types are read](https://user-images.githubusercontent.com/11546815/95808981-92926f80-0d05-11eb-876d-8f7b3538cd98.png)

So long as the typescript's language service presents all the other info from the parameters then it'll be ok.

PR #607, opened 2020-10-15
(feat) Jsdoc template completion

jasonlyu123:
Part 1 of #572 
Add snippet completion for generating JsDoc. Also auto close `/**`. Typescript currently won't provide the parameter type in the template for files other than `.js` and `.jsx`, see [here](https://github.com/microsoft/TypeScript/blob/0c7d45a9eddcec7ef2bfa1a99208de670db244d2/src/services/jsDoc.ts#L295), so this would behave like a typescript file and generate something like this:
```js
/**
 * @param abc
*/
function abc(abc){}
```


packages/language-server/src/plugins/typescript/features/getJsDocTemplateComletion.ts
```diff
@@ -0,0 +1,84 @@
...
+    let snippet: string;
+
+    // When typescript return a empty single line template
+    // return the default multi-lines snippet,
+    // making it consistent with VSCode typescript
+    if (newText === '/** */') {
+        snippet = DEFAULT_SNIPPET;
+    } else {
+        snippet = templateToSnippet(newText);
+    }
```
dummdidumm, line 53:
```suggestion
    const snippet =
        // When typescript returns an empty single line template
        // return the default multi-lines snippet,
        // making it consistent with VSCode typescript
        newText === '/** */' ? DEFAULT_SNIPPET : templateToSnippet(newText);
```

dummdidumm (approved):
Very nice! I added one minor code style change, that's all.

PR #639, opened 2020-10-29
Watch js/ts file add

jasonlyu123:
Took too long for this, alomst forgot it. There's a raw property in the result of `parseJsonConfigFileContent` that is merged with the extending `tsconfig` though it was typed as `any`.

I originally wanted to make it only read the directory where the added file was in to update the project file and has difficult implement it. So I make it read workspace directory once now. It's slower for sure. I test it on my project stored in SATA SSD it took somewhere from 50 - 100ms. Consider most of the time you would have to type something and switch file after creating a file, it seems good enough for me.

### DIFF
--- a/packages/language-server/src/plugins/typescript/TypeScriptPlugin.ts
+++ b/packages/language-server/src/plugins/typescript/TypeScriptPlugin.ts
@@ -58,6 +58,7 @@ import { convertToLocationRange, getScriptKindFromFileName, symbolKindFromString
 import { getDirectiveCommentCompletions } from './features/getDirectiveCommentCompletions';
 import { FindReferencesProviderImpl } from './features/FindReferencesProvider';
 import { SelectionRangeProviderImpl } from './features/SelectionRangeProvider';
+import { SnapshotManager } from './SnapshotManager';
 
 export class TypeScriptPlugin
     implements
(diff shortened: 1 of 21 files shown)


packages/language-server/src/plugins/typescript/TypeScriptPlugin.ts
```diff
@@ -334,24 +336,32 @@ export class TypeScriptPlugin
         return this.findReferencesProvider.findReferences(document, position, context);
     }
 
-    onWatchFileChanges(fileName: string, changeType: FileChangeType) {
-        const scriptKind = getScriptKindFromFileName(fileName);
+    onWatchFileChanges(onWatchFileChangesParas: OnWatchFileChangesPara[]) {
+        const doneUpdateProjectFiles = new Set<SnapshotManager>();
```
dummdidumm, line 340: I think this can be simplified to a simple `boolean`

dummdidumm (approved):
This is so awesome! No more "file only known after first import by hand" 🎉 
Added one minor suggestion, other than that it looks great.

packages/language-server/src/plugins/typescript/TypeScriptPlugin.ts
jasonlyu123: The reason I use a `Set` is because there might be possibility where files belongs to different `SnapshotManager`, for example a sub directory with seperate `tsconfig.json`

dummdidumm: Oh right, I forgot about that case. All good then, thanks for the explanation!

Issue #295, opened 2020-07-09
Pug syntax highlight


jasonlyu123:
**Is your feature request related to a problem? Please describe.**
Provide basic support for pug markup. Related: #106 

**Describe the solution you'd like**
VSCode's syntax highlight is done by Textmate grammar. Basically it's defining some regex pattern to tokenize the code. And we could include other languages' rules as long as vscode knows how to deal with the included language. 

I am not using pug so I don't know about pug syntax. It would be great if anyone with more experience with svelte-pug can help.

***General guide on how to implement it***
It's not hard to implement.  First, fork this repo and follow [the development instruction](https://github.com/sveltejs/language-tools#development). Follow the [vscode's official guide for syntax hightlight](https://code.visualstudio.com/api/language-extensions/syntax-highlight-guide). Add the patterns needed to the syntax file which is located in `packages/svelte-vscode/syntax`. 

It seems like syntax for the pug in svelte is similar to Vue's. So we could probably copy some code from vetur

This for defining a block to start and end identifying pug syntax:
https://github.com/vuejs/vetur/blob/master/syntaxes/vue.tmLanguage.json#L87

This for telling vscode that we have embedded pug code
https://github.com/vuejs/vetur/blob/master/package.json#L119

These two would add the basic syntax highlight. What's left would be adding svelte specific syntax like `+each('posts as post')` under pug patterns. Add more patterns and link them to the existing syntax token.
For example each block:
```json

        "svelte-pug-each": {
            "begin": "(\\+)(each)\\(",
            "beginCaptures": {
                "1": {
                    "name": "punctuation.definition.tag.begin.svelte"
                },
                "2": {
                    "name": "keyword.control.conditional"
                }
            },
            "end": "\\)",
            "endCaptures": {
                "0": {
                    "name": "punctuation.definition.tag.end.svelte"
                }
            },
            "patterns": [
                {
                    "begin": "'",
                    "end": "\\s(as)\\s+",
                    "endCaptures": {
                        "1": {
                            "name": "keyword.control"
                        }
                    },
                    "patterns": [
                        {
                            "include": "source.ts"
                        }
                    ]
                }
            ]
        }

```

jasonlyu123:
[...]

Issue #106, opened 2020-05-22
I would like Pug support, and also Stylus


BatsShadow:
My current tech stack is Angular + Pug + Stylus. I would like to explore moving to Svelte, but it would be infeasible for me to move off of Pug. Stylus is a nice bonus too.

I believe I was able to get the preprocessor to do what I needed, but the lack of decent VSCode support was a deal breaker. I'm super glad this project is taking off!

In summary: please add some sort of pug and stylus support to the language processor and vscode. Thanks!

dummdidumm:
Both are supported through https://github.com/sveltejs/svelte-preprocess. You need to setup a `svelte.config.js` for that - see the docs of Svelte plugin beta for more info. Or is there something more advanced you desire support for?

dummdidumm:
Revising my comment a little: Both are supported through `svelte-preprocess`, but the intellisense support of Svelte Beta (the plugin, not the build config) is not very good. Stylus preprocessor support is there, but it may be that you don't get the syntax highlighting that you desire. template preprocessor support is missing completely at the moment, so no pug.

So what definetely should be done is to add that some basic template preprocessor support. But on how far to go with the intellisense, I'm not sure at this point, it might just get too many things at once to support, especially if there are not ready-to-use language servers around.

vaaski:
I'm having a similar issue, my compilation with stylus works fine, but formatting in vscode doesn't work.
It seems to be trying to use prettier to format the code, but afaik prettier doesn't have support for stylus.

Coming from Vuejs, it worked fine with the Vetur extension.

I initially thought that the svelte extension treats the stylus style block as normal css, but that doesn't seem to be the case.
![image](https://user-images.githubusercontent.com/17879327/87822378-a4cd5080-c871-11ea-8bdf-9e9643f8157d.png)
Compared to not including `lang="stylus"`
![image](https://user-images.githubusercontent.com/17879327/87822441-c3cbe280-c871-11ea-8f29-7a3fc329efad.png)

But upon trying to format, it prints `Using Prettier v2.0.5 from *path*` followed by some json and `Error: unknown node type: Style` into the Svelte output console.

```JSON
{
    "type": "Style",
    "start": 69,
    "end": 152,
    "attributes": [
        {
            "start": 76,
            "end": 89,
            "type": "Attribute",
            "name": "lang",
            "value": [
                {
                    "start": 82,
                    "end": 88,
                    "type": "Text",
                    "raw": "stylus",
                    "data": "stylus"
[...]

PR #645, opened 2020-10-30
vscode: improved Svelte-Pug syntax highlighting

Monkatraz:
Practically fixes https://github.com/sveltejs/language-tools/issues/295, and helps with https://github.com/sveltejs/language-tools/issues/106

#### About
Adds improved syntax highlighting when using Pug with Svelte. It assumes the user is using `svelte-preprocessor`, as the highlighting matches against the mixins that come with that plugin.  As far as I can tell, all of the usual features of Svelte appear mostly natural in Pug now, albeit surrounded in quotation marks for anything embedded.

Please note: this is only a bunch of Textmate bundles. This does not add any additional functionality beyond highlighting and embedding, unfortunately. If less clunky Pug support was to be added, I suspect it would require a modified Pug compiler, or at least a lot of preprocessing with regex.

#### Neat Features:
* Does not clone the original Pug grammar, it instead injects into Pug whenever Pug appears inside a `source.svelte` scope
* Has absolutely no effect on normal Pug files, and probably allows for any extensions interacting with Pug to continue to do so within Svelte
* Validates interpolated `{...}` blocks in certain contexts
  - Checks for mismatched quotes on attribute assignment selections
  - Warns against using a single unsafe `=` instead of `!=`
  - Checks if you just forgot quotes entirely
* Specially handles Svelte's element directives, making it more clear about what type of value you are binding/assigning
* Has a special `{...}` interpolation case, `{{...}}` for object literals instead of statement blocks. This means that when setting opts. objects for functions the syntax highlighting will be appropriate for a JS/TS object
* Handles Pug's wacky shorthand syntax, e.g. `div: Component()`
* Handles inlined Pug, e.g. `p Some text #[+if('someVar') more text] even more text`
* It may actually be more feature-rich than Svelte's native highlighting which I find funny

#### Compromises:
* Requires three separate injected grammars to work properly. I couldn't find any other way to do this - you cannot override an embedded language from a parent language as far as I can tell.
* Most of the scopes only work on one line. This was to make the grammar less agonizing to create. I suspect in 95% of use cases this will be suitable.
* Some of the ways the highlighting works is... dubious, due to it's inherit 'duct-tape on top of duct-tape' nature. I suspect there is edge cases I haven't seen.

#### What's missing?
* Doesn't do anything special with the `svelte:[something]` elements. Doesn't break the highlighting either, though.

#### The bad:
* It's not particularly tested, but I figured I'd rather get it out there and trialed by fire. It's simply too difficult to know how the grammar will be used.
* Textmate grammars are terrible.
* I think Regex took my soul.

#### Some images
Code_RwJdTdoFvw
Code_R3OhOElzqf
Code_MB7ry1UTdL

### DIFF
--- a/packages/svelte-vscode/package.json
+++ b/packages/svelte-vscode/package.json
@@ -298,6 +298,38 @@
                     "source.ts": "typescript"
                 }
             },
+						{
+							"scopeName": "svelte.pug",
+							"path": "./syntaxes/pug-svelte.json",
+							"injectTo": [
+								"source.svelte"
+							],
+							"embeddedLanguages": {
+								"source.ts": "typescript",
+								"text.pug": "jade"
+							}
+						},			
+						{
+							"scopeName": "svelte.pug.tags",
+							"path": "./syntaxes/pug-svelte-tags.json",
+							"injectTo": [
+								"source.svelte"
+							],
+							"embeddedLanguages": {
+								"source.ts": "typescript",
+								"text.pug": "jade"
+							}
+						},			
+						{
+							"scopeName": "svelte.pug.dotblock",
+							"path": "./syntaxes/pug-svelte-dotblock.json",
+							"injectTo": [
+								"source.svelte"
+							],
+							"embeddedLanguages": {
+								"source.ts": "typescript"
+							}
+						},
             {
                 "scopeName": "markdown.svelte.codeblock",
                 "path": "./syntaxes/markdown-svelte.json",
--- a/packages/svelte-vscode/syntaxes/pug-svelte-dotblock.json
+++ b/packages/svelte-vscode/syntaxes/pug-svelte-dotblock.json
@@ -1,0 +1,55 @@
+{
+  "scopeName": "svelte.pug.dotblock",
+  "fileTypes": [],
+  "injectionSelector": "L:text.block.pug -meta.embedded.ts",
+  "patterns": [
+    {
+      "include": "#interp-object-literal"
+    },
+    {
+      "include": "#interp"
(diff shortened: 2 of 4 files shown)


dummdidumm (changes requested):
Thank you very much for this! It indeed looks better highlighting-wise than the existing "normal" Svelte highlighting. You are welcome to add some additional highlighting to that, too, if your head is not too blurry already from this PR 😄 

One thing that needs a second look though is that I now get slightly different syntax highlighting in Svelte without Pug, too. I think it's because the injectionSelector contains `meta.tag.other` which matches not just template tags, but every html tag according to `svelte.tmlanguage.json`. I'm not sure if that could just be omitted.

Monkatraz:
> One thing that needs a second look though is that I now get slightly different syntax highlighting in Svelte without Pug, too. I think it's because the injectionSelector contains `meta.tag.other` which matches not just template tags, but every html tag according to `svelte.tmlanguage.json`.

Fixed. Probably. I somehow totally blanked on making sure that injector was only specific to Pug. Funny enough, all you have to do is just prepend it with "`text.pug`" to make it work.

Injection selectors (aka scope selectors) are utterly undocumented. The more complex ones in this PR I have no idea if they're actually correct. They work, is all I know. I seriously question why VS Code chose to exclusively support Textmate grammars, when they already have a much, much better system with Monarch - with documentation. It's literally implemented into the editor, just only on the browser version.

dummdidumm (approved):
👍 
Yeah this style of writing syntax highlighting is kinda messy.

Issue #647, opened 2020-10-31
Get an error when binding to exported const variable


dkzlv:
**Describe the bug**
If you use two-way binding with an exported const variable, it shows an error.

**To Reproduce**
Svelte allows to use two-way binding syntax to bind a local variable to component's exported const variable. [REPL](https://svelte.dev/repl/03e3a418fe2a48f5aff3a4db3ee4b138?version=3.29.4).

Currently the plugin will show an error, saying `Property <> does not exist on type IntrinsicAttributes & ...`.

**Expected behavior**
I didn't know you can do that (usually I bind component instance to a variable and then access the exported variable through instance), but if Svelte supports it so should the plugin, I guess.

**System (please complete the following information):**
 - OS: OS X
 - IDE: VSCode latest
 - Plugin/Package: Svelte for VS Code

PR #649, opened 2020-11-02
export const, function and class binding

jasonlyu123:
#647 
We might have to try finding a way to typecheck it later. Currently can't tell if that's a binding or normal props.


dummdidumm:
Do you remember why we originally did not mark these as possible props?

jasonlyu123:
Yeah, I guess it's because of the warning if you use it as a input props. Doesn't know it can be binded, didn't see it on the docs.

dummdidumm:
This and the if-scoping feels to me like the "convert to JSX"-approach has some limitations we increasingly have to hack around against. Not sure if we have good alternatives though.

PR #653, opened 2020-11-04
(feat) typescript signature help

jasonlyu123:
Here I am again with another mostly copy-pasted feature XD

圖片

### DIFF
--- a/packages/svelte-vscode/package.json
+++ b/packages/svelte-vscode/package.json
@@ -113,8 +113,14 @@
                 "svelte.plugin.typescript.selectionRange.enable": {
                     "type": "boolean",
                     "default": true,
-                    "title": "TypeScript: SelectionRange",
+                    "title": "TypeScript: Selection Range",
                     "description": "Enable selection range for TypeScript"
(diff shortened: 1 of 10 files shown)


packages/svelte-vscode/package.json
```diff
@@ -113,9 +113,15 @@
                 "svelte.plugin.typescript.selectionRange.enable": {
                     "type": "boolean",
                     "default": true,
-                    "title": "TypeScript: SelectionRange",
+                    "title": "TypeScript: Selection Range",
                     "description": "Enable selection range for TypeScript"
                 },
+                "svelte.plugin.typescript.signatureHelp.enable": {
```
jasonlyu123, line 119: Wondering if we should use the LSP name in the config or just `parameter hints`, maybe that's more understandable.

dummdidumm: I prefer the LSP name just because they put some thought into the name (I hope 😄 ) so I would go with it.

dummdidumm (approved):
👍

PR #654, opened 2020-11-05
[svelte2tsx] Add comments to blanked script/style tags for perf

halfnelson:
jasonlyu09 from discord discovered Svelte's parse method is very slow when scripts contain lots of whitespace. 

We assume it is something to do with a regex that cleans trailing spaces. (may be similar to this https://github.com/sindresorhus/trim-right/pull/3)

Instead of changing the regex in svelte, we just stop making too much white space by introducing `/**/` comments at the end of lines when blanking out the script tag.

Added a test.

With fix:
image

Without fix:
image


dummdidumm (approved):
👍 
FYI @jasonlyu123

Issue #309, opened 2020-07-15
Color Picker not working in VS Code with Svelte Extension


ejoflo:
According to the feature list for the Svelte for VS Code extension:

CSS / SCSS / LESS
"Color highlighting and color picker"

However, when I open a .svelte file, and add a color or background-color property to under the <style> tag of said .svelte file, there is no color picker displayed.

The settings have the following enabled:
svelte.plugin.css.colorPresentations.enable - TRUE

jasonlyu123:
hmm... that works for me. Can you provide the content in your style?

ejoflo:
Here's a screenshot of a .svelte file, notice that there is no color picker on the background-color property.

![Screen Shot 2020-07-14 at 10 40 57 PM](https://user-images.githubusercontent.com/51680874/87523860-2da27d00-c623-11ea-83b1-cf0d344bf53f.png)

jasonlyu123:
oh! only the top-level one would have that. If it's under another tag it won't have CSS language feature. Only the component style would.

ejoflo:
> oh! only the top-level one would have that. If it's under another tag it won't have CSS language feature. Only the component style would.

Thanks for the reply. Could you clarify what you mean by "top-level"? The style tag above is inside of a Header.svelte tag. Is that not the same as a "component style"? Thank you.

jasonlyu123:
not a child of any tag. 
```svelte
<!-- this one -->
<style>

</style>

<svelte:head>
	<!-- not this one -->
	<style>
		.a {}
	</style>
</svelte:head>
```
https://svelte.dev/repl/6bdac66b78064f57ab2d12e425ff60f1?version=3.24.0

ejoflo:
This bug seems to be related to the Svelte prettier plugin. My current configuration is as follows:

 "svelteSortOrder" : "scripts-markup-styles",

Since my <style> tag is located at the bottom of my .svelte file, it loses the color picker. When I move the <style> tag all the way to the top of the file, the color picker re-appears.

jasonlyu123:
if it's not under any tag? Do you have any self-close tag before it?

ejoflo:
No, it is not under any tag. The <style> tag is independent.

jasonlyu123:
Can you provide the entire file or just the reproduable

ejoflo:
> Can you provide the entire file or just the reproduable

Sure, here it is on REPL:
[...]

PR #671, opened 2020-11-13
ignore html end-tag-like inside moustache

jasonlyu123:
related #309 

kind of hacky, but it would fix some situation where style tag not being extracted when it's after a self-closed tag with a arrow function props or a event handler.
```svelte
<Foo on:click={() => console.log('yeah!!!')} />
<style></style>
```
There're cases like this:
```svelte
<Foo selected={a > b} />
```
not being covered, But I think dealing with that with this method would have a higher change removing the actual tag end.

Another approach is to use `html-language-service`'s internal api, like [this branch](https://github.com/jasonlyu123/language-tools/blob/enhance-parser-internal/packages/language-server/src/lib/documents/parseHtml.ts)

Or we can alter the `html-language-service`'s `htmlScanner`. [vetur](https://github.com/vuejs/vetur/blob/master/server/src/modes/template/parser/htmlScanner.ts) has a similar scanner to [it](https://github.com/microsoft/vscode-html-languageservice/blob/master/src/parser/htmlScanner.ts).

### DIFF
--- a/packages/language-server/src/lib/documents/Document.ts
+++ b/packages/language-server/src/lib/documents/Document.ts
@@ -1,6 +1,7 @@
 import { urlToPath } from '../../utils';
 import { WritableDocument } from './DocumentBase';
-import { extractScriptTags, extractStyleTag, TagInformation, parseHtml } from './utils';
+import { extractScriptTags, extractStyleTag, TagInformation } from './utils';
+import { parseHtml } from './parseHtml';
 import { SvelteConfig, loadConfig } from './configLoader';
 import { HTMLDocument } from 'vscode-html-languageservice';
 
--- a/packages/language-server/src/lib/documents/parseHtml.ts
+++ b/packages/language-server/src/lib/documents/parseHtml.ts
@@ -28,7 +28,7 @@ const createScanner = parser.createScanner as (
 ) => Scanner;
 
 /**
- * scan the text and remove any `>` that cause the tag to end short
+ * scan the text and remove any `>` or `<` that cause the tag to end short,
  */
 function preprocess(text: string) {
     let scanner = createScanner(text);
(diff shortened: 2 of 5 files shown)


dummdidumm:
Altering the Scanner seems more robust but more maintenance work in the future. Your preprocessor step looks good for me since it only blanks moustache tags if I understand that correctly (maybe you could add a short JSDoc above the preprocessor function explaining why it's needed and what is done). The only downside is that it may be a very small performance hit since we now parse the string twice, but I guess that's negligible.
(Maybe we need some kind of performance tests some day)

jasonlyu123:
It actally not only blank moustache. This scanner basically goes through the code and compare the code character by character to given char code and condition. 
```svelte
<a on:click={() =>| console.log('a')}></a>
|<div></div>
```
if the scanner sees a '>' inside a tag it would think the tag is closed. So until the next token would be the next open tag, the content (for example the code between two `|` ) would all be considered as text content. By using this method it would blank the valid end of start/end tag. that's why it can only deal with `=>` and `<=`

The internal code of the `html-language-service` actually has an `initialState` parameter by using that we can reset the state and make it restart the scan like it's still within the tag. it would have a higher chance of finding out the actual end of the start tag. it could cover the bigger-than case:
```
<Foo selected={a > b} />
```

jasonlyu123:
I use the version utilized `html-language-service` internal API instead

dummdidumm (changes requested):
Could you add two more test cases testing the "normal" cases (html without a moustache tag, html with a moustache tag but no > character)?
Also does it work correctly for `<div checked={a < 1}></div><style></style>`? A test for that would be good, too.

jasonlyu123:
Thanks! added the support

dummdidumm (approved):

Issue #477, opened 2020-08-23
Cannot use the <...> notation to typecast when using lang="ts"


dummdidumm:
**Describe the bug**
One cannot use the `<...>` notation to typecast when using `lang="ts"`. Because we transform the code to TSX, we get a "JSX element '...' has no corresponding closing tag." error.

**To Reproduce**
```html
<script lang="ts">
    const blubb = <any>null;
</script>
```

**Expected behavior**
We should modify the code in a way that this works (maybe using `MagicString.move` to transform it to `as` syntax).

**System (please complete the following information):**
 - OS: Windows
 - IDE: VSCode
 - Plugin/Package: Svelte for VSCode", `svelte2tsx`

**Additional context**
Noone brought this up yet so it does not seem that much of an issue.

PR #679, opened 2020-11-18
(svelte2tsx)Type assertion

jasonlyu123:
#477 
transform type assertion to "as" expression
```typescript
<Type>a
```
to
```typescript
a as Type
```
While it works, because the source map would map `<Type>a|` to `a as Type|` (`|` being cursor) so the autocomplete won't work. I can't find a way to improve it.


dummdidumm:
Yeah these kind of position errors are in various places, for example in `$: ` lines. MagicString would need to provide a way to add the source map one character later before all else that one character later on.

PR #681, opened 2020-11-20
(feat) ts import user preferences

jasonlyu123:
Add the ability to config import user preference: `importModuleSpecifierPreference` and `importModuleSpecifierEnding`. 
The former is shortened to `importModuleSpecifier` in the config to make it consistent with vscode's "javascript.preferences.importModuleSpecifier" setting.

Auto-import can now be configured to always use path alias (`importModuleSpecifierPreference: 'non-relative'` ) like this

```svelte
<script>
import {} from '~/utils/abc'
</script>
```
The default behavior is to use the shorter path, for example:
```svelte
<script>
import {} from '~/utils/abc'
import {} from './abc'
</script>
```
in this case, it would use relative.


dummdidumm:
Nice feature! But I think it would be better if we would reuse the vscode settings for js/ts. After all, js/ts is embedded into the Svelte script blog so for me it makes sense to reuse its settings, too, then. What do you think?

jasonlyu123:
That would also be ok. It's just there are separate settings for ts and js. Which one should we use? if both who take priority?

dummdidumm:
Why vscode, whyyyyy 😄 if we got access to the document we also got access to whether or not it's lang=ts, so we could branch on that.

packages/language-server/src/plugins/typescript/TypeScriptPlugin.ts
```diff
@@ -419,4 +424,18 @@ export class TypeScriptPlugin
             this.configManager.enabled(`typescript.${feature}.enable`)
         );
     }
+
+    private getUserPreferences(): ts.UserPreferences {
```
dummdidumm, line 428: Maybe we could put this into the lsandtsdocresolver instead? Everyone already has this as a dependency so we don't need to add a new one. Semantically I also think it makes sense in there. Maybe the getLsAndTsDoc function in there could even return `preferences` as a third object.

jasonlyu123:
Since it's using ts/js preferences config of vscode now, should we also check `quoteStyle`?

dummdidumm:
Where would we check it?

jasonlyu123:
it's also a preference config in vscode `javascript.preferences.quoteStyle`

dummdidumm:
And where would we use it? Since it's part of the `preferences` object, don't we now use it already then?

jasonlyu123:
I added it. The config name of vscode is different from the `UserPreferences` of typescript.

PR #719, opened 2020-12-16
(feat) control flow for stores

dummdidumm:
#493
$store --> (__svelte_store_get(store), $store)
By using the comma operand we allow TypeScript's control flow to work. Before that, it was $store --> __svelte_store_get(store) which loses control flow info every time.

We need to wait for TS 4.2 before we can merge this because we need a not-yet-published control flow fix for this solution to work.
I tested rename, go-to-declaration, find-references, hover - all looks good. But a second look once this is finalized would be good.

Props to @GrzegorzKazana from which most of this code is - thanks!


GrzegorzKazana:
This is great - can't wait for it to be released! 🚀 
Kudos for not giving up on this approach 😅

dummdidumm:
Yes, my objection in the now close PR of yours "this won't work because X" was actually unjustified, because my code example was wrong: It said `let $bla: {...sometype} = ...`, but it's `let $bla = __svelte...` which means TS is smart enough to defer all the type intellisense to `bla`. So most of your code was already right and working. And since TS 4.2 will make `(get(bla), $bla).blubb` possible with type narrowing (I just checked with TS nightly and can confirm it works), we can ditch the more cumbersome `(get(bla), $bla.blubb)`-solution.
So now we just wait 😄

PR #737, opened 2021-01-04
svelte2tsx/test 

pushkine:
Extracted `htmlx2jsx` and `svelte2tsx` tests common code into a `test_sample` function in `helpers.js`
Added check for random unused files in samples (warns locally, throws in CI)
Added auto generation of the expected output if absent (marks test as skipped and logs a notice at the end locally, throws in CI)
Made samples starting with a dot show up as skipped tests instead of silent

Made the `htmlxparser` timeout test generate its input dynamically

Removed checks for `.solo` folders in 3 places for a single CI-only match-all glob in index


packages/svelte2tsx/test/helpers.js
```diff
@@ -0,0 +1,71 @@
+const fs = require("fs");
+const assert = require("assert");
+function benchmark(fn) {
+	return -Date.now() + (fn(), Date.now());
+}
+function readFileSync(path) {
+	return fs.existsSync(path) ? fs.readFileSync(path, "utf-8").replace(/\r\n/g, "\n").replace(/\s+$/, "") : null;
+}
+function check_dir(path, { allow, require = allow }) {
```
dummdidumm, line 9: Can we change this to `required` and `allowed`? I was thinking of functions (especially for `require`) first.

packages/svelte2tsx/test/helpers.js
```diff
@@ -0,0 +1,71 @@
...
+	if (allow !== require) allow = [...require, ...allow];
```
dummdidumm, line 10: I prefer these to have a separate line.
```suggestion
	if (allow !== require) {
		allow = [...require, ...allow];
	}
```

packages/svelte2tsx/test/helpers.js
```diff
@@ -0,0 +1,71 @@
+const fs = require("fs");
+const assert = require("assert");
```
dummdidumm, line 2: Small nit: empty line between functions

packages/svelte2tsx/test/helpers.js
```diff
@@ -0,0 +1,71 @@
...
+function readFileSync(path) {
+	return fs.existsSync(path) ? fs.readFileSync(path, "utf-8").replace(/\r\n/g, "\n").replace(/\s+$/, "") : null;
+}
+function check_dir(path, { allow, require = allow }) {
+	if (allow !== require) allow = [...require, ...allow];
+	const unchecked = new Set(require);
+	const unknown = new Set();
+	a: for (const fileName of fs.readdirSync(path)) {
+		for (const name of unchecked)
+			if (name === fileName || (name.startsWith("*") && fileName.endsWith(name.slice(1)))) {
```
dummdidumm, line 15: Why the `startsWith("*")`?

packages/svelte2tsx/test/helpers.js
```diff
@@ -0,0 +1,71 @@
...
+			}
+		if (!allow.includes(fileName)) unknown.add(fileName);
+	}
+
+	if (unknown.size) {
+		after(() => {
+			for (const f of unknown) {
+				const file = `${path.split("/").slice(-1)}/${f}`
+				if (process.env.CI) throw new Error(`Unexpected file ${file}`);
+				console.info(`Ignored file ${file}`);
```
dummdidumm, line 27:
```suggestion
				const errorMsg = `Unexpected file ${file}`;
				if (process.env.CI) {
					throw new Error(errorMsg);
				} else {
					console.info(errorMsg);
				}
```

packages/svelte2tsx/test/helpers.js
```diff
@@ -0,0 +1,71 @@
...
+}
+function check_dir(path, { allow, require = allow }) {
+	if (allow !== require) allow = [...require, ...allow];
+	const unchecked = new Set(require);
+	const unknown = new Set();
+	a: for (const fileName of fs.readdirSync(path)) {
+		for (const name of unchecked)
+			if (name === fileName || (name.startsWith("*") && fileName.endsWith(name.slice(1)))) {
+				unchecked.delete(name);
+				continue a;
```
dummdidumm, line 17: Can we remove that `continue` statement and make the other `if` an `else if` instead?

dummdidumm (changes requested):
These are good changes! I left some minor refactoring comments in there. Also it seems that you used tabs instead of spaces throughout. I don't know if you use prettier, but if you do, it should pick up the configuration if you open the workspace from the root.

packages/svelte2tsx/test/helpers.js
pushkine: dirty way to support checking for a file format
an input with `'*.svelte'` would check for files that end with `.svelte` instead of trying an exact match

packages/svelte2tsx/test/helpers.js
pushkine: I restructured it and corrected some things but I don't think you can get much better due to the wildcard thing

pushkine:
> it should pick up the configuration if you open the workspace from the root

Ah, svelte2tsx/test was in `.prettierignore`, I removed it if that's okay

dummdidumm:
> > it should pick up the configuration if you open the workspace from the root
> 
> Ah, svelte2tsx/test was in `.prettierignore`, I removed it if that's okay

Could you alter it so that the tsx/jsx files are ignored? Otherwise it might be very cumbersome to change these files.

dummdidumm (approved):
Thanks!

PR #735, opened 2021-01-04
avoid get on $property

pushkine:
#691

PR #740, opened 2021-01-06
chore: format repository with tabs

ignatiusmb:
> This repo uses spaces for intendation (maybe we should align that with the main repo some day ...)

_Originally posted by @dummdidumm in https://github.com/sveltejs/language-tools/pull/735#discussion_r551332594_

***

The first commit formats all the files using the existing config and the third commit formats it again after `useTabs` is changed to true. There's a couple of file extensions intentionally not included like

- `package.json` files, which is designed to uses spaces (main repo too)
- `*.md` files, which the Svelte code blocks now gets formatted too and usually uses [markdownlint](https://github.com/DavidAnson/markdownlint) to format it as well
- `*.json` files, feels weird to format with tabs and 4 spaces

Proposed changes

- [x] always use arrow parens (applied on the first commit, not sure why)
- [ ] change `tabWidth` to 2 as well
- [ ] add [markdownlint config](https://github.com/DavidAnson/vscode-markdownlint#markdownlintconfig) file to format markdowns


packages/svelte-vscode/src/html/autoClose.ts
```diff
@@ -67,17 +60,23 @@ export function activateTagClosing(
         }
         const lastChange = changes[changes.length - 1];
         const lastCharacter = lastChange.text[lastChange.text.length - 1];
-        if ('range' in lastChange && (lastChange.rangeLength ?? 0) > 0 || (lastCharacter !== '>' && lastCharacter !== '/')) {
+        if (
+            ('range' in lastChange && (lastChange.rangeLength ?? 0) > 0) ||
+            (lastCharacter !== '>' && lastCharacter !== '/')
```
ignatiusmb: I've ran the tests locally and double checked everything, this is the only thing I'm uncertain since it adds another parentheses for the first two conditionals. It should be the same, except when `'range' in lastChange` is false, it will still check the next conditional after the `||`, while previously it would fail immediately?

dummdidumm: `&&` has priority over `||` so this should be the same

dummdidumm:
When you say "intentionally not included", what do you mean by that? Prettier did format them but you did not include the change?

ignatiusmb:
Yes, I formatted the whole repo and excluded those files from the commit. I was hoping to get some feedback first before I add any additional config settings here.

For `*.json` files, they're formatted as expected. But, `package.json` could be considered an exception where it usually uses 2 spaces for indentation. I could add a config to override only `package.json`s (or all `.json`) to not use tabs and 2 spaces indentation. Do we want to use tabs for all except `package.json` or enforce the same style as `package.json` for `.json` files?

For `*.md` files, they're still formatted... but, it's result is inconsistent. It won't format where there's a mix of space and tabs[1], code blocks are messed up[2], and for some reason it decides to format list items to have 3 spaces between the first character[3]. I decided to leave it for the time being, but I could fix it manually since they're relatively few in number (a total of 10 files), perhaps in a different PR.

| [1] | [2] | [3] |
|-----|-----|-----|
| image | image | image |

ignatiusmb:
Sorry this took a while to get back to, but I think I finally got it! 

Looking at the main repo for reference, it seems that, `.json` files are indeed formatted with tabs and 4 spaces too. I've also strike the last 2 points as 2 spaces `tabWidth` only applies to `package.json`. As for the markdown files, I still think it's better to use markdownlint to help lint them if we're going to fully utilize `.md` for documentations, but that can wait in another PR.

ignatiusmb:
In some rare cases, prettier will delete some lines in code blocks with Svelte language identifier. Somehow, explicitly setting the sort order helps calm it down.

PR #741, opened 2021-01-06
Add missing attribute as

guidobouman:
The `as` attribute is used by the `link` element: `<link rel="preload" src="my-video.mp4" as="video">`.

On another note: Do we really need an extra copy of all this? Can we not use the TypeScript-native implementation?

### DIFF
--- a/packages/svelte2tsx/svelte-jsx.d.ts
+++ b/packages/svelte2tsx/svelte-jsx.d.ts
@@ -232,6 +232,7 @@ declare namespace svelte.JSX {
       allowtransparency?: boolean;
       allowpaymentrequest?: boolean;
       alt?: string;
+      as?: string;
       async?: boolean;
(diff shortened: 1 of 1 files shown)


jasonlyu123:
What do you mean by the typescript-native implementation?

dummdidumm:
I think he means the TS `dom.lib.d.ts` (or sth like that) type definitions. I don't think we can use them though because they contain the typings for programmatic interaction with the elements, which means they also contain things like `addEventListener`, and some attribute names are different (`className` instead of `class`).

guidobouman:
I was referring to `lib.dom.d.ts`, yes. The programmatic interaction is problematic. Although, maybe we can strip that part? 

About the other attributes: Svelte does not use `className`, right? Or is this not used by the language server / html markup checker?

PR #744, opened 2021-01-07
mini repl for svelte2tsx

pushkine:
Adds a svelte file under svelte2tsx/repl that automatically compiles its content to tsx on dev mode
Adds a `debug` package script, attaching a debugger to it will inspect the compiling process of the repl directly in the ts source


packages/svelte2tsx/package.json
```diff
@@ -25,13 +29,10 @@
...
         "periscopic": "^2.0.2",
         "rollup": "^2.28.0",
-        "@rollup/plugin-commonjs": "^15.0.0",
         "rollup-plugin-delete": "^2.0.0",
-        "@rollup/plugin-json": "^4.0.0",
-        "@rollup/plugin-node-resolve": "^9.0.0",
-        "@rollup/plugin-typescript": "^6.0.0",
         "source-map": "^0.6.1",
         "source-map-support": "^0.5.16",
+        "sucrase": "^3.17.0",
```
dummdidumm, line 35: Could we use `ts-node` instread? We already have that in this repo and it might serve the same use case and we avoid another dependency.

packages/svelte2tsx/rollup.config.js
```diff
@@ -1,9 +1,33 @@
...
 import commonjs from '@rollup/plugin-commonjs';
-import resolve from '@rollup/plugin-node-resolve';
 import json from '@rollup/plugin-json';
+import resolve from '@rollup/plugin-node-resolve';
+import typescript from '@rollup/plugin-typescript';
 import builtins from 'builtin-modules';
+import fs from 'fs';
+import path from 'path';
+
+function repl() {
```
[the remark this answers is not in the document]
pushkine: This part uses rollup watch, so just by having the dev script running in the background it automatically updates the repl results on save. i.e. Everything is instant, just hit ctrl+s on any change anywhere ( source or repl input ) and the results get live updates like an actual repl would

It is true that sucrase computes the same thing but you have to run it yourself and checking values in the debugger is a much slower process

packages/svelte2tsx/package.json
pushkine: I agree and I tried many things, but ts-node refuses to work due to a side issue in tsconfig
Sucrase just works, it's also way faster

packages/svelte2tsx/rollup.config.js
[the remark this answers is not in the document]
pushkine: On a side note it only outputs tsx right now, I imagine there are other outputs that could be interesting to print

dummdidumm:
I got this running insive VS Code with this addition to `launch.json`:

```json
{
            "type": "node",
            "request": "launch",
            "name": "Run REPL with debugger",
            "runtimeArgs": ["-r", "ts-node/register"],
            "args": ["${workspaceFolder}/packages/svelte2tsx/repl/debug.ts"],
            "env": {
                "TS_NODE_COMPILER_OPTIONS": "{\"esModuleInterop\":true, \"target\": \"es2018\"}",
                "TS_NODE_TRANSPILE_ONLY": "true"
            },
            "console": "integratedTerminal",
            "internalConsoleOptions": "neverOpen"
        },
```
Does this work for you, too? If so, I would propose ditching the debug script inside package.json and instead use that launch script.

packages/svelte2tsx/repl/debug.ts
```diff
@@ -0,0 +1,13 @@
+import fs from 'fs';
+import svelte2tsx from '../src';
+const content = fs.readFileSync(`${__dirname}/index.svelte`, 'utf-8');
+svelte2tsx(content);
+/**
+ * To enable the REPL, simply run the "dev" package script.
+ *
+ * The "/repl/index.svelte" file will be converted to tsx
```
dummdidumm, line 8: This comment needs to be updated since you now use the vs code debug functionality

packages/svelte2tsx/src/htmlxtojsx/index.ts
```diff
@@ -1,12 +1,11 @@
 import { Node } from 'estree-walker';
 import MagicString from 'magic-string';
-import svelte from 'svelte/compiler';
+import * as svelte from 'svelte/compiler';
```
dummdidumm, line 3: I think these import changes can be reverted because "esModuleInterop" should take care of it.

packages/svelte2tsx/repl/debug.ts
pushkine: The REPL still updates through rollup, the debug thing is secondary

packages/svelte2tsx/src/htmlxtojsx/index.ts
pushkine: nope 
Code_2021-02-25_07-47-53

dummdidumm: I stand corrected. After some digging I think this is actually correct behavior when not using Rollup's interop magic, since the compiler has no default export. What could be changed though is to use a named import here and in the other location instead of `* as svelte`, because in both instances only parse/walk is used. If you could do that and resolve the merge conflict then I'm good to merge this.

Issue #739, opened 2021-01-06
Cannot use @debug block with multiple arguments


dkzlv:
**Describe the bug**
If you pass multiple arguments to `{@debug}` it will result in an error (will compile though).

**To Reproduce**
```svelte
<script lang="ts">
  let a = 1,
    b = 2;
</script>

<!-- Error: JSX expressions may not use the comma operator. Did you mean to write an array? -->
{@debug a, b}
```

**System (please complete the following information):**
 - OS: Mac OS
 - IDE: VSCode
 - Plugin/Package: Svelte for VSCode

PR #746, opened 2021-01-07
fix comma operator in debug

pushkine:
#739


jasonlyu123 (changes requested):
Thanks for your contribution. However, we might need a more complex solution on transformation for better sourcemap.

dummdidumm (approved):
Thanks!

dummdidumm (changes requested):
Found a bug in combination with `$store` usage. `{@debug $store}` is translated to `__sveltets_store_get(myfile)}`. Notice the missing `{` at the start. I guess either the `remove` in `debug` removes it or the inner store transformation. One way to fix this could be to remove the `@debug` like it was done before and overwrite only the `,` inbetween with `}{`.

pushkine:
It's related to the MagicString API, I have no idea how to use prepend/append left/right, using prependLeft on both seems to cover the bug you found

dummdidumm (approved):

Issue #71, opened 2020-05-08
Add support for Semantic Highlighting (Tokens)


icebr9:
**Is your feature request related to a problem? Please describe.**
Implement semantic tokens so theme makers can apply Svelte-specific styling in their themes. (This is a feature since VSCode 1.43)
[https://code.visualstudio.com/api/language-extensions/semantic-highlight-guide#theming](https://code.visualstudio.com/api/language-extensions/semantic-highlight-guide#theming)
Following addition, guidelines/docs for theme makers should be added to the plugin description so they can target these tokens.

**Describe the solution you'd like**
Sample implementation can be found here, [https://github.com/Microsoft/vscode-extension-samples/tree/master/semantic-tokens-sample](https://github.com/Microsoft/vscode-extension-samples/tree/master/semantic-tokens-sample)

**Describe alternatives you've considered**
Make it a seperate optional plugin instead of integrating into `Svelte Beta`.

PR #752, opened 2021-01-11
(feat) Semantic tokens

jasonlyu123:
related to #71 

This implements the [semantic token](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_semanticTokens) for typescript. And because of the [typescript implementation](https://github.com/microsoft/TypeScript/blob/f1dca6a61f577bfb83dbc91d2e064057880df209/src/services/services.ts#L1863), this feature doesn't work with javascript. 

I tried to add a special highlight to the event dispatcher function but the event semantic token type is not supported by a lot of themes. It ends up not highlighted. I didn't include it in the PR but maybe we could make it optional and hide it behind a setting so that people with the supported theme can enable it.

The original feature request asks for svelte specific highlight but I didn't think of any svelte specific semantic token type. Most of the special syntax doesn't have much to be semantically analyzed. The action, transition, and animation are all supposed to be a function. What do you guys think? Is there any syntax you would like to be covered in semantic tokens?

## Screenshot:  
**Theme: GitHub dark**

Without semantic tokens
圖片
With semantic tokens
圖片


packages/language-server/src/plugins/PluginHost.ts
```diff
@@ -400,6 +402,20 @@ export class PluginHost implements LSProvider, OnWatchFileChanges {
...
+        const document = this.getDocument(textDocument.uri);
+        if (!document) {
+            throw new Error('Cannot call methods on an unopened document');
+        }
+
+        return this.execute<SemanticTokens>(
+            'getSemanticTokens',
+            [document, range],
+            ExecuteMode.Collect,
+            collectSemanticTokens
```
jasonlyu123, line 415: This can be removed and changed to `ExecuteMode.FirstNotNull` if we decide not to include svelte semantic tokens.

packages/language-server/src/plugins/collector/collectSemanticToken.ts
```diff
@@ -0,0 +1,18 @@
+import { Range, SemanticTokens } from 'vscode-languageserver';
```
dummdidumm, line 1: Can we inline this function into `PluginHost`? While I appreciate moving code into separate files, I think in this instance it breaks the nice "all folders within the plugins folder are a plugin"-semantic. If I understand your other code comment we might even remove this altogether since only the TS plugin provides semantic tokens.

packages/language-server/src/plugins/typescript/features/SemanticTokensProvider.ts
```diff
@@ -0,0 +1,106 @@
...
+                generatedOffset,
+                generatedLength
+            );
+            if (!originalPosition) {
+                continue;
+            }
+
+            const [line, character, length] = originalPosition;
+
+            // remove identifers whose start and end mapped to the some location
```
dummdidumm, line 57:
```suggestion
            // remove identifers whose start and end mapped to the same location
```

packages/language-server/src/plugins/typescript/features/SemanticTokensProvider.ts
```diff
@@ -0,0 +1,106 @@
...
+            fragment.positionAt(generatedOffset + generatedLength)
+        );
+        const startOffset = document.offsetAt(startPosition);
+        const endOffset = document.offsetAt(endPosition);
+
+        return [startPosition.line, startPosition.character, endOffset - startOffset];
+    }
+
+    /** TSClassification =
+     *  (TokenType + 1) << TokenEncodingConsts.typeOffset + TokenModifier */
```
dummdidumm, line 93:
```suggestion
    /**
     *  TSClassification = (TokenType + 1) << TokenEncodingConsts.typeOffset + TokenModifier
     */
```

packages/svelte-vscode/README.md
```diff
@@ -120,6 +120,10 @@ Enable rename functionality for JS/TS variables inside Svelte files. _Default_:
 
 Enable signature help (parameter hints) for JS/TS. _Default_: `true`
 
+##### `svelte.plugin.typescript.semanticTokens.enable`
+
+Enable semantic tokens (semantic highlight) for Typescript. Doesn't applied to JavaScript. _Default_: `true`
```
dummdidumm, line 125:
```suggestion
Enable semantic tokens (semantic highlight) for TypeScript. Doesn't apply to JavaScript. _Default_: `true`
```

packages/svelte-vscode/package.json
```diff
@@ -130,6 +130,12 @@
                     "title": "TypeScript: Rename",
                     "description": "Enable rename functionality for JS/TS variables inside Svelte files"
                 },
+                "svelte.plugin.typescript.semanticTokens.enable": {
+                  "type": "boolean",
+                  "default": true,
+                  "title": "TypeScript: Semantic Tokens",
+                  "description": "Enable semantic tokens (semantic highlight) for Typescript. Doesn't applied to JavaScript"
```
dummdidumm, line 137:
```suggestion
                  "description": "Enable semantic tokens (semantic highlight) for TypeScript. Doesn't apply to JavaScript"
```

dummdidumm (commented):
Very nice! Only thing I gotta check some more is how much it affects performance and how different it looks (in terms of default `true` or `false`). ~~We might also think about checking the TypeScript settings and not enable semantic highlighting if the user turned it off for TypeScript, even if it's turned on for Svelte~~.

About your question: I don't have that much of an opinion on that and I like the current coloring. We might start with what we have here and enhance if there's more feedback.

packages/language-server/src/plugins/PluginHost.ts
dummdidumm: What do you mean by "Svelte semantic tokens"? I don't see another plugin providing tokens besides the TS plugin, so yeah we could change this to `FirstNonNull`.

jasonlyu123: Yeah. I was initially thinking maybe we could have a svelte specific one but can't think of any. I'll just remove it. And if anyone got any idea of svelte specific semantic tokens we can do it in another PR

packages/language-server/src/plugins/typescript/features/SemanticTokensProvider.ts
```diff
@@ -0,0 +1,107 @@
...
+
+    async getSemanticTokens(textDocument: Document, range?: Range): Promise<SemanticTokens> {
+        const { lang, tsDoc } = this.lsAndTsDocResolver.getLSAndTSDoc(textDocument);
+        const fragment = await tsDoc.getFragment();
+        const textSpan = range
+            ? convertToTextSpan(range, fragment)
+            : {
+                  start: 0,
+                  length: fragment.text.length
+              };
```
dummdidumm, line 24: We can save some computation here by ignoring everything that's appended by svelte2tsx which will not be mappable. (Tests would need to be adjusted)

```suggestion
        const textSpan = range
            ? convertToTextSpan(range, fragment)
            : {
                  start: 0,
                  length: tsDoc.parserError
                      ? fragment.text.length
                      : // This is appended by svelte2tsx, there's nothing mappable afterwards
                        fragment.text.lastIndexOf('return { props:') || fragment.text.length
              };
```

dummdidumm:
Had a close look, seems either my eyes don't see it or my theme doesn't colorize as much, either way there's not much difference although I see some semantic tokens applied when inspecting with the token inspector 😄 
I noticed one thing we can do for speedup, see code comment.

jasonlyu123:
Another problem I forgot to mention is that the source map issue of each block is highlighted by the semantic token. So it does look a little bit odd.
圖片

dummdidumm:
I'm okay with this since it's opt-out. We should really tackle this off-by-one-error at some point, if that's even possible.

Monkatraz:
Just to throw my two cents in about this feature: My personal theme makes heavy use of semantic colors, and I really appreciate the effort undertaken to implement them here.

On a more technical note: Many themes for VSCode are quite old - they don't specify semantic token colors even though they could make the syntax highlighting _significantly_ more consistent or meaningful. A good example would be with `const` - using semantic tokens, you can tell at a glance which variables are `const` and which are `let`. This extends to functions being passed as variables, too. 

tldr: on a lot of themes you won't see much of a difference, but on others it's huge

packages/language-server/src/plugins/PluginHost.ts
```diff
@@ -420,11 +438,18 @@ export class PluginHost implements LSProvider, OnWatchFileChanges {
         args: any[],
         mode: ExecuteMode.Collect
     ): Promise<T[]>;
+    private execute<T>(
```
dummdidumm, line 441: We don't need this additional type definition and the `if` inside `ExecuteMode.Collect` anymore, right?

PR #763, opened 2021-01-14
map end of range using SourceMapConsumer.LEAST_UPPER_BOUND

jasonlyu123:
Because of the transformation of svelte2tsx. sometimes the end of the range is the index of a generated code so the source map is mapped to the previous character where the generated code is appended to. So I think mapping the start index of the character would help.

```
Identifier
```
where the range is before `I`(0) and after `r`(10) so we can map before r(9) and offset back the mapped position. 

Current problem:
Remove import code action seems to rely on mapping to the start of the next import. So organize import would be left with multiple empty lines


dummdidumm:
Very interesting! From what I understand after a quick look this is essentially a workaround for the MagicString-limitation we are dealing with. So if MagicString would have support for something like "insert X before Y and if you want the original column for X you get Y, not the character before X", we would not need this right?

jasonlyu123:
Yeah. Or maybe we can also try to extend `MagicString` class and use its internal API to make append and prepend stick with the chunk it belongs.

jasonlyu123:
Find another method, the `SourceMapConsumer.originalPositionFor` method has a [bias option](https://github.com/mozilla/source-map#sourcemapconsumerprototypeoriginalpositionforgeneratedposition) that can be used to determine 
>whether to return the closest element that is smaller than or greater than the one we are searching for,

In this way, the unmappable append would be mapped to the next character when mapping the end of the range. It can also solve each block issue. The downside is it makes some of the overwrite edits mappable, like `__sveltets_each`. But it does not need the extra translation overhead so I tend to go this way rather than the original way The performance difference quite big.
I run it 1000 times for each to test the difference. In a 300 lines component, It's 160-200ms for the offset method and only  0 - 2ms using a different bias option.

dummdidumm:
That does sound very interesting! Only thing I wonder is if we want this behavior for all mappings or just specific ones, and if so, how we determine this.

dummdidumm:
Before merging this I want to explore if altering the code of `MagicString` is possible in such a way that we don't need the workarounds afterwards. Still thanks a lot already for your investigations :thumbsup:

dummdidumm:
I did some investigations about solving this inside `MagicString` and this gives us the behavior of marking the intro as part of the next column:

```diff
// inside generateDecodedMap this.firstChunk.eachNext

if (chunk.intro.length) {
+      mappings.rawSegments.push([mappings.generatedCodeColumn, sourceIndex, loc.line, loc.column]);
       mappings.advance(chunk.intro);
}
```

However, I now doubt that this is the behavior we always want, or more precisely, in some situations we don't want to do `prependRight/appendRight` because it's semantically incorrect and could introduce bugs in the long run. So I now think the solution of correcting the range afterwards is better.

However, I may have found a less introsive alternative which would - I think - fix the same problems your solution does with less additional code by altering the `mapRangeToOriginal` function like this:

```ts

export function mapRangeToOriginal(
    fragment: Pick<DocumentMapper, 'getOriginalPosition'>,
    range: Range
): Range {
    // DON'T use Range.create here! Positions might not be mapped
    // and therefore return negative numbers, which makes Range.create throw.
    // These invalid position need to be handled
    // on a case-by-case basis in the calling functions.
    const originalRange = {
        start: fragment.getOriginalPosition(range.start),
        end: fragment.getOriginalPosition(range.end)
    };
    // <-new code::-> Range may be mapped one character short - reverse that
    if (
        originalRange.start.line === originalRange.end.line &&
        range.start.line === range.end.line &&
        originalRange.end.character - originalRange.start.character ===
            range.end.character - range.start.character - 1
    ) {
        originalRange.end.character += 1;
    }
    return originalRange;
}
```

What do you think? We could start with this change, and if that does not fix all occurences we can always go towards your `bias`-solution later on.

jasonlyu123:
Ummm. That might just be enough to fix our problem. The test all passed. And there are just not many cases we append a new line.

dummdidumm:
Do we need the other changes if we go with that solution? Could you briefly explain what these other code changes fix? Are these things that appeared after you changed the range mapping or are these mostly unrelated?

jasonlyu123:
bf8bc51f891a580778dc059941ae618cbba66943 is unrelated, just another bug that related to `mapRangeToOriginal`.
bd354e6dc639a8f83ae3e8c6618f3254a0924f41 only applies to the bias solution. but maybe we can keep the test.
As for the `isSvelte2tsxGeneratedIdentifer` function, this is also used to fix errors in the bias solution I think we can keep it.

dummdidumm:
Thanks for the info! So this would mean applying the simpler (but not 100% correct) solution means we only need to adjust the tests that are failing and nothing else - good to hear 😄 

I suggest the following:
- I create a separate PR with the range fix
- We close this PR and leave the code as is for future reference if we ever want to add the full solution
- You create a new PR with the additions/fixes from the commits you mentioned
Reason is that our future selves will have an easier time to understand which PR dealt with what issue.

jasonlyu123:
ok!

PR #769, opened 2021-01-18
Make Options.target optional

pushkine:
A target is only required when a component is created manually, the compiler does not use it

Copy pasting the compiled result of the following component in its own `script` throws
`Property 'target' is missing in type '{}' but required in type 'Svelte2TsxComponentConstructorParameters<Partial<{ prop: any }>>'.ts(2345)`
```svelte
<script lang="ts">
	import Component from "./Component.svelte";
</script>

<Component />
```

Another option that is compiler-only is `$$inline`, should we add a different interface for options used by the compiler ? 

I've published [a gist at this url](https://gist.github.com/pushkine/d2ebc3e7131f7a658e5ca7bdf0150574) with accurate typings of svelte's internals for more information

### DIFF
--- a/packages/svelte2tsx/svelte-shims.d.ts
+++ b/packages/svelte2tsx/svelte-shims.d.ts
@@ -50,7 +50,7 @@ interface Svelte2TsxComponentConstructorParameters<Props extends {}> {
     /**
      * An HTMLElement to render to. This option is required.
      */
-    target: Element;
+    target?: Element;
     /**
      * A child of `target` to render the component immediately before.
      */


dummdidumm:
Could you please provide a reproducible? Given your description I don't fully understand what the problem is or how to reproduce it.

pushkine:
My bad, basically the compiler does not use the `target` property
Code_2021-02-09_14-23-11

dummdidumm:
Is that a real use case where this occurs? Using the generated code of Svelte inside a Svelte code block seems ... strange 😄 My fear is (need to check that) that if we change this to optional, that TS will then complain if people want to assign concrete components to  `SvelteComponentTyped` or `SvelteComponentDev` ("constructor signatures different, etc etc").

pushkine:
Talking about constructor signatures errors, I think that might be why I raised the issue in the first place 😄

I wrote a component manually to imperatively access slots and portal stuff around, I naturally just like the compiler don't use target so my types threw a constructor signature error when I imported one of those into a svelte file

I suggested in the OP to make a type for the compiled options and another for the user options, wouldn't that work out ?

Issue #538, opened 2020-09-14
Autocompletion/Diagnostics broken when accessing property on object within moustache tag (`bla.`)


lake2:
code：
```js
{#if user.loggedIn}<button on:click={toggle}> Log out </button>{/if}

<script>
    let user = { loggedIn: false };

    function toggle() {
        user.loggedIn = !user.loggedIn;
    }
</script>
```
when type `user.`，auto complete is broken：

![image](https://user-images.githubusercontent.com/6293752/93059583-7fa76380-f6a3-11ea-94b8-be4d61512143.png)

when type `user.a` this is no error:
```js
// expect an error here 
{#if user.a}<button on:click={toggle}> Log out </button>{/if}

<script>
    let user = { loggedIn: false };

    function toggle() {
        user.loggedIn = !user.loggedIn;
    }
</script>

```

dummdidumm:
Blocked by https://github.com/sveltejs/svelte/issues/4818

jasonlyu123:
Maybe we can replace something like `a.}` and `a?.}` to `{a }`  `{a  }` before we hand it over to the svelte compiler. That won't fix all the issue but can workaround simple one. But is it worth the hassle and will it introduce some unwanted behavior?

dummdidumm:
I think that's a good idea to explore. I guess if it sometimes will result in strange behavior (maybe, don't know) it will still be better than the current behavior.

dummdidumm:
The mentioned use case of "`.` before `}` is fixed. There surely are other cases (for example #776) where intellisense will not work properly. For these, please open a separate issue.

PR #786, opened 2021-01-28
(svelte2tsx)blank possibly operator or property access near end of mustache before svelte compile

jasonlyu123:
Run a simple scan before sending the code to the svelte compiler. Because it's mostly a syntax error to have an operator right before the end of the mustache tag. This would allow the errored syntax to be emitted untouched. Allowing better dx. But this workaround should only apply to a syntax error. So we have to check if any of it is a valid syntax.
for example:
```sv
{a?.}
{a. }
{a ? }
{a &&}
{a+}
```
#538


packages/svelte2tsx/src/utils/htmlxparser.ts
```diff
@@ -113,3 +114,57 @@ export function parseHtmlx(htmlx: string): Node {
     }
     return svelteHtmlxAst;
 }
+
+const possibleOperatorOrPropertyAccess = [
```
dummdidumm, line 118: For performance it might be better to use a `Set` here.

dummdidumm:
This looks very nice. I'm trying to think of situations where this would break otherwise correct output.
If we want to be even more safe, we could try to do two passes: One without that scan, if that works -> go on; if it doesn't -> try again with the scan. What do you think?

jasonlyu123:
Kind of wanted to just go for it XD.

dummdidumm:
I think we need this to be configurable. `svelte-check` uses `svelte2tsx`, too, and in this case we want these errors to appear.

jasonlyu123:
I agree. What about `emitOnTemplateError` in the option parameter of `svelte2tsx`

dummdidumm:
Sounds good! It should be `false` by default. We also somehow need to enhance the tests so that you can invoke `svelte2tsx` with arbitrary options and test whether it fails or not. Maybe check if there's `test.js` in the test case folder and if yes invoke that method within instead.

dummdidumm (commented):
Nice work! One more thing, then I we are good to merge I think: You put the `test.js` in the same folder as the tests for a conversion from svelte to tsx/jsx. The test utils are written such that if there's a `test.js`, the other tests will not be done anymore. So I think you need to split out the the `test.js` tests into separate folders. This would also make it more clear that these are two different tests.

jasonlyu123:
Do you mean a folder in the test folder dedicated to the custom tests?

dummdidumm:
I meant instead of 
```
html2jsx/edit-moustache
  | test.js
  | input.svelte
  | expected.jsx
```
Two sparate test cases
```
html2jsx/edit-moustache
  | input.svelte
  | expected.jsx
html2jsx/moustache-error
  | test.js
```
(same for the svelte2tsx test)

jasonlyu123:
This means we'll make the default test use the new option. And the custom one uses the default option to test it fail?

dummdidumm:
Oh I noticed just now that `test.js` actually contains two test cases, I thought it was only one and that the svelte->jsx/tsx test would be swallowed by that. Nevermind then 😄 I'll merge this then if you don't have any objections.

jasonlyu123:
Ok! no other to adjust.

Issue #726, opened 2020-12-25
Implement 'Synced Regions' for Svelte


kwangure:
HTML for VS Code [recently introduced Synced Regions for HTML tags](https://github.com/microsoft/vscode/issues/88424). In summary, if you edit an opening tag, the same edits appear in the closing tag and vice versa.

This API is available for other extensions [as of two weeks ago](https://github.com/microsoft/vscode/issues/109923#issuecomment-741866857).

This is an issue to track the implementation of Synced Regions for: (Tip: `|` is the cursor) 
- Svelte markup (e.g `<div|></div|>`)
- Svelte control flow (e.g `{#each| n}{/each|}`).
- ** Svelte classes & ids (e.g `<div class="apple|"></div> <style> .apple| {color: red}</style>` )
- ** JavaScript 
	- `<script> let world| = "" </script><div>Hello {world|}</div>`.
	- `<script> import Component| from "./component"</script><Component|></Component|>`

** Syncing JavaScript and CSS classes and ids is iffy since I can imagine wanting to edit a variable/class without changing others in many situations. I haven't experimented with how flexible the `LinkedEditingRangeProvider` API is, but if possible, we could require a key-binding to get into synced mode for these contentious regions. Requiring a key-binding will likely hurt feature discovery + usage of syncing for these regions, so discussion of other possible implementations is very welcome.

I'd suggest four separate settings options for HTML, flow control, CSS and JS. Where HTML & flow control are on by default, but CSS & JS are off by default.

jasonlyu123:
The markup part would be easy to implement and it probably is what it's designed for. The svelte control flow is viable but probably only possible when there's no syntax error in the file. 
The proposal of  CSS class name, id, and javascript identifier feel more like a rename symbol to me. We already have it for javascript but not CSS. And about the partly rename you can always use [multiple cursors](https://code.visualstudio.com/docs/editor/codebasics#_multiple-selections-multicursor) to achieve it.

jasonlyu123:
I tried to implement the svelte control flow but the result using svelte AST is not very useful. It's too brittle. During linked editing, the code is very likely to be invalid. So the svelte compiler keeps complaining. We might need to implement a dedicated parser for it to be useful. So this probably won't happen soon.

dummdidumm:
I also think that using the Svelte AST will be too brittle. Writing a custom parser sound like too much work for such a feature. Maybe the html parser we use in other places could be used for that.

jasonlyu123:
[...]

PR #801, opened 2021-02-05
(feat) linked editing for html

jasonlyu123:
#726 
linked editing for element and component tag. Needs `editor.linkedEditing` to work properly. Without the setting, It does highlight the linked editing ranges with `Ctrl+Shift+F2`. but it doesn't seem to enable linked editing.

This works unreliably and I can't find out why. The result on our end seems correct. But it just stops working sometimes. And it will work again after restarting the language-server, In the same position. Need more research on the vscode side to see if there's anything we missed or it's a bug on their side.


nathanblair:
Glad I'm not crazy and thought html tags inside svelte files should work with linked editing.

I haven't yet had an experience where linked editing does work but I've only been using VS Code with the linked editing enabled for about 20 minutes now.

I'll keep my findings posted here if my experience changes. Thanks for looking into this!.

jasonlyu123:
I am guessing it's because the request takes too long. The problem usually happens when I switch files and immediately started linked editing. There is usually a lot of processing when opening the file so the request might take longer.
I can see a similar problem when setting the file association to HTML. But didn't see much complaining in the vscode repo. Maybe we can just add it and see the feedback.

PR #802, opened 2021-02-05
VSCode: Extra goodies for Svelte TM grammar

Monkatraz:
* New highlighting for `$:` reactive statements in scripts
* New highlighting of the `$` symbol in `$var` store accessors in scripts
* New highlighting for Svelte built-ins, currently `$$props`, `$$restProps`, and `$$slots`
* Updated highlighting for element directives (more specific `bind:` and `class:` highlighting)
* Bugfix / redo of how 'unit' (e.g. `4px`) unquoted strings are highlighted

Some notes:
* The scope chosen for the `$` symbol in store accessors is the same scope used by PHP and a few other languages, and is thus fairly well supported. This scope is also used for `$$` in Svelte built-in constants.
* The scope chosen for `$` in reactive statements is not punctuation, but rather as if it was a full keyword.
* The reactive statement highlighting should be smart enough to avoid conflicting with types, object literals and JS labels.
* The scope for Svelte's built-ins is a standard scope intended for language built-ins, so it should be nicely highlighted for most themes.
* The grammar now has an easy way of highlighting a directive differently depending on whether or not the directive is being assigned to something. Currently, this is used for `class:` and `bind:`.

Some motivation for the new highlighting in scripts:
My main goal with the additions was to cleanup how the few syntax features Svelte adds to JS appear to users. Hopefully, these features now appear as if they're naturally integrated into the language and their appearance won't vary wildly between themes or stand out when compared to similar features in other languages, with a good example being PHP variables.

Some motivation for the element directive change:
I just got irrationally bugged that `class:foo={bar}` was highlighted like `class:bar`, as I felt that `foo` should be highlighted like a CSS class and not as if it is an existing variable being used as a shorthand. While I was at it, I made sure that `bind:` had the same behavior (except with an entity type instead of a CSS class).

Pretty pictures:
My Theme > Dark+ > Svelte Light
image
image
image
image
image
image


dummdidumm (approved):
Thanks, these are nice additions 👍

Issue #806, opened 2021-02-07
Greater than operator not working on if mixin on pug


clarenceeda:
**Describe the bug**
Greater than operator not working on pug  using +if()

**To Reproduce**
Followed [Everything](https://github.com/sveltejs/svelte-preprocess/blob/main/docs/getting-started.md)
`
$ npx degit sveltejs/template my-svelte-app
$ cd my-svelte-app
$ npm install -D svelte-preprocess
$ npm install -D pug
`
![svelte preproess bug](https://user-images.githubusercontent.com/42695235/107137985-b1c59600-68ce-11eb-9549-8570b26d39df.png)

 - OS: Windows 10 x64
 - vscode 1.53.0

jasonlyu123:
We currently don't fully support pug. Maybe in the future, we could use the pug parser to transform the template for javascript features. But we might need some help, as non of the core maintainers of this repo uses pug.

But in this case, I think we can work around it by removing the '}' or '>' in the template. it's all recognized as text in our transformation. And these two characters is allowed in the text of svelte.

dummdidumm:
This specific error should be fixed now. There may very well be other type/semantic errors though because we don't fully support pug at the moment.

PR #807, opened 2021-02-09
(svelte2tsx) Jsx not allowed char

jasonlyu123:
`}` and `>`  are not allowed in JSX but allowed in svelte. Remove it in `svelte2tsx`. Also, remove `>` from svetle2tsx blank error preprocess. As it might be a close tag.


dummdidumm:
This is for #806 right ?

jasonlyu123:
Yeah. But it doesn't fix the root pug mixin problem. Also, some markdown syntax used in MdsveX has a similar problem.

dummdidumm:
Yeah to fix the pug problem we'd need to actually have a dedicated pug parser, or run pug before handing it off to `svelte2tsx`... things for later.

Issue #808, opened 2021-02-09
VS Code freezes when the last character of a custom event handler is deleted


ProxyOxide:
**Describe the bug**

I noticed some strange behavior when using a component with custom events. If I have `<Foo on:bar={e => alert("hey")} />` and I start deleting from the end of the event handler, the VS code window hangs as soon as the body of the event handler becomes empty `<Foo on:bar={} />`

**To Reproduce**

Foo.svelte:
```
<script lang="ts">
    import { createEventDispatcher } from "svelte";
    const dispatch = createEventDispatcher();

    function func() {
        dispatch("bar", {abc: 123})
    }
</script>
```

App.svelte:
```
<script lang="ts">
	import Foo from "./Foo.svelte";
</script>

<Foo on:bar={e => alert("hey")} />
```

Place the cursor at the end of the event handler right before the closing `}`. Backspace until the event handler body becomes `{}`. 

**Expected behavior**

VSCode doesn't crash and I can continue editing the app component. 

**System (please complete the following information):**
 - OS: Windows 10 Home 2004 19041.746
 - IDE: VSCode 1.53.0 Remote on WSL2 (Ubuntu 20.04)
 - Plugin/Package: Svelte for VSCode 104.4.0

**Additional context**

Minimal project to reproduce the issue
[SvelteExample.tar.gz](https://github.com/sveltejs/language-tools/files/5948911/SvelteExample.tar.gz)

jasonlyu123:
This is weird. I can see a high CPU usage on the main process of vscode, not extension host or language server. From my understanding, the extension can't block the UI thread. 

After some more digging, I found disabling `svelte.plugin.typescript.semanticTokens.enable` would prevent the problem. But I don't know why. It freezes before asking the language server to update semantic tokens.

jasonlyu123:
Found the reason. There is a negative length in the semantic tokens result. We need to filter that out.

PR #811, opened 2021-02-09
fix negative length and wrong order semantic token of event handler

jasonlyu123:
In svelte2tsx transformation, component event handler code is moved, Thus the semantic token would have the wrong order with other attributes and a negative length($on). Filter out the negative one and then sort the token. Seems like the issue with #808 happened when vscode tried to update the semantic token with the negative token length.

### DIFF
--- a/packages/language-server/test/plugins/typescript/features/SemanticTokensProvider.test.ts
+++ b/packages/language-server/test/plugins/typescript/features/SemanticTokensProvider.test.ts
@@ -62,42 +62,42 @@ describe('SemanticTokensProvider', () => {
             modifiers: number[];
         }> = [
             {
-                line: 1,
+                line: 2,
                 character: 14,
                 length: 'TextContent'.length,
                 type: TokenType.interface,
                 modifiers: [TokenModifier.declaration]
             },
             {
-                line: 2,
+                line: 3,
                 character: 8,
                 length: 'text'.length,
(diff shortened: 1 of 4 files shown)


packages/language-server/test/plugins/typescript/features/SemanticTokensProvider.test.ts
```diff
@@ -187,7 +187,7 @@ describe('SemanticTokensProvider', () => {
 
         let index = 0;
         while (index < tokens.length) {
-            result.push(tokens.splice(index, (index += 5)));
```
jasonlyu123, line 190: Wrong array method when implementing this. meant to group the token array by 5 so the assertion error can be spotted more easily.

ProxyOxide:
I ran this branch against my project and am no longer seeing the crash from #808. 

The only thing that's slightly strange is the lack of highlighting in my components. That could just be from running inside the extension dev host, though.

jasonlyu123:
Did you build the grammar?
```bash
cd packages/svelte-vscode
yarn build:grammar
```

Seems like this isn't in the setup doc

ProxyOxide:
> Did you build the grammar?
> 
> ```shell
> cd packages/svelte-vscode
> yarn build:grammar
> ```
> 
> Seems like this isn't in the setup doc

That worked, thanks!

Issue #812, opened 2021-02-10
Reactive Statement Brackets breaks highlighting


SaintPepsi:
**Describe the bug**
Reactive statement breaks syntax highlighting when there are more than 2 and the brackets are immediately after `$:` 
Also happens with every plugin disabled except for the "Svelte for VSCode"

**To Reproduce**
Create an empty svelte file.
make sure syntax highlighting is set to svelte
Add this:
```svelte
<script>
  let test1 = 0;
  let test2 = 0;
  let test3 = 0;

  $: {
    console.log('test3 changed', test1);
    console.log('test1');
  }

  $: {
    console.log('test3 changed', test2);
    console.log('test2');
  }

  $: {
    console.log('test3 changed', test3);
    console.log('test3');
  }

  let someArray = [0, 1, 2, 3, 4];
  for (let i = 0; i < someArray.length; i++) {
    const element = someArray[i];
    console.log('element', element);
  }
</script>

<h1>div highlight gone too</h1>
<div class="div">some div..</div>

<style>
  h1 {
    text-align: center;
    margin-top: 10px;
  }

  div {
    width: 100px;
    color: red;
  }
</style>
```

**Expected behavior**
Expected Correct colours to show up for content inside script and style tag.

**Screenshots**
How it looks with more than 2 reactive statements:
![image](https://user-images.githubusercontent.com/16056759/107455859-51b83500-6b04-11eb-86d0-845760232492.png)

How it should look:
![image](https://user-images.githubusercontent.com/16056759/107456131-83310080-6b04-11eb-8db6-1e8b0e73a8b6.png)

Difference: 
![image](https://user-images.githubusercontent.com/16056759/107456220-a52a8300-6b04-11eb-8fbd-53a6cb0987b5.png)

**System (please complete the following information):**
 - OS: Windows
 - IDE: VScode
 - Plugin/Package: Svelte for VSCode

**Additional context**
On file save my prettier plugin auto complies:
```svelte
$:
{
}
```
into:
```svelte
$: {
}
```

PR #813, opened 2021-02-10
(fix) syntax highlighting

dummdidumm:
Remove `$:` highlighting as it affects the TS grammar. If you want this highlighted differently, provide a special color for `entity.name.label.ts` / `entity.name.label.js`.
Also made the `$$` match less intrusive to avoid conflict with the TS grammar and added a short README on TextMate
#812

@Monkatraz @jasonlyu123 FYI. To me this means we should not introduce more injected stuff.


SaintPepsi:
Thanks for fixing that fast.

I had to figure it out but it works like this in VScode for anyone that doesn't know:
add to `settings.json`:

```
"editor.tokenColorCustomizations": {
        "textMateRules": [
            {
                "scope": "entity.name.label.js",
                "settings": {
                    "foreground": "#c678dd" // Your color here
                }
            }
        ]
    },
```

Issue #785, opened 2021-01-28
Error: Expected 2 arguments, but got 1. (ts) for <div transition:fade >


bfanger:
Since svelte-check 1.1.29 the transition shorthand `<div transition:fade>` incorrectly reports the 
`Error: Expected 2 arguments, but got 1.` error.

When using the shorthand notation, svelte calls the fade (or any other function) with 2 arguments, the element and an empty object. [REPL](https://svelte.dev/repl/3dfed2eaa77a4ae9b5be415a22420d48?version=3.32.0)

**To Reproduce**

```html
<script lang="ts">
  import { fade } from "svelte/transition";

  let visible = true;

  function toggle() {
    visible = !visible;
  }
</script>

<button on:click={toggle}>Toggle</button>
{#if visible}
  <div transition:fade>Fades in and out</div>
{/if}
```

Demo: https://github.com/bfanger/svelte-check-transition-bug

The failed run: https://github.com/bfanger/svelte-check-transition-bug/runs/1783108223?check_suite_focus=true

dummdidumm:
This is a result of #767 , transitions/animations/use directives are now transpiled differently. Looking at Svelte's [source code](https://github.com/sveltejs/svelte/blob/master/src/runtime/transition/index.ts), the error is actually correct, so Svelte's typings need to be adjusted. Right now they don't mark the second parameter as optional.

jackwestmoretab:
I have also experienced this issue -- thanks @dummdidumm for looking into it!

bfanger:
I think the svelte patch fixes the error when using fade (and the other build ins)  but svelte-check is still incorrect

Using the transition:fade shorthand behaves as if you'd written transition:fade={ {} } and fade is passed 2 params, not 1.

dummdidumm:
What do you mean by that?
`transition:fade` currently is translated (by svelte2tsx) to `__sveltets_ensureTransition(__sveltets_ElementNode)` for intellisense purposes.  `transition:fade={ {} }` would be `__sveltets_ensureTransition(__sveltets_ElementNode, {})` which is correct.

bfanger:
Both should be translated to __sveltets_ensureTransition(fade(__sveltets_ElementNode, {})) 

Check the repl https://svelte.dev/repl/3dfed2eaa77a4ae9b5be415a22420d48?version=3.32.0

dummdidumm:
[...]

PR #816, opened 2021-02-14
fix: shorthand transition and animation params

bfanger:
Real fix for #785

I've updated the [REPL](https://svelte.dev/repl/3dfed2eaa77a4ae9b5be415a22420d48?version=3.32.3) to make it extra clear what is happening.

Directive | Nr. args | Empty object
-- | -- | --
use | 1 | false
use={{}} | 2 | true
animate | 3 | true
animate={} | 3 | true
animate={{duration}} | 3 | false
transition | 2 | true
transition={{}} | 2 | true
transition={{duration}} | 2 | false

This demonstrates that unlike the `use` shorthand, the transition and animate shorthand pass an empty object as parameter.
 `<div transition:fade>` is identical to `<div transition:fade={{}}>` as far as the svelte compiler is concerned.

This allows custom transitions like [svelte-transitions-fade-scale](https://github.com/flekschas/svelte-transitions-fade-scale/blob/master/src/index.js):
```js
function fadeScale(node, { delay = 0, duration = 200, easing = x => x, baseScale = 0 })
```
to work as shorthand `<div transition:fadeScale>` without causing `Cannot read property 'duration' of undefined` errors.

In svelte 3.32.0 the params got an empty object as default value ` = {}`, this change hides the fact that svelte2tsx  created the wrong type for the built-in transitions, silencing the "Expected 2 arguments, but got 1." error reported by svelte-check
(This is still a nice change, it makes reusing the built-in transitions with defaults easier)

### DIFF
--- a/packages/svelte2tsx/src/htmlxtojsx/nodes/animation-directive.ts
+++ b/packages/svelte2tsx/src/htmlxtojsx/nodes/animation-directive.ts
@@ -20,11 +20,7 @@ export function handleAnimateDirective(
     const nodeType = `__sveltets_mapElementTag('${parent.name}')`;
 
     if (!attr.expression) {
-        if (animationsThatNeedParam.has(attr.name)) {
-            str.appendLeft(attr.end, `(${nodeType},__sveltets_AnimationMove,{}))}`);
-        } else {
-            str.appendLeft(attr.end, `(${nodeType},__sveltets_AnimationMove))}`);
-        }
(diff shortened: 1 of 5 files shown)


dummdidumm:
I didn't initially always add the empty object because my fear is that people would create custom transitions/animations which don't expect an object as second parameter. Then again, since the compiler adds this in case nothing is passed in, I guess this change makes sense.

Issue #580, opened 2020-09-27
TypeScript Plugin


dummdidumm:
The language-server has quite a few features by now. What's missing for the complete developer experience and intellisense though is crossing the boundaries of Svelte and JS/TS files. This can be achieved through a [TypeScript plugin](https://github.com/microsoft/TypeScript/wiki/Writing-a-Language-Service-Plugin). The following features would be cool to have:

- Rename variables in Svelte files, too, when renaming starts in a TS/JS file
- Go to Svelte definition/file from a TS/JS file
- Find references
- Better diagnostics (#550)

dummdidumm:
Dumping some thoughts:
- A TS plugin needs to be synchronous, meaning all methods like `getDiagnostics` need to be synchronous. This is not possible to do right now since we use `source-map` which asynchronously generates a `SourceMapConsumer`. So either we switch the library or we do some kind of "use latest processed if available"-logic. What we definitely cannot do is reuse the language server connection and do `requestX, await response`.
- Because we cannot use the language server like the LSP protocol intends. Maybe it's possible to make the TypeScriptPlugin from the language server available for direct use while making sure it uses the same instance as the LS.

pushkine:
Hey I worked on a proof of concept today 

https://user-images.githubusercontent.com/30108880/107896725-6e56d180-6f37-11eb-8ac9-fff759572208.mp4

https://github.com/pushkine/ts-plugin-svelte/

Beside obvious bugs where everything gets stale this doesn't yet work with Components, might be a conflict with `declare global module "*.svelte"` or some tweaking to do with `svelte2tsx`

Looks good to me though, what do you think ?

dummdidumm:
This is a nice POC!

I though about this topic a little more and here is my roadmap on how we could get to a plugin:
1. Create performance-tests (#676)
2. Replace async `source-map` with sync `sourcemap-codec` + own litte mapping function similar to how I did it in [this ESLint plugin PR](https://github.com/sveltejs/eslint-plugin-svelte3/pull/83). Optionally reuse that logic in that PR to simplify+fix our mapping logic (could solve #666)
3. Test if the changes in 2 did not hurt the performance using the tests implemented in 1. If there are no big performance hits, continue.
4. Make all TypeScriptPlugin function synchronous - should be easily doable then because the only reason it was asynchronous was because of `source-map`
[...]

PR #820, opened 2021-02-15
WIP synchronous mapping

dummdidumm:
This replaces asynchronous `source-map` with a synchronous mapping implementation using `sourcemap-codec`. According to the performance test there is no noticeable difference at file lenghts like these - `source-map` is better at really large files (thousands of lines), which is not the case for 99,9% of all Svelte files the user will edit with this. A further (according to tests not noticeable) performance improvment would be to return the decoded form of the source map from svelte2tsx (it used sourcemap-coded, too).

The mapping implementation could also serve as a reference implementation for synchronous source mapping for TS plugin (#580)

### DIFF
--- a/packages/language-server/package.json
+++ b/packages/language-server/package.json
@@ -53,6 +53,7 @@
         "prettier": "2.2.1",
         "prettier-plugin-svelte": "~2.1.0",
         "source-map": "^0.7.3",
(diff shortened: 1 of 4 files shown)


dummdidumm:
Mapping was replaced a while ago and is now synchronous

PR #830, opened 2021-02-20
add trace server setting

jasonlyu123:
https://code.visualstudio.com/api/language-extensions/language-server-extension-guide#logging-support-for-language-server


dummdidumm:
This works without a dedicated setting by modifying the settings JSON file, but I think it's a good idea to make it visible through the settings UI 👍

Issue #834, opened 2021-02-21
Multiple slot prop declarations cause errors


brunnerh:
**Describe the bug**
Declaring multiple slot props causes errors even though the code is valid.

**To Reproduce**

`Component.svelte`
```svelte
<script type="text/typescript">
	export let items: string[] = [];
</script>

{#each items as value, index}
	<slot name="item" {value} {index}/>
{/each}
```

`App.svelte`
```svelte
<script type="text/typescript">
	import Component from './Component.svelte';
</script>

<Component items={['hello', 'world']}>
	<div slot="item" let:index let:value> <!-- Errors here -->
		{index}: {value}
	</div>
</Component>
```

Error on the `l` of `let:value`:

> Property destructuring pattern expected. ts(1180)

Error on the `e` of `let:value` (possibly just a result of the first):

> 'value' is declared but its value is never read. ts(6133)
> '}' expected. ts(1005)

When not using TypeScript there seem to be equivalent JS errors.

**Expected behavior**
No errors.

**Screenshots**

![image](https://user-images.githubusercontent.com/834235/108622836-bdbb6700-743b-11eb-92e3-cda641022a89.png)

**System (please complete the following information):**
 - OS: Windows 10 64bit
 - IDE: VSCode
 - Plugin/Package: Svelte for VSCode

PR #835, opened 2021-02-21
(fix) correct multiple slot let transformation

dummdidumm:
Parantheses are important..
#834


brunnerh:
@dummdidumm Thank you for all your work on this!

Issue #838, opened 2021-02-21
svelte-check --watch doesn't account for changes in helpers imported from *.ts files


non25:
Hi.
First of all, thanks for the tools.
I like workflow with `svelte-check --watch`.
No annoying stuff in the editor while I'm typing, but helpful information when I'm done and saved.

**Issue**

When I change a function imported in svelte component from a `*.ts` file, and then save the component to trigger diagnostics it will act like `*.ts` file has never changed, thus making watch mode deceiving and not really helpful.

Here's repro branch: https://github.com/non25/svelte-template-webpack/tree/svelte-check-watch-repro

To reproduce on your machine:

1. Download it with degit: `npx degit non25/svelte-template-webpack#svelte-check-watch-repro repro`
2. Install stuff like you always do
3. Launch `svelte-check` watch mode: `npm run validate` or `yarn validate`
4. Edit `src/helper.ts`, change argument type from `string` to a `number`
5. Add spaces or make other edits to `src/App.svelte`
6. See that `svelte-check` emits no errors, despite code is broken now
7. Edit `src/App.svelte`, change `helper(name)` to `helper(1)`
8. See that `svelte-check` outputs: `Error: Argument of type 'number' is not assignable to parameter of type 'string'`

PR #839, opened 2021-02-22
(fix) svelte-check: watch (d.)ts/js files

dummdidumm:
#838


non25:
Cool, I tested it and looks like I don't even need to make changes to dependend svelte file to know that I made a little uwu. :grin: 
Will edit here if I notice something unusual.
Thank you.

PR #842, opened 2021-02-24
Typescript Language Plugin

pushkine:
This TSPlugin injects arbitrary languages to the list of files the real TS Language Server supports, when a file of that language is requested it simply transforms its cached contents into another language that TS natively understands (TSX in our case), and maps LanguageService requests back to the source when returning results.

This implementation edits the cached files content back and forth between languages on the fly, svelte essentially becomes tsx in the eyes of the ts language server. Benefits include refactoring, rename, codelens, references, imports, autocomplete and so on...

(WIP) Need to change how things are mapped in svelte2tsx, will submit a PR once #864 is resolved

## Todo
- [ ] Write a short breakdown of what's going on, list mutated parts of the internal API
- [ ] Write tests

### DIFF
--- a/packages/svelte-vscode/package.json
+++ b/packages/svelte-vscode/package.json
@@ -495,9 +495,9 @@
         ],
         "typescriptServerPlugins": [
             {
-                "name": "ts-plugin-svelte",
+                "name": "typescript-svelte-plugin",
                 "enableForWorkspaceTypeScriptVersions": true,
-                "path": "../ts-plugin-svelte",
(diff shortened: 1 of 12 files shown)


dummdidumm:
I created #978 which takes a different approach which is not that deeply integrated with the TS plugin. This has the advantage of not patching that many private APIs and having more control of the process. The only drawback is that Svelte file changes need to be saved to disk in order to get noticed, which is okay to me. Later on we could investigate a more deep integration if needed.

PR #844, opened 2021-02-25
completion for ids/classes in the template

jasonlyu123:
completion for ids/classes in the template that are present in CSS
圖片


dummdidumm (approved):
Very nice! One thing that could help the visibility of this feature is to trigger the completion on space or `"`, but we would have to make sure that this does not result in too much autocompletion noise and also make sure that it will not negatively affect other completion participants (they'd need to filter that trigger character out). --> we can do that in another PR if we have the time/want it.

PR #845, opened 2021-02-25
Convert svelte2tsx tests to typescript and normalize tests for expected errors in samples

pushkine:
* Converted every non-sample test file to typescript
* Added `sucrase` to devDependencies to run typescript files in tests
* Inlined `mocha.opts`
* Refactored `"svelte2tsx/test/helpers.ts".test_samples` for clarity and added support for `expected.error.json` files in samples 

if a sample includes `expected.error.json` it is expected from the sample to throw that error on the first pass, then to succeed on a second pass with `emitOnTemplateError : true`

Errors can be auto-generated by creating an empty `expected.error.json` file


dummdidumm:
I'm not sure we need the conversion to TypeScript, especially if it means adding sucrase which adds many Babel dependencies. I think we can achieve the same with ts-node which we already have as a dev dependency. Here's how the language server tests do it: https://github.com/sveltejs/language-tools/blob/master/packages/language-server/package.json#L8 

I like the refactoring though, and I also like the idea with `expected.error.json`, although I would rename it to `expected.noEmitOnTemplateError.json`.

Since this is essentially three things in one (refactor, convert to ts, the emitOnTemplatError thing) I suggest you instead create three different PRs.
- One for emitOnTemplateError
- One for the refactoring
- If the one for the refactoring is merged, you may look into converting to TS with the contraint of using ts-node

pushkine:
`@sveltejs/svelte` already uses sucrase to run tests 
https://github.com/sveltejs/svelte/blob/master/.mocharc.js

> adding sucrase [...] adds many Babel dependencies

It doesn't http://npm.anvaka.com/#/view/2d/sucrase

dummdidumm:
> `@sveltejs/svelte` already uses sucrase to run tests
> https://github.com/sveltejs/svelte/blob/master/.mocharc.js
> > adding sucrase [...] adds many Babel dependencies
> 
> It doesn't http://npm.anvaka.com/#/view/2d/sucrase

Fair point, but still I don't want two different transpilers in the devDeps if it isn't avoidable.

pushkine:
> still I don't want two different transpilers in the devDeps if it isn't avoidable.

I opened #865 for the bare minimum changes to turn tests to typescript, I need your input to make it work with `ts-node`

PR #852, opened 2021-03-03
support missing funciton declaration quick fix

jasonlyu123:
added in typescript 4.2

shared source mapping logic from extract to function refactor


dummdidumm (approved):
Nice 👍

PR #865, opened 2021-03-08
Convert svelte2tsx tests to typescript

pushkine:
Part of #845

Changed test script to require ts-node, inlined mocha opts
Renamed extensions from `.js` to `.ts`


dummdidumm (approved):
👍

PR #866, opened 2021-03-09
Add "conservative" CSS syntax highlighting for style attributes

Monkatraz:
This adds a form of syntax highlighting inside of CSS style attributes as requested by https://github.com/sveltejs/language-tools/issues/381. This also adds a slight "fix" to attribute value number strings, which allows them to be quoted (pictures below).

Syntax highlighting is done "conservatively", as in the actual `source.css` grammar **is not included.** Instead, I wrote an "adequate" grammar, which should look fairly similar to how CSS normally looks in themes and is plenty good enough for Svelte's purposes. The primary reason I did it this way was to avoid any 'break-out' bugs. This addition, therefore, should actually avoid those bugs - however, because we don't really have a good test suite yet, I haven't been able to test this thoroughly.

What is actually highlighted in the grammar is very minimal - selectors, media blocks, etc. are not handled because, afaik, they won't ever appear in a style string. Additionally, the highlighting won't actually "check" for valid expressions (forgetting a semi-colon) because it doesn't parse the CSS using begin-end blocks. It uses `match` pretty much exclusively except for functions and strings.

Due to the lack of a test suite, I'd recommend caution when merging this - I can't see how my additions can go _too_ wrong, but I don't know for sure. Textmate grammar definitions are cursed.

One issue I do know of is that any themes that specifically handle `source.css` may highlight interpolated Svelte expressions slightly wrong - e.g. in my theme, CSS variables are red, so Svelte variables become red inside the interpolated expression. This is a tradeoff, as I'd have to make things either very messy or complex to fix this. This isn't too common in themes.

Some screenshots:

My theme
image

Dark+
image

Light+
image

Finally, here is that "attribute value number string" quote change thingy:
image

As long as it's _just_ a number/unit being quoted, it'll be highlighted as a number/unit. This was done so that you can always quote attribute values (say for stylistic or safety reasons) and still get the benefit of the number/unit highlighting.

### DIFF
--- a/packages/svelte-check/rollup.config.js
+++ b/packages/svelte-check/rollup.config.js
@@ -63,6 +63,7 @@ export default [
             // we don't want to bundle and instead require them as dependencies
             'typescript',
             'svelte',
+            'svelte/compiler',
             'svelte-preprocess',
             'import-fresh', // because of https://github.com/sindresorhus/import-fresh/issues/18
             'source-map'
--- a/packages/svelte-vscode/syntaxes/svelte.tmLanguage.src.yaml
+++ b/packages/svelte-vscode/syntaxes/svelte.tmLanguage.src.yaml
@@ -411,8 +411,16 @@ repository:
   # Matches attribute keyvalues. (and boolean attributes as well)
   # e.g. `class="my-class"`
   attributes-keyvalue:
-    begin: ([_$[:alpha:]][_\-$[:alnum:]]*)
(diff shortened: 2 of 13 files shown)


dummdidumm:
This looks nice 👍 still I want to get some syntax highlighting tests set up before merging this.

Monkatraz:
Remaking and splitting these changes up.

PR #869, opened 2021-03-09
Normalize tests for expected errors in samples

pushkine:
On top of #868
Comparison: https://github.com/pushkine/language-tools/compare/share-sample-logic...pushkine:normalize-test-errors


packages/svelte2tsx/test/htmlx2jsx/index.ts
```diff
@@ -1,6 +1,12 @@
 import { htmlx2jsx } from '../build/htmlxtojsx';
 import { test_samples } from '../helpers';
 
-describe('htmlx2jsx', () => {
-    test_samples(__dirname, htmlx2jsx, 'jsx');
+describe('htmlx2jsx', function () {
```
dummdidumm, line 4: why the change to `function` (here and below)?

packages/svelte2tsx/test/svelte2tsx/index.ts
```diff
@@ -1,14 +1,15 @@
 import svelte2tsx from '../build/index';
 import { test_samples } from '../helpers';
 
-describe('svelte2tsx', () => {
+describe('svelte2tsx', function () {
```
dummdidumm, line 4: Why the change to `function` (here and below)?

dummdidumm (commented):
Looks good to me. Could you fix the merge conflicts?

packages/svelte2tsx/test/htmlx2jsx/index.ts
pushkine: I took the habit to avoid inline functions when possible (worse performance), I should avoid doing that in PRs

dummdidumm (approved):
Thanks!

PR #872, opened 2021-03-10
(perf) improve svelte-check performance

jasonlyu123:
Because of the `setTimeout` in the `debounceSameArg`. The document update in `LSAndTSDocResolver` is actually done after all the diagnostic is done. Because of that, the typescript language service has to rebuild its type checker multiple times due to a new document being loaded in.

Remove the denounce make svelte-check run a lot faster on my machine. It went down from 1:30 to 15s for one of my projects.


dummdidumm:
That is a good observation! Maybe we could even generalize this in such a way that:
- if file is new (no prev version) or editor=false fire the change immediately
- else do the debounce

That could maybe improve startup time on opening new files in the editor.

dummdidumm:
👍

PR #874, opened 2021-03-11
(fix) Don't step out of if block on else if

Liamolucko:
`else if`s are represented with an if block node inside them, which is correctly handled when entering, but when leaving it would previously incorrectly step out of the if block leading to an internal error.


dummdidumm:
Good catch, thank you!

PR #887, opened 2021-03-18
Insert comma between bubbled events and custom events

### LINKED ISSUE #886: svelte2tsx outputs wrong invalid TSX (return value -> events)
**Describe the bug**
`svelte2tsx` outputs invalid tsx given the example.

**To Reproduce**
Parse the following code with `svelte2tsx`:
```svelte
<script>
    import { createEventDispatcher } from "svelte";

    const dispatch = createEventDispatcher();
    dispatch("mount", { input });
</script>

<input on:focus />
```

It also works, if you do it in the [Svelte2tsx REPL](https://embed.plnkr.co/plunk/JPye9tlsqwMrWHGv). There is a comma missing in the return value, in the events property.

**Expected behavior**
Output valid tsx.

 - OS: macOS, Browser
 - Plugin/Package: `svelte2tsx`, I use the npm package directly

hgiesel:
Fixes #886.

### DIFF
--- a/packages/svelte2tsx/src/svelte2tsx/nodes/ComponentEvents.ts
+++ b/packages/svelte2tsx/src/svelte2tsx/nodes/ComponentEvents.ts
@@ -204,12 +204,13 @@ class ComponentEventsFromEventsMap {
         if (this.eventDispatcherTyping) {
             return `__sveltets_toEventTypings<${this.eventDispatcherTyping}>()`;
         }
+
         return (
             '{' +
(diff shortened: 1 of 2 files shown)


jasonlyu123:
Can you add a new test case for it? Just add one in your issue into test/svelte2tsx/sample. Then run the test and it would generate an expected file.

dummdidumm:
I'm merging and adding a test myself afterwards so we can release this later today

PR #893, opened 2021-03-19
Add accessors autocomplete

vatro:
Add autocompletion for accessors.

vatro-svelte-language-tools-add-accessors-autocomplete


packages/svelte2tsx/src/svelte2tsx/nodes/exportsetters.ts
```diff
@@ -0,0 +1,12 @@
+export const createClassSetter = (name: string) =>
+    `\n
+    /**accessor*/\n${' '.repeat(4)}set ${name}(${name}) {}`;
```
dummdidumm, line 3: Could you explain why you add JSDoc here?

vatro: I though it would be cool to see in the autocomplete list that a prop is an accessor, so users immediately know it can be set / is not an exported readonly const:

image

dummdidumm: That makes sense, nice quality-of-live improvement 👍

dummdidumm:
Thanks for the PR! I'll deal with the lint/failing tests in a separate PR so we can get this in for today's release

vatro:
> Thanks for the PR! I'll deal with the lint/failing tests in a separate PR so we can get this in for today's release

I'm happy I could contribute! 😁 Keep up the great work! 👍🚀

PR #896, opened 2021-03-20
Separate sample generation from tests

pushkine:
Change in file generation is required for the `test/sourcemaps` rewrite

Instead of generating files _while_ running tests, this splits the process of writing files in two parts:
* On init: _create_
  + if one or several samples are new, skip tests to generate files
  + auto-creates `expected.error.json` if an error is thrown (prev had to create an empty file manually)

Code_2021-03-20_07-02-56

* On error: _update_
   + If `expected` files don't check out, update them and skip
  + Updates `expected.error` if it changed, still throws if it's unexpected or missing

e.g. after changing `{#if` block compilation
Code_2021-03-20_07-18-47

Also, 

* Added ANSI styling
* Made logged paths relative (still clickable)
* Made sample props private and added methods to interface with them
* Extracted the svelte2tsx config logic into a `get_svelte2tsx_config` function
* Replaced sample names by their path when they throw

Code_2021-03-20_07-23-49

### DIFF
--- a/packages/svelte2tsx/package.json
+++ b/packages/svelte2tsx/package.json
@@ -34,6 +34,7 @@
         "source-map-support": "^0.5.16",
         "svelte": "~3.35.0",
         "tiny-glob": "^0.2.6",
+        "ts-node": "^9.1.1",
         "tslib": "^1.10.0",
         "typescript": "^4.2.2"
     },
--- a/packages/svelte2tsx/package.json
+++ b/packages/svelte2tsx/package.json
@@ -34,7 +34,6 @@
         "source-map-support": "^0.5.16",
         "svelte": "~3.35.0",
         "tiny-glob": "^0.2.6",
-        "ts-node": "^9.1.1",
         "tslib": "^1.10.0",
         "typescript": "^4.2.2"
     },
--- a/packages/svelte2tsx/test/helpers.ts
+++ b/packages/svelte2tsx/test/helpers.ts
@@ -1,5 +1,5 @@
 import fs from 'fs';
-import assert from 'assert';
+import assert, { AssertionError } from 'assert';
 import { TestFunction } from 'mocha';
 import svelte2tsx from './build/index';
 import { htmlx2jsx } from './build/htmlxtojsx';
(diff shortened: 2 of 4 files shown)


dummdidumm:
These sound like nice changes! One question about "update": If I change something that results in a failing test, how can I prevent it from updating the file? I read that paragraph as "this always happens" which would not be what I want normally. I would expect to be able to turn it on somehow.

Regarding code style: there is a mix of camelCase and under_scores in the changed file now. Could you harmonize it to camelCase? Also add empty lines between each top level code item.

pushkine:
> I read that paragraph as "this always happens" which would not be what I want normally

On github it still throws an error, but locally it'll update every time, when would you not want it to ?

 Just so that we're on the same page, this is for `expected` tests (strict equal char to char exact match), custom tests (`expected.js`) are not affected

If an update you didn't expect occurred, you could revert it by fixing what's wrong and running tests again, but it's also easy to discard them all at once through source control (sorting by path they can be easily selected) if you really need to
Code_2021-03-20_09-59-42

dummdidumm:
I would not want this if I change something and an unrelated test suddenly throws. This would not be as obvious for me when it was auto-updated. I would also loose track of which tests changed intentionally and which ones by accident because they all end up in my changed git files. I suggest to make this a configuration option you can pass through the command line or by changing a flag in the test setup code.

pushkine:
I'd filter changes by staging the ones that are expected ( in source control puts in a different list without commiting)
It can be disabled by commenting out `tests/helpers.ts:100`, faking an env for `CI`, or filtering changes by staging the ones that are expected

I understand you're the one that's most affected by this so

pushkine:
> I would also loose track of which tests changed intentionally and which ones by accident because they all end up in my changed git files

Couldn't you use staging in that case ? It filters out changes you've checked as intended from the list of all changes
Code_2021-03-20_12-17-03

dummdidumm:
Then I would need to stage all upfront which I may not want to do when I'm not sure yet on all the changes. My workflow often is to stage things that I'm sure about and not stage those which I'm unsure about. In general I think automatically updating tests is not what you want. Your case is different because a lot of tests might change during the refactoring. This is why I suggest to make this configurable with the default being off.

pushkine:
Here it is opt-in with `--auto`

Code_2021-03-21_15-07-44

dummdidumm (approved):

PR #924, opened 2021-03-30
Style attribute hover/completion

jasonlyu123:
#381
圖片

Only added hover and completion as it's the easiest and doesn't require a full document scan.

### DIFF
--- a/packages/language-server/src/plugins/css/CSSPlugin.ts
+++ b/packages/language-server/src/plugins/css/CSSPlugin.ts
@@ -46,13 +46,13 @@ import { StyleAttributeDocument } from './StyleAttributeDocument';
 
 export class CSSPlugin
     implements
-    HoverProvider,
-    CompletionsProvider,
-    DiagnosticsProvider,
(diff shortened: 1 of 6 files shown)


packages/language-server/src/plugins/css/CSSPlugin.ts
```diff
@@ -113,10 +114,23 @@ export class CSSPlugin
...
+        const attributeContext = getAttributeContextAtPosition(document, position);
+        if (
+            attributeContext &&
+            this.inStyleAttributeWithoutInterpolation(attributeContext, document.getText())
+        ) {
+            const [start, end] = attributeContext.valueRange;
+            return this.doHoverInternal(new StyleAttributeDocument(document, start, end), position);
+        }
+        return null;
+    }
```
dummdidumm, line 132: please insert an empty line after this

dummdidumm (commented):
Very cool! I like the idea of an embedded on-the-fly document for this purpose to reuse all the existing functionality 👍 Added two small notes.

PR #926, opened 2021-03-30
Use AST types from `svelte` instead of `estree` in `htmlxtojsx`

pushkine:
Replaced declaration of `BaseDirective` with the one from `svelte` at `packages\svelte2tsx\src\interfaces.ts`

Replaced most `Node` types from `estree` to types from `svelte`
- if directive, by `BaseDirective`
- if attribute, by `Attribute extends BaseNode` ( adds `value` property ) 
- if text, by `Text`
- else by `BaseNode`

### DIFF
--- a/packages/svelte2tsx/src/htmlxtojsx/nodes/text.ts
+++ b/packages/svelte2tsx/src/htmlxtojsx/nodes/text.ts
@@ -1,7 +1,6 @@
-import { Node } from 'estree-walker';
 import MagicString from 'magic-string';
-
-export function handleText(str: MagicString, node: Node) {
+import { Text } from 'svelte/types/compiler/interfaces';
+export function handleText(str: MagicString, node: Text) {
     if (!node.data) {
         return;
     }
--- a/packages/svelte2tsx/src/htmlxtojsx/nodes/text.ts
+++ b/packages/svelte2tsx/src/htmlxtojsx/nodes/text.ts
@@ -1,5 +1,6 @@
 import MagicString from 'magic-string';
 import { Text } from 'svelte/types/compiler/interfaces';
+
 export function handleText(str: MagicString, node: Text) {
     if (!node.data) {
         return;
--- a/packages/svelte2tsx/src/interfaces.ts
+++ b/packages/svelte2tsx/src/interfaces.ts
@@ -6,6 +6,7 @@ import {
     MustacheTag,
     Text
 } from 'svelte/types/compiler/interfaces';
+
 export interface NodeRange {
     start: number;
     end: number;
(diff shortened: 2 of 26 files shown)


packages/svelte2tsx/src/interfaces.ts
```diff
@@ -1,6 +1,11 @@
-import { Node } from 'estree-walker';
 import { ArrayPattern, ObjectPattern, Identifier } from 'estree';
-
+import {
+    Directive,
+    TemplateNode,
+    Transition,
+    MustacheTag,
+    Text
+} from 'svelte/types/compiler/interfaces';
```
dummdidumm, line 8: please add an empty line after this

packages/svelte2tsx/src/htmlxtojsx/nodes/text.ts
```diff
@@ -1,7 +1,6 @@
-import { Node } from 'estree-walker';
 import MagicString from 'magic-string';
-
-export function handleText(str: MagicString, node: Node) {
+import { Text } from 'svelte/types/compiler/interfaces';
```
dummdidumm, line 2: please add an empty line after this

dummdidumm (commented):
This is nice! I thought about being a little stricter on those for a long time. If there's no type imports from `estree-walker` anymore, maybe https://github.com/sveltejs/language-tools/blob/master/packages/svelte2tsx/src/estree.d.ts can be removed. Could you check that?
Also I'm wondering if we should reexport the Svelte compiler types from `interfaces.ts` so that both the reexports and slight modifications of it are grouped in one place - what do you think?

pushkine:
> If there's no type imports from estree-walker anymore, maybe https://github.com/sveltejs/language-tools/blob/master/packages/svelte2tsx/src/estree.d.ts can be removed

`Node` from `estree-walker` is still good where nodes are purely javascript (expressions & co)

> Also I'm wondering if we should reexport the Svelte compiler types from interfaces.ts so that both the reexports and slight modifications of it are grouped in one place - what do you think?

It is the responsibility of `sveltejs/svelte` to export proper types, alas most of the html ast is completely untyped

dummdidumm:
> > Also I'm wondering if we should reexport the Svelte compiler types from interfaces.ts so that both the reexports and slight modifications of it are grouped in one place - what do you think?
> 
> It is the responsibility of `sveltejs/svelte` to export proper types, alas most of the html ast is completely untyped

I meant this in the sense of
```
import { .. } from 'svelte/types/compiler';
// .. other definitions
export { ..stuff from types/compiler alongside modified interfaces }
```

pushkine:
Ah okay, I thought you meant public library export

I mean I guess you could, the reason I made this PR was to at least get accurate types in the functions transforming directives 

I don't think it's really worth working much on this issue otherwise as it's an upstream job

PR #930, opened 2021-03-31
(feat) support ESM configs

dummdidumm:
With this PR config loading switches to using dynamic import statements. The PR consists of
- rework of the config loader to make sure to not load the same config twice in parallel, and to load all possible sub-svelte-configs upon tsconfig initialization, so that all configs are available synchronously to all snapshots that are retrieved. throwing out cosmiconfig in that process.
- resulting tedious `make this function async` modifications since the ts service initialization is now async


jasonlyu123:
I'll take a look next week.

ryanfiller:
This is the greatest PR I've ever seen in my life (well, in the last month anyways). Any idea what release this will be a part of? It doesn't look like this is working yet as of `1.0.0-next.1`.

dummdidumm:
I think SvelteKit is hardcoded to using `svelte.config.cjs` . I wanted to wait a little for possible "this broke" feedback before proceeding with a PR over at SvelteKit

ryanfiller:
Ah, gotcha. I was linked to this PR from the Discord and didn't pay attention to what repo this is in. Either way, super great work! I hope this is able to land in Kit eventually, it would make my life (and hopefully other's) a LOT easier.

PR #950, opened 2021-04-18
Don't watch node_modules when using the FallbackWatcher

Gelio:
`FallbackWatcher` used in the language server was watching files in the `node_modules` and `.git` directories, which resulted in a V8 heap out of memory and unbearable performance when using the language server standalone, without VSCode (e.g. using neovim LSP).

## Investigation

When using [the svelte language server in neovim LSP](https://github.com/neovim/nvim-lspconfig/blob/master/CONFIG.md#svelte) I have noticed very poor performance (when asking for a hover message, I got it after a minute, with the language server crashing right after that).

I checked in `~/.cache/nvim/lsp.log` that the language server was initializing a new TS server for each `tsconfig.json` it found in `node_modules`:

<details>
<summary>Logs with a V8 crash at the end</summary>

```
[ START ] 2021-04-18T13:18:21+0200 ] LSP logging initiated
[ ERROR ] 2021-04-18T13:18:22+0200 ] ...nt_nvimzcvtFx/usr/share/nvim/runtime/lua/vim/lsp/rpc.lua:457 ]  "rpc"   "svelteserver"  "stderr"        "Initialize language server at  file:///home/voreny/projects/personal/supreme-trobot\n"
[ ERROR ] 2021-04-18T13:18:22+0200 ] ...nt_nvimzcvtFx/usr/share/nvim/runtime/lua/vim/lsp/rpc.lua:457 ]  "rpc"   "svelteserver"  "stderr"        "Initialize new ts service at  /home/voreny/projects/personal/supreme-trobot/tsconfig.json\n"
[ ERROR ] 2021-04-18T13:18:22+0200 ] ...nt_nvimzcvtFx/usr/share/nvim/runtime/lua/vim/lsp/rpc.lua:457 ]  "rpc"   "svelteserver"  "stderr"        "Trying to load configs for /home/voreny/projects/personal/supreme-trobot\n"
[ ERROR ] 2021-04-18T13:18:22+0200 ] ...nt_nvimzcvtFx/usr/share/nvim/runtime/lua/vim/lsp/rpc.lua:457 ]  "rpc"   "svelteserver"  "stderr"        "Loaded config at  /home/voreny/projects/personal/supreme-trobot/svelte.config.js\n"
[...]

PR #955, opened 2021-04-22
(perf) Optimize fallback watcher

jasonlyu123:
One of the reasons for problem #950 and https://github.com/sublimelsp/LSP-svelte/issues/44 encountering is devastating is that the chokidar would trigger an initial event. That causes all the js/ts files to be scanned on startup. Rather than only in `npm i` as I originally thought. 

Even if we are excluding node_modules now. This behavior still needs to be disabled. So we can save all those IO operations on startup. And delay the language service creation when not needed. 

Another change is to denounce the watch file callback. Squashing multiple changes into one. This means fewer update project files call when switching branches or when generated files are written to the file system.

### DIFF
--- a/packages/language-server/src/lib/FallbackWatcher.ts
+++ b/packages/language-server/src/lib/FallbackWatcher.ts
@@ -18,7 +18,12 @@ export class FallbackWatcher {
                     gitOrNodeModules.test(path) &&
                     // Handle Sapper's alias mapping
                     !path.includes('src/node_modules') &&
-                    !path.includes('src\\node_modules')
+                    !path.includes('src\\node_modules'),
+
+                // typescript would scan the project files on init.
+                // We only need to know what got updated.
+                ignoreInitial: true,
+                ignorePermissionErrors: true
             }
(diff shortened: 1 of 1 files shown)


packages/language-server/src/lib/FallbackWatcher.ts
```diff
@@ -18,32 +22,48 @@ export class FallbackWatcher {
...
-        const para = this.convert(path, type);
-        this.callbacks.forEach((callback) => callback(para));
+    private onFSEvent(path: string, type: FileChangeType) {
+        const fileEvent = this.convert(path, type);
+
+        this.undeliveredFileEvents.push(fileEvent);
+        this.scheduleTrigger();
     }
 
+    private readonly scheduleTrigger = debounceSameArg<void>(
```
dummdidumm, line 54: We have 'lodash' as a dependency, you can simplify this and use their `debounce` function for this case

dummdidumm (approved):
👍 thanks for digging into this!

PR #956, opened 2021-04-22
fix excluded file got added to snapshot manager

jasonlyu123:
This fixes the underlying problem for #950 and https://github.com/sublimelsp/LSP-svelte/issues/44. The files are added into snapshot manager no matter if it's updated or added. This caused memory to blow up alongside the problem of the fallback watcher. The fallback watcher is not the actual problem.


dummdidumm:
Thanks! Rereading this, I think I introduced this bug about one month ago while refactoring the SnapshotManager. I changed the behavior back then to create new snapshots if one doesn't exist already. This should fix it.

jasonlyu123:
Part of it also my fault. The intention of the original code is not very clear 😄.

PR #957, opened 2021-04-22
feat(svelte2tsx): add aria jsx type declarations

theurgi:
Copied directly from [@types/react](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts)


dummdidumm:
Thank you! Now that I look at it I'm very confused why there were no error previously. From my experiments it seems that as soon as there's a dash in the attribute, JSX will just accept it, regardless of it being defined or not.

Issue #550, opened 2020-09-18
TS: Named export from module script not recognized


IgnusG:
**Describe the bug**
Importing named exports from `<script context='module'>` returns an error saying:

Module '"*.svelte"' has no exported member ...

![image](https://user-images.githubusercontent.com/6438760/93609638-64529600-f9cc-11ea-95f9-ec8d23ab03e8.png) 

> The image is from a test file that imports both the component and the helper method

The svelte component looks like this:

```svelte
<script context="module" lang="ts">
export const reverseValue = (min: number, max: number) => (
  scale: number,
): number => max - scale + min;
</script>

...
```

<details>
<summary>Typescript Config</summary>
<pre>
{
  "extends": "@tsconfig/svelte/tsconfig.json",
  "compilerOptions": {
    "baseUrl": ".",
    "paths": {"*": ["*", "typings/*"]},
    "noImplicitAny": true,
    "emitDecoratorMetadata": true,
    "skipLibCheck": true,
    "experimentalDecorators": true,
    "esModuleInterop": true,
    "noImplicitReturns": true,
    "noUnusedParameters": true,
    "strictFunctionTypes": true,
    "strictNullChecks": true,
    "types": ["jest", "chrome", "node"],
    "allowJs": true,
    "isolatedModules": false,
    "target": "ESNext",
    "allowSyntheticDefaultImports": true
  },
  "include": ["src/**/*.ts", "src/**/*.tsx", "src/**/.svelte", "typings/*.d.ts", "test/*.ts"],
  "exclude": ["node_modules"]
}
</pre>
</details>

- Svelte: `3.24.1`
- Svelte Preprocess: `4.2.1`
- Svelte TS Config: `1.0.10`
- Typescript: `4.0.2`

**Expected behavior**
The exports should be recognized

**Severity**
Not severe. Just a small hindrance

IgnusG:
I'm actually using this awesome plugin by @mrmckeb https://github.com/mrmckeb/typescript-plugin-css-modules :)  
It works very well at least within VS Code.

For the compilation step maybe it could be possible to add an option to the [typescript preprocessor for svelte-preprocess](https://github.com/sveltejs/svelte-preprocess/blob/master/src/transformers/typescript.ts) to generate the declaration for every svelte component while running. If I a tool uses svelte-preprocess it would then immediately get the correct typings

jasonlyu123:
[...]

Issue #342, opened 2020-07-23
Go to definition on Svelte component from ts file takes you to SvelteComponentDev interface


opensas:
**Describe the bug**

When you ctrl-click or right-click - Go to definition on a Svelte Component from a ts file you are taken to source code of the SvelteComponentDev Interface

**To Reproduce**

Create a new ts project:

```shell
npx degit sveltejs/template svelte-typescript-app
cd svelte-typescript-app
node scripts/setupTypeScript.js
```

Open vscode

Open main.ts

in the `import App from './App.svelte';` line, ctrlk-click on `App` or on `./App.svelte`

**Expected behavior**
Take you to the App.Svelte file, like what happens when you do it from a .svelte file

**Screenshots**
![svelte-vscode-bug](https://user-images.githubusercontent.com/481687/88343313-83f47780-cd17-11ea-9289-26682a1f967a.gif)

**System (please complete the following information):**
 - OS: Ubuntu 18.04
 - IDE: vscode
 - Plugin/Package: "Svelte for VSCode"

jasonlyu123:
I don't know if this is possible. We currently don't provide any language feature on js/ts files. And the go to the definition is done by the typescript language feature of vscode. I don't know if there is any way to inject into their result.

jasonlyu123:
Alternatively, anyone who wanted to quickly jump to the svelte file from any file. Use Ctrl+P to search the file.

opensas:
I suppossed it wasn't going to be so easy to handle .ts files, just wanted to leave it here in case anybody finds out how to handle it.

BTW, great job, the plugin is marvellous!

dummdidumm:
There exists a TypeScript plugin now which comes packaged with the VS Code extension and which you need to enable [through the settings](https://github.com/sveltejs/language-tools/tree/master/packages/svelte-vscode#svelteenable-ts-plugin). It also is [available standalone as a npm package](https://www.npmjs.com/package/typescript-svelte-plugin) if you need to use it outside of VS Code.
The plugin fixes this issue.

CoolOppo:
I have it enabled in the settings and even added the standalone npm package as a plugin, but I still have the issue.

PR #978, opened 2021-05-03
(feat) TypeScript plugin

dummdidumm:
Initially support
- rename (doesn't work for prop renames yet, I'm okay with that)
- diagnostics
- find references
- go to definition

This makes all files TSX hardcoded for now, it seems the TS server is okay with importing tsx into js.

To test this, add 
```
        "typescriptServerPlugins": [
            {
                "name": "typescript-svelte-plugin",
                "enableForWorkspaceTypeScriptVersions": true,
                "path": "typescript-plugin"
            }
        ],
```

to the `contributes` property in the vscode extension's `package.json`

In contrast to #842 this does not set `"languages": ["svelte"]` which means TS will not take over Svelte files. The advantage is that diagnostics, rename etc within Svelte files stays in the control of the language-server, and only inside TS/JS files Svelte support is added. The drawback is that you need to save file changes to disk for Svelte files in order for the TS server to notice it.

Next steps after merge:
- publish plugin to npm
- make it opt-in in the VS Code extension
- check if we can move some common stuff into a shared folder in the monorepo; maybe we need to publish another npm package

#580
#550
#342
#110

@jasonlyu123 would be great if you could take a look 😃


packages/typescript-plugin/src/language-service/definition.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export function decorateGetDefinition(
+    ls: ts.LanguageService,
+    snapshotManager: SvelteSnapshotManager,
+    logger: Logger
+): void {
+    const getDefinitionAndBoundSpan = ls.getDefinitionAndBoundSpan;
+    ls.getDefinitionAndBoundSpan = (fileName, position) => {
+        const definition = getDefinitionAndBoundSpan(fileName, position);
+        if (!definition || !definition.definitions) {
```
jasonlyu123, line 14: Maybe `if (!definition?.definitions`)? You can ignore this if you think the original code is more clear.

jasonlyu123 (changes requested):
Looks good. It's indeed a joy to see the svelte files show up in the ts language features. I listed some minor issues. Another feature I think can be decorated is `getImplementationAtPosition`. It's mostly the same as `getDefinition`. Should be easy to add.

dummdidumm:
When does `getImplementationAtPosition` get called? How could I test this?

jasonlyu123:
https://code.visualstudio.com/docs/editor/editingevolved#_go-to-implementation
This one. Or maybe we can do it in another PR.

PR #982, opened 2021-05-04
Don't use `toString`, this could be a RawSourceMap

### LINKED ISSUE #981: `createSourceMapper` incorrectly uses `.toString()` to pass maps to `SourceMapConsumer`
**Describe the bug**
Parsing of source maps fails, this shows up in VSCode as a warning -  `Unexpected token o in JSON at position 1svelte` -
on the first line of Svelte typescript files.

**To Reproduce**
Setup a project with Svelte and Typescript; open project in VS Code.

**Expected behavior**
There should be no error.

**Screenshots**

image

 - OS: Windows
 - IDE: VSCode
 - Plugin/Package:  svelte-language-server

** Observations **

The issue is in this line of code:

https://github.com/sveltejs/language-tools/blob/80828d60b421059260cb7cd0316f135cf5471d7e/packages/language-server/src/plugins/svelte/SvelteDocument.ts#L253

It uses `.toString()` to pass the value into the constructor of `SourceMapConsumer`. This is not needed, and it actually breaks the case where the source map is passed as an object - since the map becomes the string value `'[object Object]'`.

The source map should be passed as is, since `SourceMapConsumer` is able to handle both cases - see the `_factory` method here:

https://github.com/mozilla/source-map/blob/58819f09018d56ef84dc41ba9c93f554e0645169/lib/source-map-consumer.js#L1053

I propose to drop the `.toString()` - should I create a PR?

fabio-at-kumandgo:
This solves #981 - when `toString()` is used on an actual raw source map the receiving end will try to parse the string `[object Object]` as JSON, of course failing miserably.

By removing the `toString` and casting to `string | RawSourceMap` we handle both cases correctly, since the `SourceMapConsumer` is able to handle both the case where we are passing a JSON string and a raw source map object.

### DIFF
--- a/packages/language-server/src/plugins/svelte/SvelteDocument.ts
+++ b/packages/language-server/src/plugins/svelte/SvelteDocument.ts
@@ -1,4 +1,4 @@
-import { SourceMapConsumer } from 'source-map';
+import { RawSourceMap, SourceMapConsumer } from 'source-map';
 import { PreprocessorGroup, Processed } from 'svelte/types/compiler/preprocess/types';
 import type { compile } from 'svelte/compiler';
 import { CompileOptions } from 'svelte/types/compiler/interfaces';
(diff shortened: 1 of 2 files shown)


dummdidumm:
Thanks!

PR #1002, opened 2021-05-13
fix code action for ts-checked files

jasonlyu123:
Currently, code action doesn't work most of the time on ts-checked svelte files.  i.e. files with `checkJs` in the compiler option or files with `// @ts-check` comment. This is due to the `ts-nocheck` code action has a negative line. 

This PR fixes the code action for it. Besides that, also fixes some issues I found after the code action is enabled. Like the `ts-ignore` code action for props.


dummdidumm:
👍

PR #1003, opened 2021-05-13
Make on:sveltekit:xy events optional

dogagenc:
Latest version of svelte2tsx requires `on:sveltekit:xy` events which breaks applications that use `<svelte:window>` without on:sveltekit:xy events.

I have made those events optional.


dummdidumm:
Silly me, thanks for the catch!

Issue #431, opened 2020-08-05
Support custom events on native DOM elements


johanbissemattsson:
**Is your feature request related to a problem? Please describe.**
I get the following error when I'm trying to add custom events on native DOM elements: 
```
Type '{ onswipestart: (event: CustomEvent<any>) => void; onswipemove: (event: CustomEvent<any>) => void; onswipeend: (event: CustomEvent<any>) => void; style: string; class: string; }' is not assignable to type 'HTMLProps<HTMLDivElement>'.
  Property 'onswipestart' does not exist on type 'HTMLProps<HTMLDivElement>'.ts(2322)
```
The custom events are dispatched to a `div` element using Sveltes actions/use directive https://svelte.dev/examples#actions).

**Describe the solution you'd like**
Be able to type check for custom events dispatched using Svelte actions on native DOM elements.

**Describe alternatives you've considered**
I could try to convert the div to an individual Svelte component but preferably it should work on native DOM elements as well.

**Additional context**
This is how I listen to the custom events:
```
<div
  class="bottomSheet"
  class:draggable
  bind:this={bottomSheet}
  use:swipeable
  on:swipestart={handleSwipeStart}
  on:swipemove={handleSwipeMove}
  on:swipeend={handleSwipeEnd}
  style="height:{height};bottom:{bottom};transform:translateY({$coords.ty}px);"
>
```

Related Pull request for custom events on Svelte components: https://github.com/sveltejs/language-tools/pull/303

jasonlyu123:
Don't know if there is a way to make it only applies to elements with the action. but you can make it globally available like this.

```ts
declare namespace svelte.JSX {
    interface HTMLAttributes<T> {
        on:swipemove?: (event: CustomEvent<number> & { target: EventTarget & T }) => any;
    }
}
```

dummdidumm:
We could also enhance our typings with "fall back to `CustomEvent<any>` but I'd rather not do it because others may rely on the typings to throw a "does not exist" error if they mistype an event. I would be okay with something like that if it's only applied to elements with actions, but I'm not sure how we would implement that.
To infer the type from the action is an almost impossible task I think. Maybe we can find a way to let the user explicitly type the action and its possible events.

johanbissemattsson:
Thanks, letting the user explicitly type the action or explicitly choose when to fall back to `CustomEvent<any>` would probably be the best case scenario but the solution by @jasonlyu123 is good enough for me now. Thanks!

[...]

PR #1010, opened 2021-05-17
(fix) treat unknown events on element with action as custom events

dummdidumm:
Svelte is very dynamic in its nature. One feature is to add actions to DOM elements which in turn means additional events could be dispatched on that element. For the default namespace, transform unknown events on an element with an action so that they are treated as custom events.
#431


dummdidumm:
I just realized this would break people who enhanced the definition with their own, because we don't know of those and would still transform to the `CustomEvent<any>` fallback. I don't think this is desireable.

dummdidumm:
Actions can now be typed in various ways; closing

Issue #927, opened 2021-03-31
Auto import in module script adds import in component script


ebeloded:
**Describe the bug**

When trying to use built-in [auto import feature](https://code.visualstudio.com/docs/languages/typescript#_auto-imports) inside of module context script, the import line gets added to the component script instead.

**To Reproduce**

Type `readable` (or any other importable value) inside of a module script and use suggested auto-import. If a component script block exists, the import will be added there, instead of being added to the module script block.

Note that `typescript.suggest.autoImports` feature should be set to `true` (which is default)

**Expected behavior**

Expected import statement to show up in the script block where suggestion originates.

**Screenshots**

https://user-images.githubusercontent.com/5859177/113086145-513a4180-920b-11eb-9bbf-44321e73e5c1.mov

**System (please complete the following information):**
 - OS: Mac OS
 - IDE: VSCode
 - Plugin/Package: Svelte for VSCode

PR #1011, opened 2021-05-17
(fix) better import completion locations

dummdidumm:
Add import to module script if trigger in there, same for instance script
#927


packages/language-server/test/plugins/typescript/testfiles/completions/importcompletions8.svelte
```diff
@@ -0,0 +1,7 @@
+<script context="module">
```
jasonlyu123, line 1: I added a `importCompletion8.svelte` file in last PR. Maybe we can rename it to `importcompletions8` for consistency. And rename this to `importcompletions9` to clear confusion

dummdidumm: Oh didn't see that somehow. Your suggestion makes sense

Issue #847, opened 2021-03-01
Why is a required prop's type "T | undefined" with strictNullChecks enabled?


AradAral:
Say you have the following component (it's using TypeScript, of course):

`MyComponent.svelte`:
```svelte
<script lang="ts">
    export let label: string;
</script>
...
```

As you can see, the `label` prop is required, I have not supplied a default value and thus it's considered a required prop.
Now, if I have [`strictNullChecks`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#--strictnullchecks) set to `true` in my `tsconfig.json` file, I expect a required component prop like `label` in this case to not be able to receive `null` or `undefined` unless I explicitly specify that in their type.

`tsconfig.json`:
```json
"compilerOptions": {
    "strictNullChecks": true
}
```

However, Svelte appears to make the type of the `label` prop `string | undefined` behind the scenes, and that means the consumer of this component could pass undefined as the value of `label`, which obviously defeats the whole purpose of enabling `strictNullChecks` in the first place.
```svelte
<MyComponent label={undefined} /> <!-- No error! Even though "strictNullChecks" is "true" + the type of "label" is "string" -->
```

AradAral:
@dummdidumm Hi, are you sure this is related to Svelte language tools and not the Svelte compiler itself?

dummdidumm:
This is a decision from the early days that bites us now. Essentially, we decided to make props optional if the `strict` compilerOption is not set. I'd very much like to revisit this decision and change it so that properties that have no initializer are treated as being required, which would make this work as expected for you (no `| undefined`).

AradAral:
@dummdidumm 
> I'd very much like to revisit this decision and change it so that properties that have no initializer are treated as being required, which would make this work as expected for you (no | undefined).

I think that would be great and it makes MUCH more sense that way.

dummdidumm:
Extension version 105 and `svelte-check` 2.0.0 now always mark uninitialized properties as required.

Issue #319, opened 2020-07-17
Confusing / wrong code completion


vatro:
**Describe the bug**
I'm experiencing confusing code completion with both plain JS and TS. Not sure if I've set up something wrong.

**To Reproduce**
see simple setup with a `Parent.svelte` and a `Child.svelte` component:

```javascript
// Parent.svelte
<script>
    import {onMount} from "svelte"
    import Child from "./Child.svelte"

    let child

    onMount(()=> {
      
    })
    
</script>

<Child bind:this={child} />`
```

```javascript
// Child.svelte
<script>
    export let foo
    export let bar = undefined

    export const baz = () => {
        console.log("executed function 'baz'!")
    }    
</script>

<div>
    foo is: <em>{foo}</em><br/>
    bar is: <em>{bar}</em>
</div>
```

#### **JS** with `<script></script>`:

![vscode_confusing_code_completition](https://user-images.githubusercontent.com/3778969/87813293-e950f000-c861-11ea-9676-dd11f36bf9ca.gif)

- Code completion on the Child component shows `bar`, `baz`, `foo`, `ref` etc.
expected: only `foo` and `bar`
    - why is `ref` there, what does it do?
    - I guess $$prop_def, $$slot_def, __sveltets_ensureType, __sveltets_partial are related to svelte2tsx (?)
    - Child / child?
    - onMount?

- `bar`, `baz`, `foo` are all listed as optional (?:)
expected:  `bar` optional, `baz` not listed, `foo` obligatory (see console warning).

- Code completion on `child`-reference is also confusing me, beside svelte2tsx related stuff etc.:
`foo` is listed, although it's not accessible (see console error), `baz` is not listed.
expected: `foo` is not listed, `baz` is listed

- Although `baz` is not listed I can call it and it works as expected (see console log)
 
#### **TS** version with `<script lang="typescript"></script>`:
```typescript
// Parent.svelte
<script lang="typescript">
    import {onMount} from "svelte"
    import Child from "./Child.svelte"

    let child:Child

    onMount(()=> {
      
    })
    
</script>

<Child bind:this={child} />`
```

```typescript
// Child.svelte
<script lang="typescript">
    export let foo: number
    export let bar: number = undefined

    export const baz = (): void => {
        console.log("executed function 'baz'!")
    }    
</script>

<div>
    foo is: <em>{foo}</em><br/>
    bar is: <em>{bar}</em>
</div>
```
[...]

PR #1030, opened 2021-05-26
(breaking) be more strict with required props

dummdidumm:
This change now always marks properties as required if they have no initializer. Previously this was only the case when using TypeScript with strict mode. For this to work for both TS and JS files, the transformation in svelte2tsx was adjusted. 

#319
#847 

### This is a breaking change
The fix is to provide a default value for properties that are optional. That's either the specific value kind or `undefined`.
Example TS:
`export let optional: string;`
Becomes
`export let optional: string = '';` (or just `export let optional = ''`)
Example JS:
`export let optional;`
Becomes
`export let optional = undefined;`


jasonlyu123 (approved):
Looks good to me👍

PR #1032, opened 2021-05-27
bumps typescript to 4.3 and support semantic tokens on js

jasonlyu123:
typescript allows semantic tokens to run on js in 4.3. Remove the doesn't-applied note.
The [import statement completion](https://devblogs.microsoft.com/typescript/announcing-typescript-4-3/#import-statement-completions) added in typescript 4.3 is a bit more complex. I'll add the support in another PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
         "lint": "prettier --check . && eslint \"packages/**/*.{ts,js}\""
     },
     "dependencies": {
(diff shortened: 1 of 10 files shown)


dummdidumm (approved):
👍 
Since TS 4.3 seems to have changed some other things as well, I'm starting to wonder if we should use `~latest minor` instead of `*` for TS inside the language server.

Issue #1037, opened 2021-05-29
Changes to TypeScript definition files in linked modules not detected by Language Server


jstarmx:
**Describe the bug**
Svelte Language Server does not pick up changes to TypeScript definition files in linked modules without being restarted.

**To Reproduce**
1. Clone this repo: https://github.com/jstarmx/svelte-typescript-test
2. Follow setup instructions
3. Open `/app/src/App.svelte`
4. Open `/package/src/Alert.d.ts`
5. Change the type of `text` on line 4 from `string` to something else, e.g. `number`
6. Observe that the problems panel in VS Code does not show any errors
7. Restart Svelte Language Server
8. Observe that the problems panel now shows an error in `App.svelte`

**Expected behavior**
Problems panel to display the error without needing to restart Svelte Language Server

**System (please complete the following information):**
 - OS: macOS Big Sur 11.3.1
 - IDE: VSCode
 - Plugin/Package: Svelte for VSCode

**Additional context**
Happy to try to contribute a fix if you think it's do-able/worthwhile/actually an issue...

dummdidumm:
There are two problems to this:
- File paths are not the same (case-sensitivity-wise), we need to normalize them
- If the service for a TS file is searched, it doesn't take into account existing services where the file is already present but where the file is not the child of that tsconfig. This means that `package/src/Alert.d.ts` is moved to another TS service. I think we need to change the update logic there to go through all existing services, too, and update them accordingly (99% of the time there will be only one).

jstarmx:
@dummdidumm thanks for looking into this so quickly, that's fantastic 👍

fkling:
Just wanted to note that I experienced the same problem in **neovim** (it works fine in VSCode). I was able to get it to work by configuring the lsp client to use the repo root as root directory. Something like

```lua
    require('lspconfig').svelte.setup {
      capabilities = capabilities,
      on_attach = on_attach,
      settings = servers.svelte,
      root_dir = require('lspconfig').util.find_git_ancestor
    }
```

I'm not familiar with all the lsp stuff to know why it works for other clients (e.g. tsserver) and other environments (e.g. vscode) but not in this case by default 🤷🏻‍♂️

jasonlyu123:
[...]

PR #1040, opened 2021-06-01
(fix) normalize paths, deduplicate snapshots

dummdidumm:
- normalize paths, as it's not guaranteed that they are in the same format coming from different sources
- deduplicate snapshots by introducing a global snapshot manager. A snapshot is unique across all services. If a snapshot is updated, that update should be reflected in all services

#1037


jasonlyu123 (approved):
Looks good 👍

Issue #273, opened 2020-07-03
Generic props


kingdaro:
**Is your feature request related to a problem? Please describe.**
I can't seem to be able to define a TS generic at the component level. This could be useful for components that can accept any value, and inverts control to the consumer. Simplified example:

```svelte
<script lang="ts">
	export let items: T[] // where do I put the T?
</script>

{#each items as item}
	<slot {item} />
{/each}
```
```svelte
<List items={['a', 'b', 'c']} let:item>
  <!-- as it stands, this can only be `any`, but it should be `string` -->
  <p>{item}</p>
</List>
```

**Describe the solution you'd like**
It seems like some sort of new syntax might be required here, but I'm not sure what that would look like

**Describe alternatives you've considered**
The only real alternative I can think of is `any` for items, which isn't ideal

jasonlyu123:
Maybe we could hijack the jsdoc's `@template` tag. It doesn't seem to be a jsDoc standard tag but typescript and google's closure-compiler both use it. If svelte2tsx detects it at the very top of the script make the exported class and render function generic.

dummdidumm:
Nice idea!

Output then could be something like this:

```tsx
<></>;function render<T>() {

    let b: T = 7;
;
<>
<div>
    <slot a={b}>Hello</slot>
</div></>
return { props: {b} as {b: typeof b}, slots: {default: {a:b}} }}

export default class<T> {
    $$prop_def = __sveltets_partial(render().props)
    $$slot_def = render<T>().slots
}
```

What I'm still not sure about:

- How to know from the user input that `T` is a generic parameter?
   - Option 1: Add some reserved type names like `$$Slot1`, `$$Slot2`, so users can type `export let items: $$Slot1` without having to type anything else? But what if users want to constraint the type of the slot (`T extends number`), how should they do this then?
   - Option 2: If users define an interface named `$$Slots`, recognize that and somehow transform that?
   - other options..?

- How can we make this work with javascript? We cannot use type annotations there. Is this something exclusive to TypeScript? Is it even possible to have generic types through jsdoc? Proposal: Only support this for typescript for now.

jasonlyu123:
typescript's generic type for jsdoc is done though [`@template`](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#template)

[...]

PR #1053, opened 2021-06-13
(feat) add experimal support for generics

dummdidumm:
The content of a $$Generic type is moved to the render function: Its identifier is the generic name, a possibly type argument is the extends clause.

#442 
#273 
https://github.com/sveltejs/rfcs/pull/38

@jasonlyu123  during that, the seemingly unnecessary typing of props/slots/events as being optional on the render function is removed in various helper functions because it makes TS throw wrong errors in strict mode. Do you remember why we ever made these optional?

### DIFF
--- a/packages/language-server/src/plugins/typescript/features/SemanticTokensProvider.ts
+++ b/packages/language-server/src/plugins/typescript/features/SemanticTokensProvider.ts
@@ -7,9 +7,10 @@ import {
 } from 'vscode-languageserver';
 import { Document, mapRangeToOriginal } from '../../../lib/documents';
 import { SemanticTokensProvider } from '../../interfaces';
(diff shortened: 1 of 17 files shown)


jasonlyu123:
Don't remember either. Maybe just wanted to make it optional so that the generated code could save some character but never did that.

dummdidumm:
Alright, this is what I thought, too 😄 Anything else about these changes or do they look good to you?

jasonlyu123:
I notice there's a wrong semantic token on the script tag. 
圖片

Also when There's an unused generic parameter it would also gray out the same character
```svelte
<script>
type C = $$Generic
</script>
```

dummdidumm:
Oh yeah, the mapping is wrong there, for now I think it's best to somehow skip these.

PR #1057, opened 2021-06-16
(feat) event completion/hover for components with a type definition

jasonlyu123:
Provide event hover info and completion for components with a type definition, mainly for components in component libraries.


dummdidumm:
W00t this is awesome! I was thinking about something like this recently (analyzing the TS AST ourselves) and now you open this PR 😄  I'll take a closer look later, excited for this!
Could something similar be used for the slot attributes on the default slot?

packages/language-server/src/plugins/typescript/ComponentInfoProvider.ts
```diff
@@ -0,0 +1,72 @@
...
+        const typeChecker = program.getTypeChecker();
+        const classType = typeChecker.getTypeAtLocation(defClass);
+
+        if (!classType) {
+            return null;
+        }
+
+        return new JsOrTsComponentInfoProvider(typeChecker, classType);
+    }
+}
```
dummdidumm, line 68: What do you think about making the instance of this a property on TsOrJsDocumentSnapshot? Then it could be cached and the cache would only be deleted on updates to the snapshot.

dummdidumm: Not sure about the performance implications, but could this be something we could also use for the svelte2tsx-components? Then we could ditch our own implementation of creating the "events" property on the svelte2tsx return type.

jasonlyu123: Cached in the snapshot seems like a good idea. 

> could this be something we could also use for the svelte2tsx-components

Yes. Tried with svelte2tsx-component. The biggest difference between this and the svelte2tsx approach is the fallback Event in the svelte2tsx.

dummdidumm:
Do you have time to add the caching logic the next few days or should I take over? This feature is too nice to not get it in as part of the next release 😄

jasonlyu123:
OK. I'll add the caching logic soon

dummdidumm (approved):
🎉

PR #1072, opened 2021-06-21
svelte2tsx to handle @const

### LINKED ISSUE #1067: Add support for `@const` feature
https://github.com/sveltejs/svelte/pull/6413 implements https://github.com/sveltejs/rfcs/pull/33 . We need to add support for it once we are sure the feature goes into Svelte core.

tanhauhau:
Fixes #1067


dummdidumm:
Oh wow, thank you very much for this 🎉 From a quick glance this looks good. 

Notes to self on what to check:
- how it behaves when there are `#if` blocks in the parent, possibly with conflicting names
- quick check on rename/diagnostics/go to definition work and possibly adding tests, at least for the diagnostics

PR #1073, opened 2021-06-24
normalize workspace uri

jasonlyu123:
Notice some language server client has their rootUri in a different format with the workspace Uri of vs code. Specifically, the drive path in windows is not encoded. This causes the tsconfig to wrongly assumed to be not in the subpath of the workspace. Normalize it to fix this.

This is what's logged in sublime LSP, this also happens in visual studio 2019
> Initialize language server at  file:///C:/projects/...
Initialize new ts service at

And in vscode:
> Initialize language server at  file:///c%3A/projects/...
Initialize new ts service at  c:/projects/...

### DIFF
--- a/packages/language-server/src/server.ts
+++ b/packages/language-server/src/server.ts
@@ -33,7 +33,7 @@ import {
     OnWatchFileChangesPara,
     LSAndTSDocResolver
 } from './plugins';
-import { debounceThrottle, isNotNullOrUndefined, urlToPath } from './utils';
+import { debounceThrottle, isNotNullOrUndefined, pathToUrl, urlToPath } from './utils';
 import { FallbackWatcher } from './lib/FallbackWatcher';
 
 namespace TagCloseRequest {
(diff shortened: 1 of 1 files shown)


packages/language-server/src/server.ts
```diff
@@ -135,7 +135,11 @@ export function startServer(options?: LSOptions) {
         pluginHost.register(
             new TypeScriptPlugin(
                 configManager,
-                new LSAndTSDocResolver(docManager, workspaceUris, configManager)
+                new LSAndTSDocResolver(
+                    docManager,
+                    workspaceUris.map((uri) => pathToUrl(urlToPath(uri) ?? '')),
```
dummdidumm, line 140: There's `normalizeUri`, any reason why you didn't use that one?

jasonlyu123: Totally didn't remember there's this function lol. No wonder it only happens here.

PR #1082, opened 2021-07-02
fix: unsatisfied peer dependency

NickDarvey:
Adds svelte as a dev dependency because svelte2tsx [has a presently unsatisfied peer dependency on it](https://github.com/NickDarvey/language-tools/blob/d219ec4719b6d5b9425905f656c98df43136e879/packages/svelte2tsx/package.json#L43).

This means it can't be used with [pnpm](https://pnpm.io/)'s [strict-peer-dependencies](https://rushjs.io/pages/maintainer/recommended_settings#strictpeerdependencies).

### DIFF
--- a/packages/typescript-plugin/package.json
+++ b/packages/typescript-plugin/package.json
@@ -20,7 +20,8 @@
     "devDependencies": {
         "@tsconfig/node12": "^1.0.0",
         "@types/node": "^13.9.0",
-        "typescript": "*"
+        "typescript": "*",
+        "svelte": "*"
     },
     "dependencies": {
         "svelte2tsx": "~0.4.0",


dummdidumm:
This makes sense, but why is this needed for you? devDependencies should only matter while developing the package, not while using it, right? This repo uses yarn, not pnpm (although this might change at some point).

NickDarvey:
Good question... I'm a total novice when it comes to webdev dependencies so maybe I don't need this! I opened it because pnpm was yelling at me about svelte2tsx having an unmet peer dependency when I pulled in the typescript-plugin.

dummdidumm:
But then adding Svelte to the devDependencies doesn't change anything I think because when installing the package, only things within dependencies are installed with it. Furthermore, if the peer dependency on Svelte is unsatisfied then the fix would be to install Svelte version 3.24 or higher, or wouldn't it?

NickDarvey:
>adding Svelte to the devDependencies doesn't change anything I think because when installing the package, 
Yes that's right.

>the fix would be to install Svelte version 3.24 or higher
Yes, that's also right. I swear I tried this first and it didn't resolve it, but I tried it now and 👍. Thanks @dummdidumm.

NickDarvey:
I don't know if this ought to be merge. It declares a true dependency, but it doesn't functionally change anything for contributors because svelte happens to be available from other packages in this repo pulling it in.

dummdidumm:
Closing as the original issue has been resolved, and as you said this doesn't change much.

PR #1083, opened 2021-07-03
(fix) fix event completion with createEventDispatcher or with multiple declarations of the same event

jasonlyu123:
The generic type parameter is not in the `valueDeclaration` property of the typescript symbol.

### DIFF
--- a/packages/language-server/src/plugins/typescript/ComponentInfoProvider.ts
+++ b/packages/language-server/src/plugins/typescript/ComponentInfoProvider.ts
@@ -57,14 +57,17 @@ export class JsOrTsComponentInfoProvider implements ComponentInfoProvider {
         return type
             .getProperties()
             .map((prop) => {
-                if (!prop.valueDeclaration) {
+                // currently there should only be one because svelte2tsx doesn't merge event dispatcher types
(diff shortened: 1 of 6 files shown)


packages/language-server/src/plugins/typescript/ComponentInfoProvider.ts
```diff
@@ -57,14 +57,17 @@ export class JsOrTsComponentInfoProvider implements ComponentInfoProvider {
         return type
             .getProperties()
             .map((prop) => {
-                if (!prop.valueDeclaration) {
+                // currently there should only be one because svelte2tsx doesn't merge event dispatcher types
+                // And interface ($$Events) can't merge property type
```
dummdidumm, line 61: Not sure how to interpret this sentence. In #938 merging multiple decent dispatcher definitions was done. Are completions working for these, too?
Not saying we need to fix that right away if not, since this is arguably an edge case

jasonlyu123: I tested it. Even if there're multiple declarations the type would still be correct. I think I'll remove the comment and add another test for union type.

Issue #969, opened 2021-04-26
Rename breaks attribute shorthand/binding


pushkine:
```svelte
<script>
	let value = "";
</script>

<input bind:value />
<input {value} />
```
<img width="220" alt="Code_2021-04-26_16-36-26" src="https://user-images.githubusercontent.com/30108880/116100848-9ab46a00-a6ad-11eb-99e3-bfa7ec3b2d66.png">
<img width="140" alt="Code_2021-04-26_16-36-41" src="https://user-images.githubusercontent.com/30108880/116100854-9c7e2d80-a6ad-11eb-832f-a8815b7c82a7.png">

dummdidumm:
@jasonlyu123 I saw you started implementing this, looking at the branch the approach seems promising. What's the status on that? If you don't have time to continue, I can take over if you want.

jasonlyu123:
There's some edge case where edits would have overlapping ranges and cause a weird rename. I'll try again this week.

PR #1087, opened 2021-07-07
fix binding/ shorthand props rename

jasonlyu123:
#969

~~Marking as a draft for now because I want to check what's the edge case I'm encountering before. Can't quite remember the problem and didn't found many problems now in the latest commit.~~ The problem probably is when `getAdditionalLocationsForRenameOfPropInsideOtherComponent` returns the same rename as the original rename. When that happens the duplicate one is not checked thus there are some overlapping rename locations. Since renaming props might also need the check for the shorthand, I also apply the check to other files. 

I also find another rename problem when props having `__sveltets_1_any` injected behind declaration in svelte2tsx. Maybe we can fix that in a separate PR.


dummdidumm (approved):
🎉

Issue #1102, opened 2021-07-16
Property 'referrerpolicy' does not exist on type 'HTMLProps<HTMLImageElement>'


dennisjlee:
I added the referrerPolicy property to an <img> tag and it failed on svelte-check with typescript enabled (`svelte-check --tsconfig ./tsconfig.json`).

```
Error: Type '{ src: any; referrerpolicy: string; }' is not assignable to type 'HTMLProps<HTMLImageElement>'.
  Property 'referrerpolicy' does not exist on type 'HTMLProps<HTMLImageElement>'. (ts)
```

Example source: 
```
<img src="https://domain.com/myImage.jpg"
         referrerPolicy="no-referrer" />
```

It looks like this would be fixed by adding referrerpolicy to the long list of accepted HTML props in https://github.com/sveltejs/language-tools/blob/master/packages/svelte2tsx/svelte-jsx.d.ts

dennisjlee:
@dummdidumm thank you so much for the quick review and release!

PR #1103, opened 2021-07-16
Include referrerpolicy in HTMLProps

dennisjlee:
This will allow using the referrerpolicy attribute without spurious errors from svelte-check.

Resolves sveltejs/language-tools#1102


dummdidumm (approved):
Thank you!

PR #1114, opened 2021-07-23
(fix) emitDts for components with no script

### LINKED ISSUE #1112: Incorrect type generation for components without script tag
**Describe the bug**
If a component doesn't have a script tag or has nothing in a script tag, the generated type class doesn't extend from `SvelteComponentTyped`. This causes the vscode extension to complain "cannot be used as JSX component" (Screenshot below).

**To Reproduce**

I'm using `svelte2tsx` `emitDts()` to generate the types.

**Given:** Section.svelte

```svelte
<section class="my-6">
	<slot />
</section>
```

**Output:**: Section.svelte.d.ts

```ts
/** @typedef {typeof __propDef.props}  SectionProps */
/** @typedef {typeof __propDef.events}  SectionEvents */
/** @typedef {typeof __propDef.slots}  SectionSlots */
export default class Section {
}
export type SectionProps = typeof __propDef.props;
export type SectionEvents = typeof __propDef.events;
export type SectionSlots = typeof __propDef.slots;
declare const __propDef: {
    props: {};
    events: {
        [evt: string]: CustomEvent<any>;
    };
    slots: {
        default: {};
    };
};
export {};
```

**Expected behavior**

```diff
/** @typedef {typeof __propDef.props}  SectionProps */
/** @typedef {typeof __propDef.events}  SectionEvents */
/** @typedef {typeof __propDef.slots}  SectionSlots */
- export default class Section {
+ export default class Section extends SvelteComponentTyped<{
+ }, {
+  ​[evt: string]: CustomEvent<any>;
+ }, {
+  default: {};
​+ }> {
}
export type SectionProps = typeof __propDef.props;
export type SectionEvents = typeof __propDef.events;
export type SectionSlots = typeof __propDef.slots;
+ import { SvelteComponentTyped } from "svelte";
declare const __propDef: {
    props: {};
    events: {
        [evt: string]: CustomEvent<any>;
    };
    slots: {
        default: {};
    };
};
export {};
```

**Screenshots**
Screenshot from 2021-07-23 15-23-24

 - OS: Ubuntu 21.04
 - IDE:VSCode
 - Plugin/Package: `svelte2tsx`

**Additional context**
I'm able to reproduce with local repo of `language-tools`. Trying to figure out why.

bluwy:
Fixes #1112

Explanation: The previous code that strips TSX doesn't take into account for the generated code without any script tag, which cuts off the first 4 characters. I've fixed this by applying a regex that strips of any TSX fragments generated.

### DIFF
--- a/packages/svelte2tsx/src/svelte2tsx/index.ts
+++ b/packages/svelte2tsx/src/svelte2tsx/index.ts
@@ -401,7 +401,7 @@ declare function __sveltets_1_createSvelteComponentTyped<Props, Events, Slots>(
         );
         let code = str.toString();
         // Remove all tsx occurences and the template part from the output
-        code = code.replace(/\<>.*?<\/>/gs, '').replace('() => ();', '');
+        code = code.replace(/<>.*?<\/>/gs, '').replace('() => ();', '');
         return {
             code
         };
--- a/packages/svelte2tsx/src/svelte2tsx/index.ts
+++ b/packages/svelte2tsx/src/svelte2tsx/index.ts
@@ -401,7 +401,7 @@ declare function __sveltets_1_createSvelteComponentTyped<Props, Events, Slots>(
         );
         let code = str.toString();
         // Remove all tsx occurences and the template part from the output
-        code = code.replace(/\<>.*?<\/>/gs, '').replace('() => ();', '')
+        code = code.replace(/\<>.*?<\/>/gs, '').replace('() => ();', '');
         return {
             code
         };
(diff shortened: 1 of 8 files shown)


bluwy:
Fixed CI issue

bluwy:
Changed the replace strategy to handle nested tsx fragments. It should work similar as before with the same generated code.

jasonlyu123:
Tested with other test sample files and didn't found any error. Thanks!

Issue #1115, opened 2021-07-23
Improved validation of empty or malformed `svelte.config.js`


benmccann:
**Is your feature request related to a problem? Please describe.**

SvelteKit had an empty `svelte.config.js` in one of its test projects. It gave the error message:

```
│ Error while loading config
│ TypeError: Cannot read property 'compilerOptions' of undefined
```

**Describe the solution you'd like**

It'd be nice if it checked if the `config` existed before trying to read `compilerOptions` off it:

https://github.com/sveltejs/language-tools/blob/7abeef43e997d1187e239f6d1d99253332b7f5bf/packages/language-server/src/lib/documents/configLoader.ts#L159

It'd also be nice if it printed the path to the `svelte.config.js` in the error. SvelteKit had tons of them in various test projects, so it took a little while to track down

**Additional context**
See https://github.com/sveltejs/kit/issues/1988

dummdidumm:
The error message should be better now.

PR #1119, opened 2021-07-28
improve error message while loading config

jasonlyu123:
#1115

throw an error when the default export is not found and showing the failed config file path

### DIFF
--- a/packages/language-server/src/lib/documents/configLoader.ts
+++ b/packages/language-server/src/lib/documents/configLoader.ts
@@ -152,6 +152,12 @@ export class ConfigLoader {
             let config = this.disabled
                 ? {}
                 : (await this.dynamicImport(pathToFileURL(configPath)))?.default;
+
+            if (!config) {
+                throw new Error(
+                    'Missing exports in the config. Make sure to include "export default config" or "module.exports = config"'
+                );
+            }
             config = {
                 ...config,
                 compilerOptions: {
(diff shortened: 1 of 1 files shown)


packages/language-server/src/lib/documents/configLoader.ts
```diff
@@ -152,6 +152,12 @@ export class ConfigLoader {
             let config = this.disabled
                 ? {}
                 : (await this.dynamicImport(pathToFileURL(configPath)))?.default;
+
+            if (!config) {
+                throw new Error(
+                    'Missing exports in the config. Make sure to include "export default config" or "module.exports = config"'
```
jasonlyu123, line 158: @benmccann It's this error message ok to you?

benmccann: Yeah, that looks good and would help make it clearer what the issue is. Thanks!

Even more than that I'd really like `--ignore` to avoid trying to load the config in the first place, but this is still helpful

jasonlyu123: One of the problems I can think of it's that the ignored file could still be imported by another file. And when the svelte-preprocess default language config is set in that errored config file. The ignored file might be recognized as the wrong script type(js or ts). Thus creating some confusing typing errors in like props type check. I'll have to think about how to suppress the error message more.

Issue #1126, opened 2021-07-30
emitDts fails for interfaces and classes defined in script tag


bluwy:
**Describe the bug**
If an interface or class is defined in the script tag, and is indirectly exported, e.g. `export let foo: BarInterface`, emitDts fails silently and outputs no respective file.

This is because the generated `dts` mode tsx by `svelte2tsx` is something like this:

```tsx
import { SvelteComponentTyped } from "svelte"

function render() {

  interface Hello {
    world: string
  }

   let foo: Hello
;
return { props: {foo: foo} as {foo: Hello}, slots: {}, getters: {}, events: {} }}
const __propDef = __sveltets_1_with_any_event(render());
export type InputProps = typeof __propDef.props;
export type InputEvents = typeof __propDef.events;
export type InputSlots = typeof __propDef.slots;

export default class Input extends SvelteComponentTyped<InputProps, InputEvents, InputSlots> {
}
```

Note the interface is in the `render()` function. Typescript will throw an error (e.g. `Return type of exported function has or is using private name 'Hello'.`) as the interface is scoped but it's being exported. See [example typescript playground](https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABACwKYBt1wBQEpEDeAUIojGFKgE7ACGEqiAEhloSaYgO5xXoAmALkQBnKFXIBzDgF8iHCOloiRiAGJw4hOR3SooiWsJaYtAXkLdeA4QHJatW4jmk9BgEaILYVF3Wa8eVIqfRAqJAJEAAcqOCiRYUjaABpETxlnIhkgA) showing the error.

However, if you're using normal `type`s, e.g. `type Foo = {}`, it'll work. [Example typescript playground](https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABACwKYBt1wBQEpEDeAUIolAJ4AOqiAEhlogLyGIDucATugCYBciAM5ROMMAHNEAXyIlE6VFEQBDAfUxxmrDt36IA5MuX7ps0p0UhOSAokqc4lQQNvLppqUA).

It's also important to note that this happens in `declaration: true` for tsconfig only.

**To Reproduce**

Clone https://github.com/bluwy/emit-dts-interface-repro, and follow steps in readme

**Expected behavior**
Generate types for Svelte components that have interfaces or classes in the script tag. Or specifically, `svelte2tsx` should extract these types out of the `render()` function for the dts-mode tsx output.

**Screenshots**
n/a

[...]

PR #1155, opened 2021-08-31
(fix) transform interface to type in dts mode

dummdidumm:
#1126

### DIFF
--- a/packages/svelte2tsx/src/svelte2tsx/processInstanceScriptContent.ts
+++ b/packages/svelte2tsx/src/svelte2tsx/processInstanceScriptContent.ts
@@ -36,7 +36,8 @@ export function processInstanceScriptContent(
     str: MagicString,
     script: Node,
     events: ComponentEvents,
-    implicitStoreValues: ImplicitStoreValues
+    implicitStoreValues: ImplicitStoreValues,
+    mode: 'tsx' | 'dts'
 ): InstanceScriptProcessResult {
     const htmlx = str.original;
     const scriptContent = htmlx.substring(script.content.start, script.content.end);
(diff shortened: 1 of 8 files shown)


packages/svelte2tsx/src/svelte2tsx/processInstanceScriptContent.ts
```diff
@@ -385,3 +394,33 @@ export function processInstanceScriptContent(
...
+function transformInterfacesToTypes(tsAst: ts.SourceFile, str: MagicString, astOffset: any) {
+    tsAst.statements.filter(ts.isInterfaceDeclaration).forEach((node) => {
+        str.overwrite(
+            node.getStart() + astOffset,
+            node.getStart() + astOffset + 'interface'.length,
+            'type'
+        );
+
+        if (node.heritageClauses?.length) {
+            const extendsStart = str.original.indexOf('extends', node.getStart() + astOffset);
```
jasonlyu123, line 407: The interface might also be generic and has generic constraints. I found we can use `node.heritageClauses.pos` as the starting point instead.

packages/svelte2tsx/src/svelte2tsx/processInstanceScriptContent.ts
```diff
@@ -385,3 +394,33 @@ export function processInstanceScriptContent(
...
+            str.overwrite(extendsStart, extendsStart + 'extends'.length, '=');
+
+            const extendsList = node.heritageClauses[0].types;
+            let prev = extendsList[0];
+            extendsList.slice(1).forEach((heritageClause) => {
+                str.overwrite(
+                    prev.getEnd() + astOffset,
+                    heritageClause.getStart() + astOffset,
+                    ' & '
```
jasonlyu123, line 416: There's still [interface merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#merging-interfaces). But this seems like an edge case though. Don't really know if anyone uses this feature in a svelte file.

dummdidumm: I hope noone does, but even if they do, this only affects the dts output, which is worse than it would be with this change.

packages/svelte2tsx/src/svelte2tsx/processInstanceScriptContent.ts
dummdidumm: Oh yeah totally forgot about that, good catch. I'll adjust and add a test.

Issue #1161, opened 2021-09-04
svelte-check returns errors and hints for reactive assignment of value read from inline object


rjkip:
**Describe the bug**
`svelte-check` returns errors and hints for reactive assignment of value read from inline object.

**To Reproduce**

1. Include this reactive assignment in a JavaScript or TypeScript Svelte component:

```js
// Reduced example
$: themeText = { light: "Light", dark: "Dark" }["light"];
```

2. Run `npm run check`, ie. `svelte-check --tsconfig ./tsconfig.json`
3. Get the following errors:

``` 
Error: ';' expected. (ts)
Error: ',' expected. (ts)
Hint: Unused label. (ts)
Error: Left side of comma operator is unused and has no side effects. (ts)
Error: Cannot find name 'dark'. (ts)
Error: Expected 1 arguments, but got 2. (ts)
```

<details><summary><strong>📜 Full logs for <code>npm run check</code></strong></summary>
    
    > ~TODO~@0.0.1 check
    > svelte-check --tsconfig ./tsconfig.json
    
    
    ====================================
    Loading svelte-check in workspace: /path/to/project
    Getting Svelte diagnostics...
    
    /path/to/project/src/lib/theming/ButtonThemeSwitcher.svelte:4:36
    Error: ';' expected. (ts)
    
    $: themeText = { light: "Light", dark: "Dark" }["light"];
    </script>
    
    
    /path/to/project/src/lib/theming/ButtonThemeSwitcher.svelte:4:44
    Error: ',' expected. (ts)
    
    $: themeText = { light: "Light", dark: "Dark" }["light"];
    </script>
    
    
    /path/to/project/src/lib/theming/ButtonThemeSwitcher.svelte:4:20
    Hint: Unused label. (ts)
    
    $: themeText = { light: "Light", dark: "Dark" }["light"];
    </script>
    
    
    /path/to/project/src/lib/theming/ButtonThemeSwitcher.svelte:4:27
    Error: Left side of comma operator is unused and has no side effects. (ts)
    
    $: themeText = { light: "Light", dark: "Dark" }["light"];
    </script>
    
    
    /path/to/project/src/lib/theming/ButtonThemeSwitcher.svelte:4:32
    Error: Cannot find name 'dark'. (ts)
    
    $: themeText = { light: "Light", dark: "Dark" }["light"];
    </script>
    
    
    /path/to/project/src/lib/theming/ButtonThemeSwitcher.svelte:4:44
[...]

PR #1162, opened 2021-09-04
(svelte2tsx) fix invalidate reactive for nested object literal

jasonlyu123:
#1161


packages/svelte2tsx/src/svelte2tsx/nodes/ImplicitTopLevelNames.ts
```diff
@@ -50,6 +55,29 @@ export class ImplicitTopLevelNames {
...
+
+        if (ts.isBinaryExpression(node)) {
+            return this.isNodeStartsWithObjectLiteral(node.left);
+        }
+
+        if (ts.isConditionalExpression(node)) {
+            return this.isNodeStartsWithObjectLiteral(node.condition);
+        }
+
+        return node
```
jasonlyu123, line 75: Should we have this fallback? Or we can just return false here. And fix it later if someone actually brings up the case I can't think of now.

dummdidumm: The fallback here looks reasonable, so I think we can leave it like this

dummdidumm (approved):

Issue #1136, opened 2021-08-13
Provide intellisense for event modifiers


navorite:
**Is your feature request related to a problem? Please describe.**
VS Code Svelte extension doesn't provide intellisense for event modifiers.

**Describe the solution you'd like**
Add intellisense for event modifiers.

dummdidumm:
Autocompletion and hover info for event modifiers are now available.

PR #1163, opened 2021-09-04
Event modifier completion/ hover

jasonlyu123:
#1136

### DIFF
--- a/packages/language-server/src/plugins/html/HTMLPlugin.ts
+++ b/packages/language-server/src/plugins/html/HTMLPlugin.ts
@@ -256,12 +256,7 @@ export class HTMLPlugin
 
         const offset = document.offsetAt(position);
         const node = html.findNodeAt(offset);
-        if (
-            !node ||
-            possiblyComponent(node) ||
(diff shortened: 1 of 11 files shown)


packages/language-server/src/plugins/html/HTMLPlugin.ts
```diff
@@ -102,12 +103,23 @@ export class HTMLPlugin
...
               // Other attributes/events would be false positives.
               CompletionList.create([])
             : this.lang.doComplete(document, position, html);
         const items = this.toCompletionItems(results.items);
+
+        items.forEach((item) => {
+            if (item.label.startsWith('on:') && item.textEdit) {
+                item.textEdit = {
+                    ...item.textEdit,
+                    newText: item.textEdit.newText.replace('="$1"', '$2="$1"')
```
jasonlyu123, line 118: Added a tab stop before the `=` sign. Making it the second one so that it less affect users' muscle memory

dummdidumm (approved):
Awesome work! Works like a charm.

PR #1158, opened 2021-09-01
(chore) bump ts

dummdidumm:
#1156

PR #1166, opened 2021-09-05
(svelte-check) npm run build failed in svelte-check

takurinton:
There was 1 error inside the catch block when building svelte-check.  
I used a branch by instanceof and avoided it.

```sh
npm run build 

... 

src/index.ts:95:24 - error TS2345: Argument of type 'unknown' is not assignable to parameter of type 'Error'.

95         writer.failure(err);
                          ~~~

Found 1 error.
```

### DIFF
--- a/packages/svelte-check/src/index.ts
+++ b/packages/svelte-check/src/index.ts
@@ -92,7 +92,9 @@ async function getDiagnostics(
         );
         return result;
     } catch (err) {
-        writer.failure(err);
+        if (err instanceof Error) {
+            writer.failure(err);
+        }
         return null;
     }
 }


jasonlyu123:
We haven't updated to typescript 4.4 yet. And this error is probably from typescript 4.4's new `useUnknownInCatchVariables` flag. There should still be multiple places where this error might happen in the codebase. I think we prefer dealing with it together after the update. It's currently blocked by an upstream issue.

dummdidumm:
This is somewhat of a subset of #1158 which will handle this in one go after TS releases it's patch. Therefore closing this as a duplicate. Still thanks for your contribution!

PR #1177, opened 2021-09-14
(feat) error-forgiving parser

dummdidumm:
This adds some more robustness to svelte2tsx with regards to missing properties. This assumes that there's a Svelte parser mode which will continue parsing no matter what and ignore all errors along the way, resulting in a possibly very wrong but somewhat functional AST.
If such a parser mode was implemented in Svelte, it should return the first error it encountered, so that svelte2tsx can fall back to throwing this error if it encounters an error while parsing the broken AST.
I tested this and it resulted in better intellisense in situations where Svelte would bail out on the first parsing error.


jasonlyu123:
A loose parser was added in another PR, and it has already been available since [language-server-0.17.8](https://github.com/sveltejs/language-tools/releases/tag/language-server-0.17.8) and [extensions-109.4.0](https://github.com/sveltejs/language-tools/releases/tag/extensions-109.4.0). But you'll need Svelte 5.13+. If you were using Svelte 5 and still have a problem with a specific case, you can open another issue with a reproduction.

PR #1192, opened 2021-10-03
(perf) WIP: TS incremental parsing

dummdidumm:
Implement getChangeRange of the ScriptSnapshot interface so TS can incrementally parse the result of svelte2tsx.
TODO:
- Doesn't reliably work, gets into a broken state seemginly random after some time
- Wasn't able to prove this is really faster (how to best test this?)

### DIFF
--- a/packages/language-server/src/plugins/typescript/DocumentSnapshot.ts
+++ b/packages/language-server/src/plugins/typescript/DocumentSnapshot.ts
@@ -20,6 +20,7 @@ import {
     isSvelteFilePath,
     getTsCheckComment
 } from './utils';
+import { performance } from 'perf_hooks';
 
 /**
  * An error which occured while trying to parse/preprocess the svelte file contents.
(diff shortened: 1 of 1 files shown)


packages/language-server/src/plugins/typescript/DocumentSnapshot.ts
```diff
@@ -437,3 +448,42 @@ export class SvelteSnapshotFragment implements SnapshotFragment {
...
+        if (oldSnapshot.version >= currentSnapshot.version) {
+            return undefined;
+        }
+
+        const oldText = oldSnapshot.getFullText();
+        const currentText = currentSnapshot.getFullText();
+
+        let diffStart = oldText.length - 1;
+        for (let i = 0; i < currentText.length; i++) {
+            if (oldText.charAt(i) !== currentText.charAt(i)) {
```
DanielRosenwasser, line 463: It's probably better to use `charCodeAt`. `chatAt` will likely allocate intermediate strings.

Issue #965, opened 2021-04-25
TS/JS file watchers watch too many files


dummdidumm:
**Describe the bug**
The watcher in `extension.ts` and the fallbackwatcher in `FallbackWatcher.ts` both watch too many files. They ignore node_modules, but they do not ignore build output directories. This means for example for Sapper that any changes in `__sapper__` are watched and the TS plugin adds them to the list of known files, which is wrong.

**To Reproduce**
Start a Sapper project, change files and force rebuilds, see the number of files increase over time.

**Expected behavior**
Build directories are ignored

This could be done on a best-effor basis by
- always ignoring `__sapper__` and `.svelte` (SvelteKit)
- try to read a `.gitignore` file in the root of the opened workspace and ignore file changes in all those listed paths

I'm not sure we can completely transform this into a glob though, because the glob DSL is very limited. We might need to watch all files either way but filter out the events that come from the watcher based on the `.gitignore`.

jasonlyu123:
Found that `tsserver` has a size limit for files. It would check the file sizes of the "project" loaded in. If the sizes exceed the 20MB limit. it would disable some features. 

https://github.com/microsoft/TypeScript/blob/3c8e45b304b8572094c5d7fbb9cd768dbf6417c0/src/server/editorServices.ts#L1955
https://github.com/microsoft/TypeScript/blob/3c8e45b304b8572094c5d7fbb9cd768dbf6417c0/src/server/project.ts#L397

From my understanding, It seems like the disabled project would only load the files that are opened in the client. The files in the include field of tsconfig.json would not be used. I have a brief experiment with adopting the check, and it looks promising. Even if I configured to load all js files in the `node_modules`. The editor would still be responsive.

dummdidumm:
Current state
- we no longer add files that are not part of the project include to the snapshot managers
- we now have this file size checker in place

What's left is to actually make the watcher watch less files, which I'm still not sure how to actually solve this.

PR #1194, opened 2021-10-04
(perf) typescript language service size limit

jasonlyu123:
#965
Add a size limit for typescript language service similar to the one in the tsserver. When the size limit exceeds, ts language service goes into a reduced mode. Project-wide IntelliSense is disabled. It would only use the files opened in the client. Kind of like when you don't have a tsconfig.json file. When this happens a warning message will show up.  

Noted if you have a really big project and you have excluded everything needed but it still exceeds this limit. You can disable it with the `disableSizeLimit` compiler options in the tsconfig/jsconfig.

### DIFF
--- a/packages/language-server/src/plugins/typescript/service.ts
+++ b/packages/language-server/src/plugins/typescript/service.ts
@@ -13,7 +13,7 @@ import {
     ignoredBuildDirectories,
     SnapshotManager
 } from './SnapshotManager';
-import { ensureRealSvelteFilePath, findTsConfigPath } from './utils';
+import { ensureRealSvelteFilePath, findTsConfigPath, hasTsExtensions } from './utils';
 
 export interface LanguageServiceContainer {
     readonly tsconfigPath: string;
@@ -411,11 +411,7 @@ function getFilenameForExceededTotalSizeLimitForNonTsFiles(
 
     const filesNames = snapshotManager.getProjectFileNames();
     for (const fileName of filesNames) {
-        if (
-            fileName.endsWith(ts.Extension.Dts) ||
-            fileName.endsWith(ts.Extension.Tsx) ||
-            fileName.endsWith(ts.Extension.Dts)
-        ) {
+        if (hasTsExtensions(fileName)) {
             continue;
         }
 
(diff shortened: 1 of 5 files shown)


packages/language-server/src/plugins/typescript/utils.ts
```diff
@@ -313,3 +313,11 @@ export function getDiagnosticTag(diagnostic: ts.Diagnostic): DiagnosticTag[] {
     }
     return tags;
 }
+
+export function hasTsExtensions(fileName: string) {
+    return (
+        fileName.endsWith(ts.Extension.Dts) ||
+        fileName.endsWith(ts.Extension.Tsx) ||
+        fileName.endsWith(ts.Extension.Dts)
```
dummdidumm, line 321: There's `Dts` in there two times. Duplicate or missing other extension?

packages/language-server/src/server.ts
```diff
@@ -241,6 +246,16 @@ export function startServer(options?: LSOptions) {
         };
     });
 
+    function notifyTsServiceExceedSizeLimit() {
+        connection?.sendNotification(ShowMessageNotification.type, {
+            message:
+                'Svelte language server detected a large amount of js/svelte files. ' +
+                'To enable project-wide JavaScript/TypeScript language features for svelte files,' +
```
dummdidumm, line 253: I would write JS and Svelte in capital letters here

packages/language-server/src/plugins/typescript/service.ts
```diff
@@ -350,4 +375,68 @@ async function createLanguageService(
...
+
+    let totalNonTsFileSize = 0;
+
+    const filesNames = snapshotManager.getProjectFileNames();
+    for (const fileName of filesNames) {
+        if (hasTsExtensions(fileName)) {
+            continue;
+        }
+
+        totalNonTsFileSize += ts.sys.getFileSize?.(fileName) ?? 0;
```
dummdidumm, line 418: Out of curiosity: Is this performance heavy to check that? How often is this called? My fear is that if this results in many IO operations, that it slows down the intellisense for people.

dummdidumm (commented):
Very nice! The approach looks good. Left some comments.

packages/language-server/src/plugins/typescript/service.ts
```diff
@@ -248,7 +248,14 @@ async function createLanguageService(
     }
 
     function updateProjectFiles(): void {
+        const projectFileCountBefore = snapshotManager.getProjectFileNames().length;
         snapshotManager.updateProjectFiles();
+        const projectFileCountAfter = snapshotManager.getProjectFileNames().length;
+
+        if (projectFileCountAfter <= projectFileCountBefore) {
```
dummdidumm, line 255: 👍

Issue #1220, opened 2021-10-28
Cannot read property 'tags' of null


easyfrog:
```
[2021-10-29 00:13:04.152] [renderer5] [error] Cannot read property 'tags' of null: TypeError: Cannot read property 'tags' of null
	at asCompletionItem (/Users/easyfrog/.vscode/extensions/svelte.svelte-vscode-105.4.2/node_modules/vscode-languageclient/lib/common/protocolConverter.js:232:46)
	at Array.map (<anonymous>)
	at asCompletionResult (/Users/easyfrog/.vscode/extensions/svelte.svelte-vscode-105.4.2/node_modules/vscode-languageclient/lib/common/protocolConverter.js:202:51)
	at async U.provideCompletionItems (/Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/workbench/services/extensions/node/extensionHostProcess.js:94:180703)
```

I use lastest version of `vscode`  and `Svelte for VS Code v105.4.2`

I got this error, when type code in script scope of svelte file.  I don't know where is wrong.

dummdidumm:
I would be interested in finding out why it crashed without a tsconfig/jsconfig. If there's any minimum reproducible / reproduction repo you could share that would be great.

PR #1231, opened 2021-11-06
Allow usage of alternative config file names other than `tsconfig.json` and `jsconfig.json`

Not-Jayden:
This change will allow alternative config file names other than the standard convention to be used (e.g. `tsconfig.dev.json`). Issue #1232 

Was confused why the `--tsconfig` flag for svelte-check wasn't working as expected when I included the filename, but realised it's just ignored inside this helper.

### DIFF
--- a/packages/language-server/src/plugins/typescript/utils.ts
+++ b/packages/language-server/src/plugins/typescript/utils.ts
@@ -1,4 +1,4 @@
-import { dirname } from 'path';
+import { dirname, basename } from 'path';
 import ts from 'typescript';
 import {
     CompletionItemKind,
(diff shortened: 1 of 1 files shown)


ThomasOrlita:
👍 I think this also fixes https://github.com/sveltejs/language-tools/issues/1220#issuecomment-954330976

dummdidumm:
This is not a correct change. When looking for a related tsconfig/jsonfig, TypeScript will take into account only `tsconfig.json` or `jsconfig.json`, in that order. As mentioned in #1232 already, I can't reproduce the original issue, passing a specific tsconfig via the flag works for me, so a proper reproduction is needed to investigate the underlying problem.

PR #1236, opened 2021-11-09
(perf) implement project version for typescript language service

jasonlyu123:
With the project version, typescript language service would skip rebuilding program when the version is the same. 
https://github.com/microsoft/TypeScript/blob/3ef3cdddb338b8f0e6a2c5971d255390b68654ac/src/services/services.ts#L1290

Even though in most cases the rebuild doesn't affect performance very much because most of the state is reused. But update import would be drastically impacted by this. As in its deep call hierarchy, it would try to check if the program object is up-to-date multiple times. So this small difference cumulated to be a big time.  

This mostly benefits update import and svelte-check. One of my projects improved check time from 30 seconds to 19 seconds. The same project previously can't even finish update import within 5 minutes. And it's now gone down to less than 1 minute. 

One other thing I want to discuss. I also found that if I remove the custom `readFile` implement with `getSnapshot` the updated import of the previously mention project even went down to less than 5 seconds. From what I found, the reason is that typescript uses `readFile` to read some source map files (.map) and their source file. There's a certain node module ship declaration files with sourcemap comment and the `src` folder for some reason. Because `getSnapshot` add files to the compilation list. It would cause typescript to constantly rebuild the program during update import. Wondering what should we do about this. Should we remove the custom `readFile` implementation? Or we can replace it with another way to cache the read file result?


packages/language-server/src/plugins/typescript/service.ts
```diff
@@ -148,13 +149,19 @@ async function createLanguageService(
...
+        getScriptKind: (fileName: string) => getSnapshot(fileName).scriptKind,
+        getProjectVersion: () => projectVersion.toString()
     };
+
     let languageService = ts.createLanguageService(host);
     const transformationConfig = {
         transformOnTemplateError: docContext.transformOnTemplateError
     };
 
+    docContext.globalSnapshotsManager.onChange(() => {
```
dummdidumm, line 161: Is this enough? When does a project version need to change? Looking at the code in the global snapshot manager, it doesn't emit a onChange event when someone updates a TS or JS file in `updateTsOrJsFile`.

dummdidumm (commented):
This sounds great, thank you for looking into this! 

About `readFile`: Do you mean removing the logic of calling `getSnapshot` inside `svelteSys.ts`? If so, what would we replace it with? I think right now we need that hook because else we can't intercept Svelte file reads that happen as a result of file import crawls. In the TypeScript plugin we do it a little differently, maybe we can adjust that here, too.

packages/language-server/src/plugins/typescript/service.ts
jasonlyu123: I think I'll just add the emit to the `updateTsOrJsFile` if branch that doesn't call the set method. Or do you prefer I add another event?

The project version needs to be updated every time there's a file change. So that typescript would rebuild.

dummdidumm: Adding the emit event to the branch where it's not called yet sounds good, don't think another event is needed.

dummdidumm (approved):

jasonlyu123:
Besides removing the logic of calling `getSnapshot` in svelteSys.readFile. One alternative is not to call it in specific situations, like It's a ts file in the node_modules folder. Or only call it when we know it's a source file like when it's a .ts, .js, .svelte (or .json even).

I'll do more research on the readFile problem. And do it in another PR. This PR should be enough for most of the users. The readFile problem only occurs when there's a  `.d.ts` file with source map comment.

Issue #1149, opened 2021-08-27
Svelte for VS Code extension crushes react project


florianGierlichs:
**Describe the bug**
After installing the Svelte for VS Code extension in my editor to work on a svelte project, a different project written in react and typescript completely crushes and throws ts error all over the place:
```
Type '() => svelte.JSX.Element' is not assignable to type 'FC<{}>'.
  Type 'Element' is not assignable to type 'ReactElement<any, any> | null'.
    Type 'Element' is missing the following properties from type 'ReactElement<any, any>': type, props, key
```

literally every .tsx file throw errors, the project worked flawlessly before installing the svelte extension.

The only way I found so far, is to disable the extension every time I'm not working on a svelte project, which sucks, you know.

**To Reproduce**

Open a react-ts project with the svelte extension in VSCode i guess.

**Expected behavior**
svelte extension doesn't influence non-svelte projects

**Screenshots**
![image](https://user-images.githubusercontent.com/58342773/131106390-e059e208-8646-4d2b-9861-abd21876e7be.png)

![image](https://user-images.githubusercontent.com/58342773/131106439-7d8640da-c9f9-42a5-af83-505559508a3f.png)

**System (please complete the following information):**
 - OS: MAC OS
 - IDE: VSCode
 - Plugin/Package: "Svelte for VSCode"

maykefreitas:
I'm having the same problem in a monorepo containing svelte and react projects.

maykefreitas:
It backs to work if I disable the Svelte extension. Looks that somehow the extension is injecting the `svelte-jsx.d.ts` file.

dummdidumm:
Could you check if the problem persists with the latest release of the Svelte extension (105.4.2) ?

jasonlyu123:
Wondering if the tsserver reuses the same language service for both the react project and the svelte project. Can you try if adding a `jsconfig.json` to the svelte project root solved the problem?

Here's an example `jsconfig`:

```json
{
    "include": ["src/**/*.d.ts", "src/**/*.js", "src/**/*.svelte"]
}

```

dummdidumm:
The most recent versions of the extension and `svelte-check` contain a new transformation behind a flag which should fix this issue. You can try out the new transformation using

- "Svelte for VS Code" version 105.11.1 and setting svelte.plugin.svelte.useNewTransformation to true
- svelte-check version 2.4.0 with the --use-new-transformation flag.

Please try it out, more info in #1352

Issue #1077, opened 2021-06-27
Reimagining svelte2tsx


dummdidumm:
svelte2tsx has come quite a long way and is successfully representing some rather advanced use cases of Svelte in a TSX-compatible way. But there are significant drawbacks that continue to pile up:
- Code that is needed to create the output and the output itself is hard to read because of all the workarounds we have to use, especially around shadowed variables and control flow
- The way we use MagicString is hard to read, and has some limitations around good source mapping, especially when comes to ranges (length is often short by one)
- Any feature that is not directly translateable to TSX is hard to implement and we need to work around it with lots of lambda functions
- The generated TSX needs some patches of the compiler options to work with the Svelte JSX namespace, which can interfere with other (f.e. React) projects (#1149)

What I envision is going away from a somewhat TSX-compliant structure and reimagine the translation more in line with our needs. One possible change is to no longer try to keep the HTML structure and instead focus on the control flow structure.

Given this input:

```svelte
{#if hello}
    <div>
    {#await hello.foo then y}
        {y}
    {/await}
    </div>
{/if}
```

We no longer do this:

```typescript
<>{(hello) ? <>
    <div>
    {() => {/*Ωignore_startΩ*/if(((hello))) {/*Ωignore_endΩ*/let _$$p = (hello.foo); __sveltets_1_awaitThen(_$$p, (y) => {/*Ωignore_startΩ*/((hello)) && /*Ωignore_endΩ*/<>
        {y}
    </>})}}}
    </div>
</> : <></>}</>
```

Instead we do this:

```typescript
if (hello) {
  const div = createElement('div');
  const y = await hello.foo;
  y;
}
```

What this means is that we no longer try to wrap the contents of a html tag with that tag, instead we place the tag at the beginning, close it at the beginning, and can continue writing our transformations without having to use lambdas because we no longer are restricted to certain operators within the tag.
Code manipulation-wise, this means a lot more usage of MagicString's `move`.

The following is a not yet complete list of transformation ideas:

#### await
```svelte
{#await foo}
...
{:then x}
...
{:catch e}
...
{/await}
```
```typescript
try {
   ...
   const x = await foo;
   ...
} catch (e) {
   ...
}
```

#### each
```svelte
{#each array as item, i (item.id)}
...
{:else}
...
{/each}
```
```typescript
for (let i = 0; i < 1; i++) { // 1 is just a hardcoded placeholder
  const item = __svelte_ts_getEntry(array);item.id;
  ...
}
...
[...]

PR #1237, opened 2021-11-09
(feat) rewrite svelte2tsx

dummdidumm:
Goal: Get rid of a tsx-style transformation and simplify transformations along the way
#1077
#1256
#1149 

@jasonlyu123 this is an early draft of how I imagine the svelte2tsx to look like. You can have a look at the new output by running the svelte2tsx-REPL (there's a VS Code run configuration) and looking at the logged result. I also tested this out running the extension and it seems to work for the cases implemented already. So far that's attributes/props/events (not all cases though). Let me know what you think!

Notes to self:
- right now `#each foo as foo` and `#await foo then foo` are both valid Svelte code but the transformation is invalid ("variable used before declaration"). Solution could be to move the first `foo` into a different block like `{const $$foo = foo; { const foo = await $$foo; } }`. *UPDATE* this is handled now
- `<Component validPropWrongType1={true} invalidProp1={true} />` will only show an error for `validPropWrongType1` and for `invalidProp1` only after the first error is fixed. It will also underline `validPropWrongType1` and not `true`. This is a hit I'm okay to take.


packages/language-server/test/plugins/typescript/features/HoverProvider.test.ts
```diff
@@ -6,184 +6,211 @@ import { Document, DocumentManager } from '../../../../src/lib/documents';
...
-        return { provider, document };
-
-        function openDoc(filename: string) {
-            const filePath = getFullPath(filename);
-            const doc = docManager.openDocument(<any>{
-                uri: pathToUrl(filePath),
-                text: ts.sys.readFile(filePath) || ''
-            });
-            return doc;
+function test(useNewTransformation: boolean) {
```
dummdidumm, line 14: The tests are mainly changed like this: Extract exising tests into a function which accepts `useNewTransformation` as a parameter and passes that to the `LSConfigManager`. There are some small adjustments in a few tests but mostly they run without any difference for both new and old transformations.

packages/language-server/src/ls-config.ts
```diff
@@ -45,6 +45,7 @@ const defaultLSConfig: LSConfig = {
     },
     svelte: {
         enable: true,
+        useNewTransformation: false,
```
dummdidumm, line 48: This is used to determine whether or not to use the new transformation. My idea is to go into a month-long beta phase where this is `false` by default and we encourage people to switch it on to try it out. After that we toggle it to `true` by default and then see what the feedback is. If everything is good, we can remove it ~2 months afterwards and remove all the old code, simplifying the code base.

This boolean is used in a few places in some Providers to do something differently.

packages/svelte2tsx/svelte-jsx.d.ts
```diff
@@ -1,5 +1,286 @@
 /// <reference lib="dom" />
-/* eslint @typescript-eslint/no-unused-vars: off */
+
+declare namespace svelte {
```
dummdidumm, line 3: This namespace definition is mostly a copy-paste from the `svelte.JSX` namespace. Reason: I want to get rid of the `JSX` fragment in there, it confuses people ("wait why is this JSX?"). The old version exists for backwards compatibility, for example we a section in our docs where we tell people to enhance the `svelte.JSX` namespace; we would break everyone who did this, so we should keep the old version 1-2 years.

packages/svelte2tsx/svelte-jsx.d.ts
```diff
@@ -1,5 +1,286 @@
...
+  // unit-less numbers are listed here (see CSSProperty.js in React)
+
+  // eslint-disable-next-line @typescript-eslint/no-empty-interface
+  interface DOMAttributes<T extends EventTarget> extends svelte.JSX.DOMAttributes<T> {}
+
+  // eslint-disable-next-line @typescript-eslint/no-empty-interface
+  interface AriaAttributes extends svelte.JSX.AriaAttributes {}
+
+  // eslint-disable-next-line @typescript-eslint/no-empty-interface
+  interface HTMLAttributes<T extends EventTarget> extends svelte.JSX.HTMLAttributes<T> {
```
dummdidumm, line 54: The new transformation has no problem with colons in the attribute name and doesn't transform them, so they are listed here.

packages/svelte2tsx/src/htmlxtojsx_v2/nodes/Element.ts
```diff
@@ -0,0 +1,253 @@
...
+ *
+ * A transformation reads for example like this:
+ * ```
+ * // before
+ * <div class={foo} />
+ * // after
+ * { const $$_div = __sveltets_2_createElement("div", {"class": foo,}); }
+ * ```
+ */
+export class Element {
```
dummdidumm, line 27: This is the core of transformations related to HTML elements as well as svelte:options, svelte:head, svelte:window, svelte:body. I hope the comments make it clear how it works.

packages/svelte2tsx/src/htmlxtojsx_v2/nodes/InlineComponent.ts
```diff
@@ -0,0 +1,239 @@
...
+ *
+ * A transformation reads for example like this:
+ * ```
+ * // before
+ * <Comp prop={foo} />
+ * // after
+ * { const $$_Comp = new Comp({ target: __sveltets_2_any(), props: {"prop": foo,}}); }
+ * ```
+ */
+export class InlineComponent {
```
dummdidumm, line 25: This is the core of transformations related components, including svelte:self and svelte:component. I hope the comments make it clear how it works.

packages/svelte2tsx/src/htmlxtojsx_v2/utils/node-utils.ts
```diff
@@ -0,0 +1,117 @@
...
+
+/**
+ * Moves or inserts text to a specific location in order.
+ * "In order" means that the transformation of the text before
+ * the given position reads exactly what was moved/inserted
+ * from left to right.
+ * After the transformation is done, everything inside the start-end-range that was
+ * not moved will be removed.
+ */
+export function transform(
```
dummdidumm, line 15: This is the heart of the new approach. This function ensure that we can write code that is better understandable by creating an array of transforms. They are passed to this function which then moves/appends text in order to the end position and then removes any code that was not moved, freeing us from having to cleanup stuff like `<` or `{` manually.

packages/svelte2tsx/src/svelte2tsx/index.ts
```diff
@@ -251,9 +260,15 @@ function processSvelteTemplate(
         }
     };
 
-    convertHtmlxToJsx(str, htmlxAst, onHtmlxWalk, onHtmlxLeave, {
-        preserveAttributeCase: options?.namespace == 'foreign'
-    });
+    if (options.mode === 'ts') {
```
dummdidumm, line 263: The code transformations for the script tags stays mostly untouched, the htmlx2jsx transformation is completely now however.

dummdidumm (commented):
I think the svelte2tsx part is mostly finished. @jasonlyu123 could you do a review? I know it's quite a lot, so I left some comments in here so you know where to start. Also you don't need to look at every bit, it's more about getting an understanding about how the new transformation works. If there are any questions, let me know. @halfnelson if you want, you can take a look, too.

jasonlyu123 (commented):
Overall looks good. Besides those listed in the review comment. I also found some other problems:

1. this would fail to transform because of magic string edits overlap
```svelte
<script> 
	import { readable } from "svelte/store"; 
	const store = readable(11); 
</script>

<div {...$store}></div> 
```

2. style directive and const tag doesn’t seem to have an `expectedv2.json`
3. if you rename component `on:some-event` it would only show the first character of the event name in the input.

packages/svelte2tsx/src/htmlxtojsx_v2/nodes/InlineComponent.ts
```diff
@@ -0,0 +1,141 @@
...
+                ...this.startEndTransformation,
+                ...this.eventsTransformation,
+                ...this.letBindingsTransformation
+            ]);
+
+            const endStart =
+                this.str.original
+                    .substring(this.node.start, this.node.end)
+                    .lastIndexOf(`</${this.node.name}`) + this.node.start;
+            transform(this.str, endStart, this.node.end, this.node.end, this.endTransformation);
```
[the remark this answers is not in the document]
dummdidumm: done

packages/svelte2tsx/src/htmlxtojsx_v2/nodes/InlineComponent.ts
```diff
@@ -0,0 +1,239 @@
...
+        return this.isSelfclosing
+            ? this.node.end
+            : this.str.original.lastIndexOf('>', this.node.end - 2) + 1;
+    }
+
+    private computeIsSelfclosing() {
+        if (this.str.original[this.node.end - 2] === '/') {
+            return true;
+        }
+        return !!this.str.original
```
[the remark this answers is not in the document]
dummdidumm: I removed it, not sure why I added it in the first place.

packages/svelte2tsx/src/htmlxtojsx_v2/nodes/InlineComponent.ts
```diff
@@ -0,0 +1,239 @@
...
+            this.propsTransformation.push(...name, ',');
+        }
+    }
+
+    /**
+     * on:click={xxx}  -->  $$_Component.$on("click", xxx)
+     * @param name Event name
+     * @param expression Event handler, if present
+     */
+    addEvent(name: [number, number], expression?: [number, number]): void {
```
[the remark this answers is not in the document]
dummdidumm: done

dummdidumm:
Thanks! I adressed your review comments. I also noticed that I should tweak Document Symbols a little, and that we need to update the TS plugin as well. Since there's no way to tell the TS plugin to either use the new or old version I decided to switch on the new version for the TS plugin by default. Need to test this more though.

@jasonlyu123 in your opinion, will this new transformation enable us to enhance svelte2tsx with new transformations more easily? Is the transformation code easier or worse to read now in your opion?

jasonlyu123:
I think It's more focused on the core transformation. Overall should be easier to read. 

On another note. I forgot to mention a problem yesterday. I found that the props transformation moves the props value now. So this causes the trailing property access to be removed. It's there because of the special preprocess before svelte parse. 

For example:
```svelte
<Component props={obj.} />
```

But this can also be done later. I'm thinking if we can detect this trailing property access while transforming. It should also be applied to the if block and await block so that the completion is right. 

Another thing I found it's worse noting for users to migrate in the future is: Because we're now transforming it into direct component class instantiate. If you have a component typedef like this
```ts
export class Component {
  $$prop_def: SomePropsType
}
```

You can change this to use the `SvelteComponentTyped` class exported by svelte

```ts
export class Component extends SvelteComponentTyped<SomeProps> {}
```

if you're extending a component class that has different props types. Be sure to also override the constructor type.

jasonlyu123 (approved):

PR #1238, opened 2021-11-12
(fix) slot let rename

jasonlyu123:
Currently, renaming on slot let would result in invalid code because the svelte2tsx transform it into a destructured variable declaration

```svelte
<Component let:a>
</Component>
```

become

```svelte
<Component let:a: newName>
</Component>
```

This PR uses the logic in the shorthand props and transform it into 
```svelte
<Component let:a={newName}>
</Component>
```

Also, fix a sourcemap issue with multiple slot-let. Currently it would the generated code would source map back with extra whitespace. When renaming it now, the whitespace would be present in the new name input.  After rename, the white space might be removed.


dummdidumm (approved):
👍

PR #1244, opened 2021-11-16
(fix) fall back to any type when array is of type any

dummdidumm:
Fixes a overzealous check introduced through #1218


jasonlyu123 (approved):
Looks good. Not quite certain how it works but it seems like `any` is the only one that takes priority over `unknown` in here.

dummdidumm:
Yeah I'm a little confused myself about this, almost like TS takes the results of the extends clause into account when deciding what to resolve to. Thanks for having a look!

PR #1255, opened 2021-11-29
Fix typo in svelte2tsx

janosh:
`EmitDtsConig` -> `EmitDtsConfig`

### DIFF
--- a/packages/svelte2tsx/index.d.ts
+++ b/packages/svelte2tsx/index.d.ts
@@ -55,7 +55,7 @@ export function svelte2tsx(
     }
 ): SvelteCompiledToTsx
 
-export interface EmitDtsConig {
+export interface EmitDtsConfig {
     /**
      * Where to output the declaration files
      */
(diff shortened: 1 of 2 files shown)


packages/svelte2tsx/index.d.ts
```diff
@@ -55,7 +55,7 @@ export function svelte2tsx(
     }
 ): SvelteCompiledToTsx
 
-export interface EmitDtsConig {
+export interface EmitDtsConfig {
```
jasonlyu123, line 58: This is technically a breaking change. I think we can export an interface with the old name and mark it as deprecated
```ts
/**@deprecated*/
export interface EmitDtsConig extends EmitDtsConfig {}
```

packages/svelte2tsx/src/emitDts.ts
```diff
@@ -2,21 +2,25 @@ import * as path from 'path';
...
 
-export interface EmitDtsConig {
+export interface EmitDtsConfig {
     declarationDir: string;
     svelteShimsPath: string;
     libRoot?: string;
 }
 
-export async function emitDts(config: EmitDtsConig) {
+// to make typo fix non-breaking, continue to export the old name but mark it as deprecated
```
dummdidumm, line 11: The Deprecated interface needs to be added to the `index.d.ts` file, not this file. The `index.d.ts` contains all the public API.

dummdidumm:
Thanks!

PR #1258, opened 2021-12-02
Fix broken link of README.md

baseballyama:
I found a broken link so I fixed it up.
But it contains guess so could you please check it is correct.


dummdidumm:
Thanks!

Issue #1260, opened 2021-11-21
Flashing semicolons in VSCode codeActionsOnSave organizeImports


janosh:
When adding `source.organizeImports` to

```json
  "editor.formatOnSave": true,
  "editor.codeActionsOnSave": ["source.fixAll", "source.organizeImports"],
```

I get flashing semicolons on save in Svelte files but not in JS or TS files.

https://user-images.githubusercontent.com/30958850/142756983-bb7a36f6-2151-43a5-a1d0-5bc17ae555c7.mp4

PR #1261, opened 2021-12-03
(fix) Adhere to code formatting settings when organizing imports

dummdidumm:
Add semicolon or not depending on Prettier config
#1260

Doesn't work yet, for some reason TS does not care about us passing the setting in, may need to do a regex replacement.

### DIFF
--- a/packages/language-server/src/plugins/typescript/features/CodeActionsProvider.ts
+++ b/packages/language-server/src/plugins/typescript/features/CodeActionsProvider.ts
@@ -109,7 +109,7 @@ export class CodeActionsProviderImpl implements CodeActionsProvider {
                         if (!useSemicolons) {
                             // For some reason the TS language service ignores the formatOptions
                             // we set above, so remove possible semicolons here.
-                            edit.newText = edit.newText.replace(/(;(\n))|(;(\r\n))/g, (match) =>
+                            edit.newText = edit.newText.replace(/(;\n)|(;\r\n)/g, (match) =>
                                 match.substring(1)
                             );
                         }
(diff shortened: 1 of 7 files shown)


dummdidumm:
Ready for review. @jasonlyu123 maybe you have any idea why TS ignores the settings in this case? If not, I would merge this if you think the solution is fine.

packages/language-server/src/plugins/typescript/features/CodeActionsProvider.ts
```diff
@@ -77,12 +80,23 @@ export class CodeActionsProviderImpl implements CodeActionsProvider {
...
+                )
+            ).semi ?? true;
         const changes = lang.organizeImports(
             {
                 fileName: tsDoc.filePath,
                 type: 'file'
             },
-            {},
+            {
+                semicolons: useSemicolons
```
jasonlyu123, line 98: Seems like it's because this config is not boolean but 'ignore', 'insert' and 'remove'. Where 'ignore' and falsy is auto-detect.

dummdidumm: Thank you, that was it! Should've looked at the option more closely 😄

janosh:
@dummdidumm  Thanks for tackling this so quickly! The flashing semicolons are gone. I think the underlying issue remains though. The imports themselves still flash and (more annoyingly) every save adds to the files edit history. If I want to revert a change, I still have to click cmd + z twice. The first cmd + z always undoes the import sorting, even if all imports were already in sorted order.

dummdidumm:
I created #1269 to track this

Issue #854, opened 2021-03-03
[LS] Suggested import returns string with CRLFs, double quotes and no indent


non25:
<details>
<summary>svelteserver screenshots</summary>

![before](https://user-images.githubusercontent.com/27935916/109816074-0263b100-7c39-11eb-8836-c32048b5af97.png)

![after](https://user-images.githubusercontent.com/27935916/109816071-01328400-7c39-11eb-8dfa-62fe4763b886.png)
</details>

Contrary to `tsserver`, which inserts imports without CRLFs and uses quote variant that is prevalent in this file.

<details>
<summary>tsserver screenshots</summary>

![single-quotes-before](https://user-images.githubusercontent.com/27935916/109816353-4e165a80-7c39-11eb-9bae-7d053d10427e.png)

![single-quotes-after](https://user-images.githubusercontent.com/27935916/109816352-4d7dc400-7c39-11eb-9f2d-663dd509b021.png)

![double-quotes-before](https://user-images.githubusercontent.com/27935916/109816348-4ce52d80-7c39-11eb-9d26-9228e1a50fc4.png)

![double-quotes-after](https://user-images.githubusercontent.com/27935916/109816345-4c4c9700-7c39-11eb-9ffa-42f26a63d992.png)
</details>

**Edit**: I've also tried converting Promise-based function to an async in `svelte` file, it inserts CRLFs too, indent is messed up, but at least it doesn't touch my single quotes: :smile: 

I guess vcsode just autoformats everything after code actions so it goes unnoticed?
My autoformatter doesn't strip CRs from LFs. :thinking:

dummdidumm:
These options are passed into the typescript language service by us, and it works in VS Code. You likely need to pass TypeScript's user preferences into the language server [similar to this](https://github.com/sveltejs/language-tools/blob/master/packages/svelte-vscode/src/extension.ts#L92).

jasonlyu123:
With [106.0.0](https://github.com/sveltejs/language-tools/releases/tag/extensions-106.0.0). The behaviour now is:

1. If the current file contains LF but not CRLF, use LF otherwise use the system default. i.e. `require('os').EOL`
2. base indent for the module and instance script tag is now 4 by default. If you have a prettier config it would use the `tabWidth` config. If you don't want the base indent set `svelteIndentScriptAndStyle` to `false` in your prettier config.
3. `quoteStyle` can already be configured with `typescript.preferences.quoteStyle` or `javascript.preferences.quoteStyle`

[...]

PR #1264, opened 2021-12-06
(fix) new line for ts completion and code action

jasonlyu123:
Related to #854. This come up multiple times mainly for coc-svelte folks. Finally found out Typescript default to CRLF if `host.getNewLine` is not implemented and if there's no user config: 

https://github.com/microsoft/TypeScript/blob/97a7901f26290a13ea6c3b057c8720c1614cb221/src/services/utilities.ts#L2352


dummdidumm:
Thanks for finding out!

PR #1276, opened 2021-12-10
(fix) Language Server: ignore refs in generated tsx / non user code

rixo:
A lot of variables references currently end up in generated tsx but are not surrounded by "generated code" guards.

For example, this:

```svelte
<script lang="ts">
  export let foo = false;
  console.log(foo);
</script>
<p>{foo}</p>
```

becomes this:

```tsx
<></>;function render() {

   let foo = false;foo = __sveltets_1_any(foo);;
  console.log(foo);
;
() => (<>
<p>{foo}</p>
</>);
return { props: {foo: foo} as {foo?: typeof foo}, slots: {}, getters: {}, events: {} }}

export default class FindReferencesIgnoreGeneratedTsx__SvelteComponent_ extends __sveltets_1_createSvelte2TsxComponent(__sveltets_1_with_any_event(render())) {
}
```

As a result, the language servers currently emits references for locations (and refs / code) that don't exist in user's source code.

It turns out those ghost references are petty easy to filter out because, as they don't match anything in user's code, they are source mapped back to 0 length ranges.

I believe it is always safe to ignore zero-length references since I don't see how any IDE feature could benefit from them. You can't refactor/rename a zero-length text. On the other hand, references navigation plugins are harmed by those incorrect refs, since the references don't actually exist in user code.

This PR adds a test that illustrates the problem, and implements filtering of zero-length references to make it pass.

### DIFF
--- a/packages/language-server/src/plugins/typescript/features/FindReferencesProvider.ts
+++ b/packages/language-server/src/plugins/typescript/features/FindReferencesProvider.ts
@@ -4,7 +4,7 @@ import { Document } from '../../../lib/documents';
 import { pathToUrl } from '../../../utils';
 import { FindReferencesProvider } from '../../interfaces';
 import { LSAndTSDocResolver } from '../LSAndTSDocResolver';
-import { convertToLocationRange } from '../utils';
+import { convertToLocationRange, hasNonZeroRange } from '../utils';
 import { isNoTextSpanInGeneratedCode, SnapshotFragmentMap } from './utils';
 
 export class FindReferencesProviderImpl implements FindReferencesProvider {
@@ -29,7 +29,7 @@ export class FindReferencesProviderImpl implements FindReferencesProvider {
         const docs = new SnapshotFragmentMap(this.lsAndTsDocResolver);
         docs.set(tsDoc.filePath, { fragment, snapshot: tsDoc });
 
-        return await Promise.all(
+        const locations = await Promise.all(
             references
                 .filter((ref) => context.includeDeclaration || !ref.isDefinition)
                 .filter(notInGeneratedCode(tsDoc.getFullText()))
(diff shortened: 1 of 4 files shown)


dummdidumm:
Makes sense to me. For references this should indeed be the right behavior. For something like diagnostics it wouldn't because this would swallow some errors that are correct, for example "this action is not assignable to this kind of element". This is unrelated to this PR as far as I can see but I just wanted to bring this up to also remind my future self of this.

packages/language-server/src/plugins/typescript/features/FindReferencesProvider.ts
```diff
@@ -42,6 +42,10 @@ export class FindReferencesProviderImpl implements FindReferencesProvider {
                     );
                 })
         );
+
+        const nonZeroLocations = locations.filter(hasNonZeroRange);
+
+        return nonZeroLocations.length > 0 ? nonZeroLocations : locations;
```
dummdidumm, line 48: What's the reason for returning locations in case there's no non-zero range found?

rixo: A bias toward preserving existing behaviour in the face of uncertainty, I suppose. But it might be misguided in this case...

The case I had vaguely in mind was if it returned references in another file, it might still be somewhat useful to jump to this file, even if no actual precise location can be found. That said, even then, it may do more harm than good, maybe giving the impression that the language tools is a bit lost, jumping to the middle of a file for no obvious reason. The behaviour would make much more sense for definitions and I think I got a bit confused.

I'm removing that.

packages/language-server/src/plugins/typescript/features/FindReferencesProvider.ts
```diff
@@ -42,6 +42,8 @@ export class FindReferencesProviderImpl implements FindReferencesProvider {
                     );
                 })
         );
+
+        return locations.filter(hasNonZeroRange);
```
dummdidumm, line 46:
```suggestion
        // Some references are in generated code but not wrapped with explicit ignore comments.
        // These show up as zero-length ranges, so filter them out.
        return locations.filter(hasNonZeroRange);
```

dummdidumm:
Thanks!

Issue #139, opened 2020-05-30
Please can you optimize the size of the VSCode extension


mgholam:
**Is your feature request related to a problem? Please describe.**
Please can you optimize the on disk size of the VSCode extension (i.e. smaller than 150mb)

**Describe the solution you'd like**
By possibly pre-building the extension before publishing.

dummdidumm:
@orta can we do anything about this? I'm not sure how the publish-to-marketplace stuff works and how we could add a bundle/minify step before that.

orta:
The official recommendations are here: https://code.visualstudio.com/api/working-with-extensions/bundling-extension

Basically it's 'move to webpack'

dummdidumm:
Tried with a rollup config, but no success.

So far I have:
- One rollup config for the language-server. It builds the package into a single file
- One rollup config for svelte-vscode. It build the extension into a single file
- A deploy script in svelte-vscode which takes care of executing both these rollup builds and then replacing the contents of `dist` in `node_modules/svelte-language-server` in svelte-vscode with the built output. Reason: That way the `require.resolve(...)` in `extension.ts` will still work
- Added ignore for all node_module content except that of `svelte-language-server`. Since `svelte-language-server/dist` contains everything it needs, other modules are not relevant anymore. They still need to be present during deployment though so `vsce` does not throw errors.

Yes, this all feels very hacky 😄 

Problems:
- Inside the language server, there is some dynamic `require.resolve` stuff. This does not work well with the current rollup config, as, well, those are all bundled now and not found anymore like that. Don't know how to solve this.

mgholam:
I did this [https://github.com/mgholam/svelte-multi-page-app](https://github.com/mgholam/svelte-multi-page-app) with one `rollup.config.js` if it helps.

dummdidumm:
Down to less than 90mb now. Since we do some dynamic requires/depending on node_modules resolution at runtime and that is very tricky in combination with bundling, I will not explore bundling further for now.

mgholam:
What kind of numbers are we looking at if all goes well (currently at 80mb)?

The biggest folder is "typescript" at 54mb right now.

dummdidumm:
Honestly I don't know, but with bundling and minification I assume it could go down to less than 10mb.

mgholam:
Can you give me a starting point, so I can have a crack at this?

dummdidumm:
[...]

PR #1278, opened 2021-12-14
vscodeIgnore some unused files to optimize extension size

jasonlyu123:
#139 

- Ignore some different purpose builds for typescript. We didn't use them.
- Ignore typescript localization. We didn't enable localized diagnostic.
- Ignore large ESM files, We're currently using common js

Tested and it brings the size down from 105MB to 42MB. With this file size reduction, it might make sense to ship some commonly used preprocess libraries to improve UX. 
About typescript localization, it's possible to enable this. So if we want to enable it in the future, I think we can keep these files. 

To test packaging this, run these commands:
```sh
rm package.json yarn.lock
cd packages/svelte-vscode
yarn install
vsce package
```

And install the vsix file or unzip it to see the folder structure.

### DIFF
--- a/packages/svelte-vscode/.vscodeignore
+++ b/packages/svelte-vscode/.vscodeignore
@@ -2,3 +2,17 @@
 /src
 /scripts
 /.vscode
+
+node_modules/@types
+node_modules/typescript/loc
+node_modules/typescript/lib/*/*
+node_modules/typescript/lib/tsc.js
+node_modules/typescript/lib/tsserver.js
+node_modules/typescript/lib/tsserverlibrary.js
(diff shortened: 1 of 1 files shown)


dummdidumm:
This is great, thank you! If we want to enable localization later on we can always un-ignore those files again.
About adding possible other preprocessors: I'm a little bit hesitant to do this. Right now less and sass seem like a good addition, but with the rise of Tailwind people might expect something there, too, and over time it could become too bloated. I'd like to keep this to TypeScript-only for now. Maybe we should instead switch how we deal with these diagnostic errors and ignore errors that come from missing modules if there's no `svelte.config.js`? -> we should probably discuss this in an issue 😄

PR #1279, opened 2021-12-14
(test) add support for .only to diagnostics tests

rixo:
This change allows suffixing a fixture directory makes mocha run only this test, which is useful during dev/debugging. Maybe we should use `.solo` like the Svelte compiler?


rixo:
Thinking of it, if we take this in, we should probably also add support for `.skip` suffix, that would allow for "pending" tests.

dummdidumm (approved):
great addition, thanks! Feel free to send another PR for skip and for support `.solo` as an alternative (but please not as a replacement of `.only` 😄 )

PR #1282, opened 2021-12-15
(feat) Typescript go to implemenation

jasonlyu123:
We have had this in the ts-plugin for some time. Adding this to the svelte files now. In a svelte file, it's probably mainly useful when you have some DTO in object literal form.


dummdidumm (approved):
Awesome, thanks!

PR #1286, opened 2021-12-15
(fix) Silence diagnostics for no side effects variables in reactive blocks -- fixes #1277

### LINKED ISSUE #1277: Reactive expression hinting `$: x, fn()` raises TS error
**Describe the bug**
Reactive expression hinting raises TS error "Left side of comma operator is unused and has no side effects".

```js
$: x, update()
```

AFAIK this is very idiomatic Svelte to mean "update when x changes", so this shouldn't be marked as error/warning.

**To Reproduce**
Paste the following in a `.svelte` file in VSCode:

```svelte
<script lang="ts">
  let x: any

  const update = () => {}

  $: x, update()
</script>
```

`x` in the reactive expression is marked as error.

**Expected behavior**
There should be no TS errors for the above snippet.

 - OS: Ubuntu
 - IDE: VSCode
 - Plugin/Package: Svelte for VSCode v105.7.0

rixo:
For diagnostics "Left side of comma operator is unused and has no side effects.", this code analyzes the AST to determine whether the errors point to a `let` variable in a reactive statement. In this case, we assume this is purposeful reactive block hinting and suppress the error. New diagnostics are emitted for each non `let` identifiers that exists in the span of the original diagnostic.

This strategy resulted in duplicated diagnostics in some case, which doesn't seem useful, so I also added a diagnostic deduplication stage to the diagnostics processing pipeline.

To fix the main case, I had to introduce an AST based `isReactiveStatement` util function and also noticed the suppression of the "Unused label." warnings for `$:` of reactive statements used a weak heuristic. So I switched the warning suppression strategy to using the AST based function too. This is a low key enhancement since no one is using labels in JS, but it's still more solid, and almost free if we merge this, since the `isReactiveStatement` function will have to be maintained for the main purpose of the PR (and surely, people are using reactive statements hinting). I added a test that demonstrates a false positive with the current implementation and is fixed by the change, and also some more to increase coverage of the target case.


packages/language-server/src/utils.ts
```diff
@@ -43,8 +57,21 @@ export function getLastPartOfPath(path: string): string {
     return path.replace(/\\/g, '/').split('/').pop() || '';
 }
 
-export function flatten<T>(arr: T[][]): T[] {
-    return arr.reduce((all, item) => [...all, ...item], []);
+export function flatten<T>(arr: (T | T[])[]): T[] {
```
rixo, line 60: Expanded to make the behaviour closer to native JS `flatMap`: flatten arrays, keep non array values as is. The adaptation was needed to spare the client code in the function I introduced to do this dance itself.

Existing tests say this is a backward compatible change :innocent:

packages/language-server/src/plugins/typescript/features/DiagnosticsProvider.ts
```diff
@@ -5,8 +5,22 @@ import { DiagnosticsProvider } from '../../interfaces';
...
+    gatherIdentifiers
+} from './utils';
+import { not, flatten, passMap, regexIndexOf, swapRangeStartEndIfNecessary } from '../../../utils';
+
+enum DiagnosticCode {
+    // "Left side of comma operator is unused and has no side effects."
+    DIAGNOSTIC_CODE_NOOP_IN_COMMAS = 2695,
+    // "Unused label."
+    DIAGNOSTIC_CODE_UNUSED_LABEL = 7028
+}
```
dummdidumm, line 23: Really like the enum idea! Two suggestions:
1. remove the `DIAGNOSTIC_CODE_` prefix, that's already apparant from using the enum
2. Could you add the other diagnostic codes that are used in this file, too?

packages/language-server/src/plugins/typescript/features/DiagnosticsProvider.ts
```diff
@@ -61,7 +80,8 @@ export class DiagnosticsProviderImpl implements DiagnosticsProvider {
             .filter(hasNoNegativeLines)
             .filter(isNoFalsePositive(document, tsDoc))
             .map(enhanceIfNecessary)
-            .map(swapDiagRangeStartEndIfNecessary);
+            .map(swapDiagRangeStartEndIfNecessary)
+            .filter(dedupDiagnostics());
```
dummdidumm, line 84: You said that some of the work results in duplicate diagnostics. If I understood correctly, only the diagnostics related to reactive statements are duplicated. If so, would it make more sense to add this filter only to `resolveNoopsInReactiveStatements` instead?

dummdidumm (changes requested):
Thanks for the PR! AST traversal is a good idea here, and thanks for cleaning that up for the unused label diagnostic, too. I added some suggestions inline. Also I would prefer if you transformed all short-hand `if (...) return` / `if (...) return x` into
```
if (...) {
  ...
}
```  
Maybe there's an ESLint rule for that, guess I have to find out.
There are also some other Lint errors which need fixing.

packages/language-server/src/plugins/typescript/features/DiagnosticsProvider.ts
rixo: Sure.

packages/language-server/src/plugins/typescript/features/utils.ts
```diff
@@ -157,3 +158,103 @@ export function findContainingNode<T extends ts.Node>(
...
+            if (!next) return false;
+            const current = next;
+            next = next.parent;
+            return predicate(current);
+        });
+    };
+}
+
+const isRenderFunction = isLineage<ts.FunctionDeclaration & { name: ts.Identifier }>(
+    (node) => ts.isFunctionDeclaration(node) && node?.name?.getText() === 'render',
```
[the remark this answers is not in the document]
rixo: This is what it does. Maybe `isLineage` needs a better name, but I already struggled a lot to settle on this, so suggestions are welcome. Maybe it should be named `isNodeAndEachParents`? Would like something less verbose if possible.

`isLineage` takes a list of predicates, run the first one on the node, the next one on the first parent, the third on the next parent, and so on... and it returns `true` only if all predicates return `true`.

So line 224 does ensure that the immediate parent is `SourceFile`.

packages/language-server/src/plugins/typescript/features/DiagnosticsProvider.ts
rixo: - It is easier to filter after the conversion from ts.Diagnostic to VSCode Diagnostic, because the latter has a smaller API.
- By testing VSC Diagnostic (we could push to test all props of the type, not just those that are used in the conversion here), we're almost certain that a detected dup is absolutely the same thing. I considered exact dups were never desirable, so might as well put it at the end and be everything proof.

That said, I have no strong opinion on this. I can probably put something together to filter them out just of  `resolveNoopsInReactiveStatements`.

Your call?

packages/language-server/src/plugins/typescript/features/utils.ts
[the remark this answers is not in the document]
dummdidumm: Oh I see, thanks. Yeah the name is indeed a hard one to get right. `nodeAndParentsSatisfyRespectivePredicates`? I know it's verbose but in these cases I prefer verbosity over possible confusion.

packages/language-server/src/plugins/typescript/features/DiagnosticsProvider.ts
dummdidumm: I prefer to have it inside `resolveNoopsInReactiveStatements` for two reasons:
1. Performance - probably doesn't really make a difference but that way only the ones that we know can be duplicates are filtered out
2. Colocation - reading this could I would assume that there are no duplicates because I would expect TS to know what it's doing and not create duplicate diagnostics on its own. Right now the new function is the only one creating new diagnostics out of existing ones so that is the only source for duplicates, so I prefer to have handled that there. Else someone could think "mhm this looks unnecessary" later on and remove it.

rixo:
@dummdidumm Thanks.

I have addressed the obvious items in your feedback. Waiting your opinion before moving on the dedup question.

Oof! We've got a stylistic divergence about single line `if (...) return :sweat_smile: I am a big fan of using those as logic guards, to get rid of non targeted conditions as soon as possible, and offload them from brain and indentation level. And I like to separate them semantically as much as possible, so this can get hairy fast if expanding all if blocks...

```js
if (a) {
  return
}
if (b) {
  return
}
if (c) {
  return
}
```

...

Anyway, sure! No problem adhering to the style of the project even if it's not my favorite. Would be great if we had an eslint autofix rule for them, to catch 'em all back now :sweat_smile: 

And yeah, I did notice that I had linting issues after pushing. Sorry, I meant to mention that. There seems to be something off with my setup (or something I don't get)... For example, if I run the following from the root of the repo, I don't get any error / warning (just a warning that I'm using an unsupported version of TS (?)):

```bash
yarn eslint packages/language-server/src/plugins/typescript/features/DiagnosticsProvider.ts
```

I need to try and figure out what differs between CI and my local setup, to spare relying on CI to know if I have linting issues. Do you have any idea what I can do to fix that?

packages/language-server/src/plugins/typescript/features/DiagnosticsProvider.ts
rixo: OK I'm changing it.

dummdidumm:
> Oof! We've got a stylistic divergence about single line `if (...) return 😅 I am a big fan of using those as logic guards, to get rid of non targeted conditions as soon as possible, and offload them from brain and indentation level. And I like to separate them semantically as much as possible, so this can get hairy fast if expanding all if blocks...

I like this style aswell, I just don't like these bracket-less control blocks, especially if you later want to add something to it, or the line gets too long and then it's broken into two lines by prettier and I definetly don't want
```
if (something)
  return;
```

> I need to try and figure out what differs between CI and my local setup, to spare relying on CI to know if I have linting issues. Do you have any idea what I can do to fix that?

Honestly don't know, and I honestly don't mind if you "abuse" the CI to fix your lint errors 😄

dummdidumm:
I need parts of this for my svelte2tsx rewrite (and also to not have conflicts all other the place) so I'm going to add the final changes myself and then merge.

dummdidumm:
Thank you very much for the fix and the AST enhancements!

Issue #1297, opened 2021-12-26
Accept boolean as a value for radio input bind groups


lovasoa:
### Describe the problem

Currently, the following code:

```svelte
<script lang="ts">
    export let firstSelected = true;
</script>

<input type="radio" bind:group={firstSelected} value={true}/>
<input type="radio" bind:group={firstSelected} value={false}/>
```

Raises the following type error:

```
Error: Type 'true' is not assignable to type 'string | number | string[]'. (ts)
Error: Type 'false' is not assignable to type 'string | number | string[]'. (ts)
```

(see https://stackoverflow.com/questions/65841762/bind-radio-group-to-a-boolean-value-using-svelte )

### Describe the proposed solution

boolean could be accepted as a radio input value

### Alternatives considered

A working alternative is to just let the user code their own two-way binding, but that is not very clean, and may be error-prone:

```svelte
<script lang="ts">
    export let firstSelected = true;
    
    $: _firstSelected = Number(firstSelected);
    
    function handleChange(e) {
        firstSelected = Boolean(_firstSelected);
    }
</script>

<input type="radio" bind:group={_firstSelected} value={1} on:change={handleChange}/>
<input type="radio" bind:group={_firstSelected} value={0} on:change={handleChange}/>
```

### Importance

would make my life easier

lovasoa:
:tada: thank you very much, @dummdidumm !

PR #1309, opened 2022-01-03
(fix) relax value typings

dummdidumm:
A radio group value can be anything, so we need to relax the type
#1297


dummdidumm:
@jasonlyu123 not 100% sure on this. What do you think about relaxing the type here? I don't see another way.

jasonlyu123:
I also don't see another way. From the compiled code it seems like the compiler just stores the value in the element instance. It really can be anything.

PR #1315, opened 2022-01-05
Special highlighting for variable props

Monkatraz:
Was part of my old PR. Quoting myself:
> In detail, normal properties use the `entity.other.attribute-name.svelte` scope, while CSS variables use the `support.type.property-name.svelte` scope. These tend to have slightly different syntax highlighting in most themes, but the way they are highlighted is consistent with CSS properties. I figure that it should be clear - even in highlighting - that these are not really attributes in the normal sense.
> 
> Monokai Dimmed shows this well (My theme, Dark+, and Light+ all highlight those two scopes the same way)
> image

### DIFF
--- a/packages/svelte-vscode/syntaxes/svelte.tmLanguage.src.yaml
+++ b/packages/svelte-vscode/syntaxes/svelte.tmLanguage.src.yaml
@@ -339,24 +339,36 @@ repository:
   # Matches attribute keyvalues. (and boolean attributes as well)
   # e.g. `class="my-class"`
   attributes-keyvalue:
-    begin: ([_$[:alpha:]][_\-$[:alnum:]]*)
-    beginCaptures: { 1: { name: entity.other.attribute-name.svelte } }
+    begin: ((?:--)?[_$[:alpha:]][_\-$[:alnum:]]*)
+    beginCaptures:
+      0:
+        patterns:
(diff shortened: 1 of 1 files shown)


packages/svelte-vscode/syntaxes/svelte.tmLanguage.src.yaml
```diff
@@ -339,17 +339,26 @@ repository:
...
+    patterns:
+    - begin: '='
+      beginCaptures: { 0: { name: punctuation.separator.key-value.svelte } }
+      end: (?<=[^\s=])(?!\s*=)|(?=/?>)
+      patterns: [include: '#attributes-value']
 
-  # The value part of attribute keyvalues. e.g. `="my-class"` in `class="my-class"`
+  # The value part of attribute keyvalues. e.g. `"my-class"` in `class="my-class"`
   attributes-value:
-    begin: '='
```
dummdidumm, line 350: You removed this and instead placed it at the two sites where `attributes-value` is included. Could you explain why?

Monkatraz: Honestly, I think these changes are for the style strings PR but I did this so long ago I mixed up what was necessary. I think it makes more sense like this anyways? Makes it easier to have different types of attribute value parsing

PR #1316, opened 2022-01-05
Highlight units/digits in quoted attribute values

Monkatraz:
Was part of my old PR. This extends the attribute highlighting to handle quoted units/numbers in strings. It will only highlight if the entire string is matched. This behavior already exists for unquoted attribute values.

image

### DIFF
--- a/packages/svelte-vscode/syntaxes/svelte.tmLanguage.src.yaml
+++ b/packages/svelte-vscode/syntaxes/svelte.tmLanguage.src.yaml
@@ -354,9 +354,12 @@ repository:
     # No quotes - just an interpolation expression.
     - include: '#interpolation'
     # Units, meaning digit characters and an optional unit string. e.g. `15px`
-    - match: ([0-9._]+[\w]{,4})(?=\s|/?>)
-      name: constant.numeric.decimal.svelte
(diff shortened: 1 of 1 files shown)


packages/svelte-vscode/syntaxes/svelte.tmLanguage.src.yaml
```diff
@@ -354,9 +354,12 @@ repository:
     # No quotes - just an interpolation expression.
     - include: '#interpolation'
     # Units, meaning digit characters and an optional unit string. e.g. `15px`
-    - match: ([0-9._]+[\w]{,4})(?=\s|/?>)
-      name: constant.numeric.decimal.svelte
-      patterns: [ include: '#interpolation' ]
+    - match: (?:(['"])([0-9._]+[\w%]{,4})(\1))|(?:([0-9._]+[\w%]{,4})(?=\s|/?>))
```
dummdidumm, line 357: The new match is a non-capturing group. Could you explain why?

Monkatraz: It's to more cleanly group the two alternatives in the pattern. Each of the alternatives is wrapped in a non-capturing group.

Issue #1310, opened 2022-01-04
updateImports deletes comments between imports


janosh:
**Describe the bug**

With VS Code organize imports on save enabled,

```json
  "[svelte]": {
    "editor.defaultFormatter": "svelte.svelte-vscode",
    "editor.codeActionsOnSave": ["source.fixAll", "source.organizeImports"],
  },
```

saving the following file deletes the comment line between imports.

**Before**

```svelte
<script>
  import A from './A.svelte'
  // foo
  import B from './B.svelte'
</script>

<A />
<B />
```

**After**

```svelte
<script>
  import A from './A.svelte'
  import B from './B.svelte'
</script>

<A />
<B />
```

**Expected behavior**
Comment should stay as is.

janosh:
@dummdidumm Thanks! Awesome work you and all the other contributors are doing on the Svelte language tools. 👍

PR #1320, opened 2022-01-06
(fix) prevent comment being removed in organizes imports

jasonlyu123:
#1310 

Side effort, this also allows the `@ts-ignore` comment to silence import diagnostics.


dummdidumm (approved):
👍

Issue #1139, opened 2021-08-16
5 second delayed intellisense with Svelte Preprocess


PuruVJ:
**Describe the bug**
I have a 23 components big app, with TypeScript and SCSS. The intellisense for just about **anything**(Svelte related or JS related) inside `.svelte` files takes around 5-6 seconds to show up. In the very beginning when there was just 1 component, it was instantaneous.

This isn't a problem with my computer, because I get instantaneous intellisense in .js, .ts, .tsx file even with 100+ files in the project.

In this svelte app, the intellisense in .ts files is still instantaneous, so it's an issue with the Svelte extension only.

**To Reproduce**
1. Clone this repo: https://github.com/puruvj/macos-web
2. Run `npm install`
3. Go to any Svelte file with some local state and try typing out variable names in the `<script>` tag or in the template or the snippets autocompletion of `if`, `each`, `await` blocks, etc. See the speed of intellisense
4. Now, delete `node_modules`, go to the same `.svelte` file as before, and now try intellisense. It will be instantaneous now.

I think this issue is happening because of `svelte-preprocess`, because I have noticed this lag in every other Svelte project with TS/SCSS with svelte-preprocess, whereas projects without it are really fast.

**Expected behavior**
I want the intellisense to be as instantaneous with svelte-preprocess as it is is without it 🙂

**System (please complete the following information):**
 - OS: Windows
 - IDE: VSCode
 - Plugin/Package: Official Svelte extension for VSCode

dummdidumm:
The intellisense gets slower for me as well, which is expected since there are now much more files the TypeScript language service has to scan. I don't see the massive slowdown you see however. Could you give more details?
- which file exactly should I open, what exactly should I type to see it?
- what exactly to you mean by "intellisense"? Auto completion suggestions?
- Which version of the extension do you have installed
- What's your hardware spec (roughly)?

PuruVJ:
Sure!

I noticed an additional thing that when I open the workspace in VSCode and open a .svelte file, the lag isn't that bad. It's like 1-2 seconds, which is manageable, but as I open more files and work on them, then it starts getting slower. Closing the files doesn't make it any faster.

And here are the details for your questions:

dummdidumm:
I tried to narrow this down, but no luck. The biggest delay I get is within a function of the TypeScript language service which we have no control over. I'm not sure if this is what slows this down for you. What you could try to narrow this down:

Add this to your user settings (JSON):
```
[...]

Issue #1130, opened 2021-08-04
Extension slow to respond


brunnerh:
**Describe the bug**
Getting auto-completion, symbol navigation or symbol information can be very slow (tooltip stuck on `Loading...`).
(It is probably the language server, because the profile from "Running Extensions" does not pick it up as causing CPU load.)

**To Reproduce**
Invoke commands like "Trigger Suggest" and "Go to Definition".

As for the conditions necessary to cause the slowdown, i am not sure. It may have to do with using `carbon-components-svelte`/`carbon-icons-svelte` which introduce a lot of Svelte files.
The "Output" panel shows this for Svelte:
> SnapshotManager File Statistics:  
> Project files: 26022  
> Svelte files: 7210  
> From node_modules: 33251  
> Total: 33276  

**Expected behavior**
Language features are reasonably fast.

**Screenshots**
![image](https://user-images.githubusercontent.com/834235/128189452-4a5cbbd7-f6c8-4add-9ffa-bbda921e6486.png)

**System (please complete the following information):**
 - OS: Windows 10, 64bit
 - IDE: VSCode
 - Plugin/Package: Svelte for VSCode

jasonlyu123:
I did encounter a similar issue before. Still can't found out why, but it seems to be slowing down in the typescript language service part. It only pops up occasionally for me so this is not easy to debug. 
Some other context might help. Do you use pnpm? There's seem to be some pnpm issues in the typescript repo. And what version of svelte for vscode you're using? if it's 105.3.3, does roll back to 105.3.1 helps? Or does turning off `typescript.suggest.includeCompletionsForImportStatements` and `javascript.suggest.includeCompletionsForImportStatements` helps?

stolinski:
We're having major slowdown and often get a pop up saying that VSCode is waiting of the Svelte formatter.

Some details
v. 105.3.3
Svelte Kit App
TS for js
Scss for css 
NOT using pnpm, just npm.

jasonlyu123:
I publish a new version(105.3.4). Hope this would address the performance issues you're facing.

brunnerh:
That might have fixed it. Right now the file statistics do not grow as much as previously:

```
SnapshotManager File Statistics:
Project files: 21
Svelte files: 6
From node_modules: 382
Total: 404
```

brunnerh:
I will have to revoke that statement.
[...]

PR #1325, opened 2022-01-07
(perf) cache fileExists resolutions

dummdidumm:
This makes getCompletions perform many times faster, the bigger the project the more it benefits from this cache.

I noticed that there's a lot of time spent in `fileExists` during analyzing whether or not #1192 does make any difference. I then also found that TypeScript does something similar for its watcher at https://github.com/microsoft/TypeScript/blob/1298f498f4cde7fa53cbf050ebfeefdf85c848ec/src/compiler/watchUtilities.ts#L42 , and that `ts.sys.fileExists` does not cache anything, so we need to do it ourselves.

Should help #1130 and #1139

I also found out during that that we are likely not properly removing Svelte snapshots when its file is deleted. I'll try this out later on.

Another possible speed-up I found was [this method in completions](https://github.com/sveltejs/language-tools/blob/master/packages/language-server/src/plugins/typescript/features/CompletionProvider.ts#L269), according to the perf analysis some time is spent in that area in total, probably due to source mapping. We could speed that up in this case because the textEdit ranges are probably all the same, so we can create a cache there. Update: I was wrong, textEdit is rarely filled especially when it's about global completions. But I found that we can compute `path` once only because `pathToUrl` is not that fast either.

@jasonlyu123 thoughts?

### DIFF
--- a/packages/language-server/src/lib/documents/Document.ts
+++ b/packages/language-server/src/lib/documents/Document.ts
@@ -17,6 +17,11 @@ export class Document extends WritableDocument {
     configPromise: Promise<SvelteConfig | undefined>;
     config?: SvelteConfig;
     html!: HTMLDocument;
+    /**
+     * Compute and cache directly because of performance reasons
+     * and it will be called anyway.
+     */
+    private path = urlToPath(this.url);
 
     constructor(public url: string, public content: string) {
         super();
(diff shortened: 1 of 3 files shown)


jasonlyu123:
Looks good. The only downside I could think of it's where files from outside of the workspace or inside node_module got created or deleted. But it has also been cached in module resolution cache anyway.

PR #1327, opened 2022-01-08
Add syntax highlighting support for `style:foo` props

Monkatraz:
Does what the title says.

The following patterns are accounted for:
```
style:foo
style:foo=
style:--var    # highlighted as a JS variable, but who knows what this would actually mean
style:--var=
```


dummdidumm:
That was fast 😄 thanks!

PR #1346, opened 2022-01-27
Expose svelte2tsx component documentation

Tymek:
No currently available tool is able to generate **complete** component documentation from svelte file. There are many attempts listed here https://github.com/sveltejs/kit/issues/1979#issuecomment-1014259443. With this change it will be possible to access documentation of component and use `svelte2tsx` as a _docgen_ tool.

### DIFF
--- a/packages/svelte2tsx/index.d.ts
+++ b/packages/svelte2tsx/index.d.ts
@@ -3,11 +3,24 @@ export interface SvelteCompiledToTsx {
     map: import("magic-string").SourceMap;
     exportedNames: IExportedNames;
     events: ComponentEvents;
+    componentDocumentation: string;
+    slots: Map<string, Map<string, string>>;
+    generics: {
+        definitions: string;
+        references: string;
+    };
 }
 
-export interface IExportedNames {
-    has(name: string): boolean;
-}
+export type IExportedNames = Map<
+    string,
+    {
+        isLet: boolean;
+        type?: string;
+        identifierText?: string;
+        required?: boolean;
+        doc?: string;
+    }
+>;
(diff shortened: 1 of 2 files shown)


packages/svelte2tsx/index.d.ts
```diff
@@ -3,11 +3,24 @@ export interface SvelteCompiledToTsx {
     map: import("magic-string").SourceMap;
     exportedNames: IExportedNames;
     events: ComponentEvents;
+    componentDocumentation: string;
+    slots: Map<string, Map<string, string>>;
+    generics: {
+        definitions: string[];
```
dummdidumm, line 9: The types say these are strings, but that's not true from the implementation (besides it's `definition` in the implementation). In the implementation it's just one string fitting the use case which is producing correct code, not something for use later.

dummdidumm (commented):
Could you explain what's the benefit of exposing additional properties here, and what format would be most fitting?

Tymek:
Most fitting format will probably be similar if not identical to what sveld is using. [ParsedComponent interface in sveld/src/ComponentParser.ts#L98-L106](https://github.com/carbon-design-system/sveld/blob/1cfd742dcb0c3847d001c6e6bba0e2f088c7a86a/src/ComponentParser.ts#L98-L106)

With this change using other parsers like sveld will not be necessary to output slots, generics etc.

I get it if this use case is better left for another package.

mrtnbroder:
@Tymek I think it would also be great to add examples on how to use it. Not neccessarily needed in this PR, but for the general public it still looks quiet abstract.

PR #1348, opened 2022-01-28
Update yarn instruction for yarn 2

BeLeap:
Command have been changed https://yarnpkg.com/getting-started/editor-sdks#:~:text=yarn%20dlx%20%40yarnpkg/sdks%20vscode


dummdidumm:
Is this a backwards-compatible change they did? If not, we should also mention the old command.

BeLeap:
@dummdidumm It does not work at all.
image

dummdidumm:
Oh I see now, this is a new/separate package, so this will work for everyone as the corresponding package is installed if not present.

PR #1354, opened 2022-01-30
add basic types for svelte:component

ivanhofer:
This PR adds basic types for the `<svelte:component />` element

Currently you won't get a TypeScript error if you forget to set the `this` prop.
With this PR an error is shown if `this` is missing.

I could not find out what the type of a real SvelteComponent is so I just used the `object` type which represents the component as closest without adding an additional type.


dummdidumm:
The new transformation works differently, it instantiates the component directly, so adding `sveltecomponent` to the types does nothing.
Also I'm not sure about the "no this error if you forget it" - the Svelte parser throws an error if it's missing, and the intellisense surfaces this error as a TS error, too.

Issue #1352, opened 2022-01-29
useNewTransformation Feedback Thread


dummdidumm:
We introduced a new transformation which powers the Svelte intellisense under the hood. The goal is to get rid of a tsx-style transformation and simplify transformations along the way.

Advantages for users:

- no messing with projects who use JSX for some other reason and then have conflicting JSX definitions (#1256, #1149)
- better type inference in some cases, for example around svelte:component

Internal advantages:

- TS control flow easier to keep flowing
- overall easier transformations of Svelte specific syntax like await, each etc.

You can try out the new transformation using 
- "Svelte for VS Code" version 105.11.1 and setting `svelte.plugin.svelte.useNewTransformation` to `true`
- `svelte-check` version 2.4.0 with the `--use-new-transformation` flag.

Please provide feedback in here about any improvements or regressions.

dummdidumm:
Note for myself: We need to make the "no type definition found" or "type definition is wrong" error which we enhance for the old transformation [here](https://github.com/sveltejs/language-tools/blob/master/packages/language-server/src/plugins/typescript/features/DiagnosticsProvider.ts#L269) better for the new transformation and possibly also wrap the `new Component` call in a function which falls back to some defaults - don't know yet.

TomTomB:
Hi, so far I've found 3 issues when using the new transformation. 

**Data attributes**
Using a data attribute (e.g. data-popper-arrow for popper.js) results in the following error 
```
Argument of type '{ class: string; "data-popper-arrow": boolean; }' is not assignable to parameter of type 'HTMLProps<HTMLDivElement>'.
  Object literal may only specify known properties, and '"data-popper-arrow"' does not exist in type 'HTMLProps<HTMLDivElement>'
```

**Some attributes got renamed**
This is a kind of breaking one. Before you could use something like `xlink:href="someId"` inside a svg. now it needs to be `xlinkHref="someId"`. The same applys for `xmlns:xlink` and  `tabIndex` (now `tabindex`).

**Typing custom events from actions is a total loss**
Before, there was a workaround to type custom events dispatched via `use:someAction`. 


```typescript
// custom-events.d.ts
declare namespace svelte.JSX {
 interface HTMLAttributes<T> {
    'onclick-outside'?: (
      event: CustomEvent & {
        target: EventTarget & T;
      },
    ) => void;
  }
}
```

```svelte
svelte component

<script>
  import { clickOutside } from 'somewhere';

  const onClickOutside = () => {
[...]

PR #1369, opened 2022-02-04
(fix) no TS/JS completions in html text

dummdidumm:
#1352
This makes use of the Svelte AST that is now returned from svelte2tsx (privately, do not use this elsewhere!) by checking if the AST node at the given position is of type text. This also helped refactor some other checks into something less hacky

@jasonlyu123 what do think of this approach, using the Svelte AST more in the language server?


jasonlyu123:
Seem like a good idea. We probably can't use the AST from vscode-html-languageservice here as most of the svelte block and moustache is treated as text.

PR #1373, opened 2022-02-06
Add snippets

### LINKED ISSUE #1372: Wrap with snippet support
I sometimes write something out and then later have to put it inside an if check, currently I put the starting tag for the if block. Navigate to the end of what I want to wrap and insert the ending tag.

VSCode 1.64 introduced `surround with snippet` support for JS and TS [[link](https://code.visualstudio.com/updates/v1_64#_surround-with-snippets-for-jsts)]. It'd be nice to have something similar for svelte.

**Additional context**
Currently the `Insert snippet` command only lists custom snippets by the user, the snippets from the extension don't show up there.

gtm-nayan:
Added a few common snippets which also support wrapping existing code. 
#1372 

`{: }` parts have been omitted to reduce duplication, the language server provides those as autocompletions.


jasonlyu123:
Wondering if this can have a different prefix. As it would also show up with `{#` duplicating with existing completion. Also, accepting the snippet would result in duplicated `{` like this:
```svelte
{{#if condition}
    
{/if}}
```

gtm-nayan:
For accepting the snippet creating duplicate brackets, removing them from the snippet will cause
```svelte
#if condition}
    wrapped content
{/if
```
when used for wrapping.

A different prefix sounds good, what should it be?

dummdidumm:
I wonder if we somehow can dedupcliate this. Options:
- Remove the autocompletion of Svelte tags and only rely on snippets. Advantage: Easy deduplication. Disadvantage: You won't get completions for continuing completions (you type `{#await` first then add another `{:then` block which you get autocompletion for)
- Add some kind of smart detection that shuts off autocompletion when it detects that this is the same as a prefix of a snippet. This won't handle the duplicate `{` insertion though.
- Smarter snippets? I haven't dug into how snippets work very much, but maybe they can be fine-tuned to not show up in specific instances (like: "Only show up when you are not prefixed with a `{`" or "Only show up in the surround with snippet case")

gtm-nayan:
The first one is probably off the table because the snippets would only apply for svelte-vscode. Other editors and plugins that rely on the LSP would not be happy.


> "Only show up in the surround with snippet case"

Turns out [it's possible](https://code.visualstudio.com/docs/editor/userdefinedsnippets#_can-i-remove-snippets-from-intellisense) for users to do it in their editor, but I couldn't find anything related to doing it in the snippet definition itself. Maybe it's worth filing a feature request in VSCode.

gtm-nayan:
Didn't notice that prefixes are completely optional and removing it from the definition hides them from intellisense while still having it show up in the command.

@dummdidumm I think it's ready now.

dummdidumm (approved):
Just tested this out, this no longer shows up when typing because there's no `prefix`, which is great. So for me this is good to go. Only question left is whether or not we want more snippets (if-else, await-catch). I'm in favor with starting out small (like this). @jasonlyu123 what's your opinion on this?

NatoBoram:
Oh, I didn't realize that the extension provided hidden snippets. Anyway, here's mine, if someone wants to take a look at them.

```json
{
  "await": {
    "prefix": "await",
    "body": [
      "{#await $1}",
      "$2",
      "{:then $1}",
      "$0",
      "{:catch}",
      "$3",
      "{/await}"
    ]
  },
  "catch": {
    "prefix": "catch",
    "body": ["{:catch}", "$0"]
  },
  "const": {
    "prefix": "const",
    "body": ["{@const $1 = $2}", "$0"]
  },
  "each": {
    "prefix": "each",
    "body": ["{#each $1 as $2 ($3)}", "$0", "{/each}"]
  },
  "else if": {
    "prefix": "elseif",
    "body": ["{:else if $1}", "$0"]
  },
  "else": {
    "prefix": "else",
    "body": ["{:else}", "$0"]
  },
  "if else": {
    "prefix": "ifelse",
    "body": ["{#if $1}", "$0", "{:else}", "$2", "{/if}"]
  },
  "if": {
    "prefix": "if",
    "body": ["{#if $1}", "$0", "{/if}"]
  },
  "if else if": {
    "prefix": "ifelseif",
    "body": [
      "{#if $1}",
      "$3",
      "{:else if $2}",
      "$4",
      "{:else}",
      "$5",
      "{/if}",
      "$0"
    ]
  },
  "then": {
    "prefix": "then",
    "body": ["{:then $1}", "$0", "{:catch}", "$2"]
  }
}
```

PR #1375, opened 2022-02-07
(fix) no error for component typedef with event

jasonlyu123:
For the new transformation. Currently, it would show an error with a component having events in the typedef. For example: 
```ts
class Component extends SvelteComponentTyped<{}, { click: MouseEvent }, {}>{}
```

### DIFF
--- a/packages/language-server/test/plugins/typescript/features/diagnostics/fixtures/component-invalid/expected.json
+++ b/packages/language-server/test/plugins/typescript/features/diagnostics/fixtures/component-invalid/expected.json
@@ -1,8 +1,8 @@
 [
     {
         "range": {
-            "start": { "line": 13, "character": 1 },
-            "end": { "line": 13, "character": 11 }
+            "start": { "line": 25, "character": 1 },
+            "end": { "line": 25, "character": 11 }
         },
         "severity": 1,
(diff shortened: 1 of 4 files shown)


dummdidumm (approved):
Thanks!

For my future self, this is the error that would appear:
```
Argument of type 'typeof Works2' is not assignable to parameter of type 'new (args: { target: any; props?: any; }) => Svelte2TsxComponent<any, any, any>'.
  Types of construct signatures are incompatible.
    Type 'new (options: IComponentOptions<{ hi: string; }>) => Works2' is not assignable to type 'new (args: { target: any; props?: any; }) => Svelte2TsxComponent<any, any, any>'.
      Construct signature return types 'Works2' and 'Svelte2TsxComponent<any, any, any>' are incompatible.
        The types of '$on' are incompatible between these types.
          Type '<K extends "click">(type: K, callback: (e: { click: MouseEvent; }[K]) => void) => () => void' is not assignable to type '<K extends string>(event: K, handler: (e: any) => any) => () => void'.
            Types of parameters 'type' and 'event' are incompatible.
              Type 'K' is not assignable to type '"click"'.
                Type 'string' is not assignable to type '"click"'.(2345)
```

Issue #1370, opened 2022-02-04
Control flow does not work for uninitialized exported props in script tag


dummdidumm:
**Describe the bug**
Control flow does not work for uninitialized but exported props

**To Reproduce**
```svelte
<script lang="ts">
    type Test =
        | { type: 'TypeA'; uniqueToA: string }
        | { type: 'TypeB'; uniqueToB: string };

    export let test: Test;

    // works
    console.log(test.type);
    
    // doesn't work
    if (test.type == 'TypeA') {
        console.log(test.uniqueToA);
    }

    // works
    $: if (test.type == 'TypeA') {
        console.log(test.uniqueToA);
    }
</script>

<!-- works -->
{#if test.type == 'TypeA'}
    {test.uniqueToA}
{/if}
```

**Expected behavior**
Control flow should work in all cases

**System (please complete the following information):**
 - OS: -
 - IDE: VS Code
 - Plugin/Package: svelte2tsx/language-server

**Additional context**
Possible solution: Initialize the variable in the generated code (AFAIK we do that for booleans already).

Maybe the solution here helps with #1335, but could very well not be the case.

jasonlyu123:
Initializing it would also help #1201. I tried it before. But the sourcemap would cause the completion to be calculated at the end of the generated code. Causing global identifiers completion instead of type completion. Which hurts the experience.

Issue #1201, opened 2021-10-10
[typescript] uninitialized variable produces no error


mohe2015:
### Describe the bug

With
```typescript
let name: string;
```
I would expect that it's ensured the type of `name` is always `string` (minus typecheck errors). Unfortunately svelte-check didn't produce any. This leads to unexpected results as other code relying on that assumption breaks.

I think the better thing would be that an initialization value is required / you need to specify `string | undefined`

### Reproduction

https://github.com/mohe2015/svelte-bug-typescript-uninitialized-slot

### System Info

```shell
System:
    OS: Linux 5.14 NixOS 21.11 (Porcupine) 21.11 (Porcupine)
    CPU: (8) x64 AMD Ryzen 7 4700U with Radeon Graphics
    Memory: 1.97 GB / 15.05 GB
    Container: Yes
    Shell: 5.1.8 - /nix/store/wadmyilr414n7bimxysbny876i2vlm5r-bash-5.1-p8/bin/bash
  Binaries:
    Node: 16.10.0 - /nix/store/501yjfjm2kdk3sj3z3335275q1jdraj8-nodejs-16.10.0/bin/node
    npm: 7.24.0 - /nix/store/501yjfjm2kdk3sj3z3335275q1jdraj8-nodejs-16.10.0/bin/npm
  Browsers:
    Firefox: 93.0
  npmPackages:
    rollup: ^2.3.4 => 2.58.0 
    svelte: ^3.0.0 => 3.43.1
```

### Severity

annoyance

dummdidumm:
When tackling this we need to take into account possibly conflicting requirements:
- we need to make sure TS control Flow doesn't pin such variables to a more specific type once it's assigned
- we need to make sure we do not get "not assigned" when using `bind:something={variable}`

riouh:
Hey! I would also be interested to see this issue sorted out! 🙂 This leads to runtime errors because of variables being actually null, whereas a priori not typed as such in the code, which would be a great thing to avoid.
To specify how in which exact conditions I encountered this problem: I'm using svelte 3.54.0, svelte-check 2.10.2, typescript 4.9.4.
In my tsconfig.json, I have set "strict": "true". According to the [Svelte tests](https://github.com/jasonlyu123/language-tools/blob/master/packages/language-server/test/plugins/typescript/features/diagnostics/fixtures/unInitialized/input.svelte), the following code gives an error "Variable 'hi' is used before being assigned.":
`<script lang="ts">
    let hi: string
[...]

PR #1378, opened 2022-02-09
(fix) initialize props so control flow works

jasonlyu123:
#1370 

This removes the need to suppress the used-before-assigned error for uninitialized props. Also, fix duplicated edit issue with rename props. Find out we could just reuse the `preprendStr` and `overwriteStr` utility for the sourcemap issue I mentioned before. 

I change it to always append the initializing expression when the props has type specified. We could also init the props without types. But the only benefit I found is that it won't be an `undefined` type in strict mode. [See the typescript playground here](https://www.typescriptlang.org/play?#code/DYUwLgBAhg3AUHKA6MB7AymAlgOwOYAUAlPHKJAEbwUQC8EA+gwM4BuIwY4zDAjA1BwBPAhRJwKCAGYBXHAGNsqHIxbtO3PgOEEAJjIC2BoQC5owomcFCIAbwC+QA)


dummdidumm:
Would this also help with #1201 ? AFAIK there's a check in `DiagnosticsProviderImpl` that removes "used before assigned" errors. Not sure if that would mean we also need to add this transformation to untyped variables (what about JSDoc-typed variables btw?).

jasonlyu123:
I think the untyped variable would be implicitly any. So it's acceptable to be used before assigning. But it would be typed undefined if both `strictNullChecks` and `noImplicitAny` was enabled. Because we didn't suppress this error this behaviour also exist before this PR. JSDoc typed variable would also be appended with the initializing expression. 

About #1201, Sorry I was wrong. It won't fix it. I tested it after removing "used before assigned". And realize the uninitialized variable is in the markup. Which mean it's wrapped in a function in svelte2tsx and TypeScript won't complain about it. I must be remembering stuff wrong or miss read the issue at the first place.

dummdidumm (approved):
Makes sense - thank you!

Issue #1302, opened 2021-12-28
Code completion for expressions in the markup doesn't work properly


aradalvand:
**Describe the bug**
Code completion in JavaScript expressions in the markup doesn't work (as least as reliably) as it does in the script tag.

**To Reproduce**
I have a `client` object that contains various functions that send fetch requests, when I'm in the `<script>` and I type `client` and then hit `.`, this is the list of suggestions that appears, exactly what's expected:
![image](https://user-images.githubusercontent.com/26527405/147587330-7ec0ebb4-3714-4316-8842-67a206515cbf.png)

But when I do the same thing in an `{#await}` block, for example, this is the list of suggestions:
![image](https://user-images.githubusercontent.com/26527405/147587760-09f1ef59-6d40-43a9-8d94-06821f640911.png)
As you can see, there is all sorts of unrelated stuff in there. I'm assuming this has something to do with this extension or the Svelte language server.
I've had this problem since the beginning, but I always thought "it's probably gonna be fixed soon", and I assumed there was an already-existing issue for it or something.

**Expected behavior**
I expect to see the members of the `client` objects properly as they are shown when I do `client.` in the script tag.

**System (please complete the following information):**
 - OS: Windows 10
 - IDE: VSCode
 - Plugin/Package: Svelte for VSCode

jasonlyu123:
This is a limitation we currently have. The syntax transformation under the hooks relies on the AST the svelte compiler returned. So it won't work if there's a syntax error in the markup. See sveltejs/svelte#4818. It should be working again once there's no syntax error. Like if you have at least an identifier after `.`.

This won't be completely fixed anytime soon. What we can do is find a workaround for each case to slightly improve it.

ptrxyz:
Funny enough autocompletion works if you try it with brackets. Type `client["` and it'll show you the properties.

And even more funny: if you simply put `client.` (yes, simply this syntactically wrong line) as last line in your **script tag**, autocompletion in the **template** also works fine.

There is some black magic going on here. :man_shrugging:

PR #1381, opened 2022-02-11
improve property access completion in markup

jasonlyu123:
Related to #538, #786 and #1302

This PR added a utility to check if there's trailing property access behind an expression. The Reason we needed this is that we blank the operators before parsing. So the operator won't exist in the AST. We need to check it so that the expression transformation also include it. 

This should only be called with an expression that is a direct child of a template node. And since we only go backwards from the end of the moustache (`}`), it only makes sense to call this with the expression that might end directly before the end of the moustache. 

I also change the v2 transformation for `{@debug}`, `{@const}` and `{@await}` to have a better sourcemap. I can't find a way to make `{@debug}` and `{@const}` work with the `transform` function. So I change it to directly overwrite. 

I have also experimented, but not committed yet, with skipping `)` while blanking operators before parsing here:
https://github.com/sveltejs/language-tools/blob/cad755f7eef616547392fda75299b7605d10b998/packages/svelte2tsx/src/utils/htmlxparser.ts#L167

```diff
- } else if (!/\s/.test(char)) {
+ } else if (char !== ')' && !/\s/.test(char)) {
```
This would let the property access also work in a function call and the key of each block. Haven't thought of a downside for this. Tested and it works well and the existing test still passed. Wondering if I should do this in this PR or a separate one.


packages/svelte2tsx/src/htmlxtojsx_v2/utils/node-utils.ts
```diff
@@ -174,3 +174,43 @@ export function isShortHandAttribute(attr: Node): boolean {
 export function isQuote(str: string): boolean {
     return str === '"' || str === "'";
 }
+
+/*
+ * Found if there's a member access trailing behind an expression
+ * Usually it's there because of the preprocess we have before svelte parses
+ * the template
+ */
+export function getNodeEndIncludingTrailingPropertyAccess(
```
jasonlyu123, line 183: Should I import from the v1 utils? I found there's some utility copied from v1. Maybe the plan is to remove v1 eventually this way it's easier to remove?

packages/language-server/test/plugins/typescript/features/CompletionProvider.test.ts
```diff
@@ -115,6 +115,51 @@ function test(useNewTransformation: boolean) {
             });
         });
 
+        const editingTestPositionsNewOnly: Array<[number, number, string]> = [
```
jasonlyu123, line 118: These four doesn't work in v1. Wondering if it still makes sense to make it work.

dummdidumm: I wouldn't worry too much about the old transformation at this point, so I'm fine with it only working for the new transformation

packages/svelte2tsx/src/htmlxtojsx_v2/utils/node-utils.ts
dummdidumm: Yeah the plan is to remove v1 at some point, so it's okay to duplicate it here

packages/svelte2tsx/src/htmlxtojsx/utils/node-utils.ts
```diff
@@ -273,3 +273,36 @@ export function buildTemplateString(
         str.appendLeft(attr.end, trailingOverride);
     }
 }
+
+/*
+ * Found if there's a member access trailing behind an expression
+ * Usually it's there because of the preprocess we have before svelte parses
+ * the template
+ */
+export function getNodeEndIncludingTrailingPropertyAccess(
```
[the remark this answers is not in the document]
dummdidumm: What about `getOriginalNodeEnd`?

packages/svelte2tsx/src/htmlxtojsx/utils/node-utils.ts
```diff
@@ -273,3 +273,36 @@ export function buildTemplateString(
...
+    originalText: string,
+    position: number
+): number {
+    let index = position;
+
+    while (index < originalText.length) {
```
dummdidumm, line 288: Could we simplify this to `originalText.indexOf('}', position) - 1` ? This is always used in a situation where there's a closing bracket of a mustache tag at the end. Or is this wrong because we only can do this for property assignments and not things like `+`, `&` etc which we also blank?
(having written this out, I think so, so I think it's fine as it is now)

dummdidumm:
Looks good to me, left a couple of small comments. About the `)`: Can't think of a drawback either, but maybe we should do it in a separate PR.

packages/svelte2tsx/src/htmlxtojsx/utils/node-utils.ts
jasonlyu123: Yeah, it's currently only used for `.` and `?.`. Not other operators we blanked. This condition for the while loop is mostly for preventing an infinite loop. In most cases, it would hit other characters and `break` the loop.

dummdidumm (approved):
I shortened the name adjusted the `transform` util which makes the catch overwrite obsolete and as a nice side effect ensures document symbols are correct in more cases now. Good to merge now 👍

PR #1384, opened 2022-02-11
(fix) better ensureComponent check

dummdidumm:
#1352

@jasonlyu123 I wasn't able to come up with a better way to solve both "events" cases described in the shims file. What do you think about the way this is handled now? What I like about it is that there's now a more descriptive error message for the new transformation because it prints out the type in the message.


jasonlyu123:
Looks good. I couldn't find a better way to handle all these event conflicts either. And as you said. It isn't important to be strict. The point is just to have a friendly error for missing component typedef.

PR #1390, opened 2022-02-16
(fix) regression with boolean literal props type inference

jasonlyu123:
Revert part of the change with #1378 regarding this. Misread the condition and thought it can be removed. My bad


dummdidumm:
Ah, slipped my review, too 😄 happens, thanks for the quick fix 👍

PR #1391, opened 2022-02-18
add snapshot test for syntax highlight grammar

jasonlyu123:
the test setup is adapted from [vscode-ng-language-service](https://github.com/angular/vscode-ng-language-service). Might also have to add some edge cases we fixed before.


packages/svelte-vscode/package.json
```diff
@@ -9,7 +9,7 @@
         "build": "npm run build:ts && npm run build:grammar",
         "vscode:prepublish": "npm run build && npm prune --production",
         "watch": "npm run build:grammar && tsc -w -p ./",
-        "test": "echo 'NOOP'"
+        "test": "npm run build:grammar && node test/grammar/test.js"
```
jasonlyu123, line 12: The reason for this `test.js` script is that we can just read the `dummy` directory to build the argument list instead of having to write a big long list in the `package.json`

dummdidumm (approved):
Awesome :tada: This will help us greatly to adjust the grammar in the future and not having to worry about regressions - so much better than having a manual look at all the colors.
I think the tests cover the syntax pretty well, we can always add more if we run into something.

Issue #1382, opened 2022-02-11
Typescript: store variable type inference error


asv7c2:
**Describe the bug**
When using prefix or postfix increment of store variable type is deduced to void.

**To Reproduce**
See screenshots

**Expected behavior**
Type infered to number

**Screenshots**
![1](https://user-images.githubusercontent.com/13184913/153564066-f5ac9616-0838-4379-a763-c1fd7f01d57f.png)
![2](https://user-images.githubusercontent.com/13184913/153564068-7d7d70fc-dd40-4a2d-a414-6072cbc8bffb.png)

**System (please complete the following information):**
 - OS: Windows 10 x64
 - IDE: VSCode
 - Plugin/Package: svelte-language-server

**Additional context**

PR #1395, opened 2022-02-20
(fix) $store++/$store-- return type

dummdidumm:
#1382

This feels like a safe change, but paranoid me feels like this syntax could maybe clash in some situations with other code? I didn't come up with a situation where that's actually true, though - @jasonlyu123 can you think of a situation where the resulting code might be syntactically invalid?


jasonlyu123:
Looks good. I also can't think of a situation that is would cause a problem.

Issue #1358, opened 2022-01-31
Triple reload required in VS Code after updating extension


rafrafek:
**Describe the bug**
After updating Svelte for VSCode extension I have to reload VS Code 3 times.

**To Reproduce**
Update Svelte for VSCode extension.

**Expected behavior**
Single reload should be enough.

**Screenshots**
1.
![pic1](https://user-images.githubusercontent.com/23004737/151769798-685f493b-152c-41ad-be14-f60b8728cafb.png)
2.
![pic2](https://user-images.githubusercontent.com/23004737/151769826-dbf3990b-8f96-45a1-a316-4763ab08191f.png)
3.
![pic3](https://user-images.githubusercontent.com/23004737/151769851-c203b385-9060-4b29-95fe-e11c187d4951.png)

**System (please complete the following information):**
 - OS: Windows
 - IDE: VSCode
 - Plugin/Package: Svelte for VSCode

dummdidumm:
I fear we can't change this. The problem is that it's not possible AFAIK to dynamically enable or disable a TypeScript plugin that comes with an extension. We provide a TypeScript plugin which can be enabled and is disabled by default. To provide a plugin you need to add a `typescriptServerPlugins` entry to the `package.json` of the extension. To disable it we therefore we use somewhat of a hack and add `typescriptServerPlugins-disabled` initially, and if it should be enabled, the `-disabled` part is removed. This causes the third prompt. Now every time an update is installed, the extension's `package.json` is "reset" and the `-disabled` part appears again, making the prompts show up again. We can't tell VS Code to somehow not add the `-disabled` part back.

benignant:
Is there a downside to having the plugin enabled? Maybe it would make sense to have it enabled by default. If most people want it enabled, then it would be the minority who has to go through the triple reload.

dummdidumm:
The plan is to enable it by default some day but I haven't gathered enough feedback yet to feel confident about it - or maybe everyone's using it and happy and I just don't know yet 😄

dummdidumm:
Mhhhm it seems this may actually be possible: https://github.com/microsoft/vscode-typescript-tslint-plugin/issues/5 seems to implement plugin settings through VS Code settings, so if this does what I think it does we could add a setting to the plugin whether or not it should be enabled and get rid of this weird workaround.

PR #1401, opened 2022-03-02
(feat) improve Ts plugin enable config

jasonlyu123:
https://github.com/sveltejs/language-tools/issues/1358. Based on the tslint plugin method. 

The way it works is the svelte extension sends a command to ts extension and then to the tsserver. So it only works when the svelte extension is activated. To achieve that we needed the extension to be started on javascript and typescript files too. While only activating the language server until a svelte file is opened for better performance. 

On the typescript plugin side. There's a new ConfigManager as a config getter. And provide an event for config update so we can clear the cache after the toggle. I changed to use a Proxy while decorating the language service, so it can be easier to toggle it with the enable config. Or we could change to pass config manager to every function that decorates language service methods.

@dummdidumm what do think about this approach? Also, do we still want to send a message when we enable the plugin?

### DIFF
--- a/packages/typescript-plugin/src/index.ts
+++ b/packages/typescript-plugin/src/index.ts
@@ -49,6 +49,7 @@ function init(modules: { typescript: typeof ts }) {
         );
 
         configManager.onConfigurationChanged(() => {
+            // enabling/disabling the plugin means TS has to recompute stuff
             info.languageService.cleanupSemanticCache();
             info.project.markAsDirty();
         });
--- a/packages/typescript-plugin/src/index.ts
+++ b/packages/typescript-plugin/src/index.ts
@@ -4,8 +4,11 @@ import { Logger } from './logger';
 import { patchModuleLoader } from './module-loader';
 import { SvelteSnapshotManager } from './svelte-snapshots';
 import type ts from 'typescript/lib/tsserverlibrary';
+import { ConfigManager, Configuration } from './config-manager';
 
 function init(modules: { typescript: typeof ts }) {
+    const configManager = new ConfigManager();
+
     function create(info: ts.server.PluginCreateInfo) {
         const logger = new Logger(info.project.projectService.logger);
         if (!isSvelteProject(info.project.getCompilerOptions())) {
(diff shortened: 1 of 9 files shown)


packages/typescript-plugin/src/index.ts
```diff
@@ -66,7 +89,11 @@ function init(modules: { typescript: typeof ts }) {
         }
     }
 
-    return { create, getExternalFiles };
+    function onConfigurationChanged(config: Configuration) {
```
dummdidumm, line 92: I saw this setting, too, in another plugin, but wasn't sure if it helps us. Does this mean we don't need to tell people to reload VS Code anymore?

dummdidumm (commented):
This is awesome! Fun fact, I was just about to start looking at this myself, using a very similar approach. What I hadn't figured out yet was how to make the LS start on TS/JS, too, but don't start the server - which you solved already. 

To answer your question: If we don't need to tell people to restart VS Code anymore, then I'd say we don't need that message anymore.

packages/typescript-plugin/src/index.ts
```diff
@@ -4,16 +4,26 @@ import { Logger } from './logger';
...
             return info.languageService;
         }
 
-        logger.log('Starting Svelte plugin');
+        configManager.updateConfigFromPluginConfig(info.config);
+        if (configManager.getConfig().enable) {
+            logger.log('Starting Svelte plugin');
+        } else {
+            logger.log('Svelte plugin disabled');
+            logger.log(info.config);
```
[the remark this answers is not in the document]
dummdidumm: Ah no, wait, we don't since the plugin attaches/detaches itself (because of the Proxy) without the need to restart.

packages/typescript-plugin/src/index.ts
jasonlyu123: Yeah. My understanding is that tsserver would call this function when it receives the update config request from the typescript extension. The tslint plugin also opens a file system watcher to watch the tsconfig file. wondering if we should do that too. 

BTW if you have anything to add. Feel free to do so 😄.

PR #1405, opened 2022-03-07
(feat) Add support for svelte-jsx.d.ts and svelte-native-jsx.d.ts to emitDts

rgossiaux:
The svelte.JSX namespace, and in particular the svelte.JSX.HTMLAttributes interface, is used by several component libraries. But SvelteKit's package command will fail for components that use them because these definitions are missing. See kit#3783

### DIFF
--- a/packages/svelte2tsx/src/emitDts.ts
+++ b/packages/svelte2tsx/src/emitDts.ts
@@ -5,6 +5,8 @@ import { svelte2tsx } from './svelte2tsx';
 export interface EmitDtsConfig {
     declarationDir: string;
     svelteShimsPath: string;
+    svelteJsxPath?: string;
+    svelteNativeJsxPath?: string;
     libRoot?: string;
 }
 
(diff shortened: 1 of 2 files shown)


packages/svelte2tsx/src/emitDts.ts
```diff
@@ -5,6 +5,8 @@ import { svelte2tsx } from './svelte2tsx';
 export interface EmitDtsConfig {
     declarationDir: string;
     svelteShimsPath: string;
+    svelteJsxPath?: string;
+    svelteNativeJsxPath?: string;
```
rgossiaux, line 9: Another option is adding a `[string`] option for additional paths instead of just listing these out individually, and another alternative is `emitDts` just finding & reading these itself (I wasn't sure why it's structured this way right now). *Another* option is combining all of the declarations into one file. 

Obviously you folks would know best, I just opened this PR to get the ball rolling.

dummdidumm:
Closing as the svelte.JSX namespace is now deprecated

Issue #1195, opened 2021-10-05
Make "Go to references" in regular JS/TS recognize uses in Svelte files


cpiber:
**Is your feature request related to a problem? Please describe.**
To share functions, I put them in a .ts file. Using "Go to references" does not find uses of the functions in Svelte files.

**Describe the solution you'd like**
Supply references in Svelte files.

**Describe alternatives you've considered**
\-

**Additional context**
\-

jasonlyu123:
Have you tried the `svelte.enable-ts-plugin` config? this would enable a typescript plugin for "go to references" to find usage in the svelte files

cpiber:
But that's not how finding references works. And I'm not using filesystem-based routing.

In my project, I have one `index.js`, which loads the entry point, which loads other components. Some of the components load `.ts` files and I can go to definition from there, but not back to the references.

dummdidumm:
Could you provide a minimum reproduction repository? Or at least a few code snippets which form a minimum reproducible?

dummdidumm:
The problem is indeed as @jasonlyu123 thought: Svelte files which are not imported in any TS file of the project don't exist for the TS server. We need to enhance the logic so that these files are somehow made known to TS. Maybe finding a way to add extra file extensions to the ts config loading behavior similar to [here](https://github.com/sveltejs/language-tools/blob/master/packages/language-server/src/plugins/typescript/service.ts#L305) can help.

odeniso:
hello hello. has there been any progress on this? :pray:

PR #1407, opened 2022-03-08
(feat) Support svelte project files in typescript plugin

jasonlyu123:
#1195. Tried various solutions and end up similarly as we update the project js/ts files in the svelte language server. 

The `extraFileExtensions` parameter in `parseJsonConfigFileContent` do work but it's less performant. I did find a way to configure the tsserver to accept extra file extensions while parsing config and updating project files. But that method will restart the project and load the plugin again. Meaning an infinite loop until it throws an error. 

Implementation wise, First, we read the directory matching the include, exclude add those files to the typescript server project. And set up a directory watcher base with the watching option in the parsed command line. When there's svelte files update, update the files again. 

The reason why we need to manually add it to the typescript server project is that for reason typescript only add the initial external files to the root. A newly created file won't. Meaning it will never be updated.

### DIFF
--- a/packages/typescript-plugin/src/index.ts
+++ b/packages/typescript-plugin/src/index.ts
@@ -36,7 +36,7 @@ function init(modules: { typescript: typeof ts }): ts.server.PluginModule {
             logger.log(info.config);
         }
 
-        // This called the ConfiguredProject.getParsedCommandLine
+        // This call the ConfiguredProject.getParsedCommandLine
         // where it'll try to load the cached version of the parsedCommandLine
         const parsedCommandLine = info.languageServiceHost.getParsedCommandLine?.(
             getConfigPathForProject(info.project)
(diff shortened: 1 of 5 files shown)


packages/typescript-plugin/src/index.ts
```diff
@@ -57,18 +79,24 @@ function init(modules: { typescript: typeof ts }) {
...
+            decorateLanguageService(info.languageService, snapshotManager, logger, configManager),
+            projectSvelteFilesManager ?? {
+                dispose() {}
+            }
         );
     }
 
-    function getExternalFiles(project: ts.server.ConfiguredProject) {
-        if (!isSvelteProject(project.getCompilerOptions())) {
+    function getExternalFiles(project: ts.server.Project) {
```
jasonlyu123, line 98: This is just a small typing correction. It's not always a ConfiguredProject.

dummdidumm:
Awesome! I tried this out and it works great. I refactored some things:
- made files and watcher a set that always is defined (gets rid of some undefined checks, instead clear the set)
- made sure files are also removed on dispose, and not loaded when plugin is not enabled
- the watcher change can also fire when a file no longer exists, so the file should be removed from the set, added logic for that

I also tried out getting rid of the "read directory files acording to includes/excludes" completely when a watcher fires, but the problem is that the file name is all-lowercase and therefore wrong, it seems the "read directory" method has the correct casings. If there's a way to get the correct casing from the incorrect one that might be a perf improvement, but I didn't find a method to do so.

Anyway, thanks a lot for implementing this :tada:

PR #1408, opened 2022-03-09
(feat) Semantic document highlight

jasonlyu123:
This adds support for [document highlight](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_documentHighlight). Making the document highlight be based on semantic/syntactic meaning instead of the current word-based default by VSCode. This also makes the document highlight available for other LSP client since a lot of them doesn't have a word-based fallback and it does look kind of weird without it. 

As usual, the result is provided by `vscode-css-languageservice`, `vscode-html-languageservice` and typescript. As for the svelte plugin, I highlight the svelte block and tags. For example, if block highlights `#if`, `:else if`, `:else` and `/if`. 

The downside of the current approach is that ts, svelte document highlight in the markup won't work if there's a syntax error in the markup. Also, for unsupported embed language users, like pug, sass, postcss and stylus. It might be a worse experience. With that said I think it would still be a good addition. 

To discuss: 

1. Do we think this is better than the word-based in vscode? If not maybe we could make it disablable and disable this by default in vscode. But enable it in the server by default. For it to work, we have to set the server capability based on config. 

2. To address the worse experience for unsupported embed language. I am thinking if we can implement a word-based by ourselves and return the word-based result for these languages. 

3. The svelte compile result may not be reliable enough. So maybe we can blank the style and script tag like in svelte2tsx and use that instead. Or maybe even move to the typescript plugin and use the AST returned by `svelte2tsx` instead?


igorlfs:
Hey, is there any chance of this getting merged? It'd be nice to have when using other clients

jasonlyu123:
The main reason for the inactive is that this is not reliable. It won't work when there is a syntax error in the template. But maybe it's still better than nothing. We could probably mark it as experimental and off by default.

jasonlyu123:
It's now marked as experimental and off by default. We could make it on by default if people don't care much about the parser error problem. Similar to the folding range. I also added a word-based solution for unsupported embedded languages. The svelte result is also calculated using the AST returned by svelte2tsx. Unlike the folding range, there isn't a word-based fallback when there is a parser error because the result is too different.

packages/svelte-vscode/package.json
```diff
@@ -399,6 +399,12 @@
                         "none",
                         "ts"
                     ]
+                },
+                "svelte.experimental.documentHighlight.enable": {
+                    "type": "boolean",
+                    "default": false,
+                    "title": "Experimental: Document Highlight",
+                    "description": "Enable experimental document highlight support. Require restart."
```
dummdidumm, line 407:
```suggestion
                    "description": "Enable experimental document highlight support. Requires a restart."
```

dummdidumm (commented):
Sorry for taking so long to respond - now that Svelte 5 has a loose parser, does that alleviate some of the concerns with not enabling this by default?

Could you show some screenshots of before/after and also some of the cases where it breaks down? I gotta admit I'm not that knowledgeable about the whole semantic document highlighting topic.

jasonlyu123:
Hmmm. While I was trying to take some screenshots of it breaking. I found that there is weirdly a word-based fallback now and HTML doesn't have the same behaviour. Not sure if this is supposed to be a new feature or a VSCode bug. I also found some bugs with nested blocks. I'll check both later. And about experimental or not, I think I am leaning toward enabling it by default now we have the loose parser.

jasonlyu123:
Okay, the word highlight fallback should be intentional. If we return an empty array instead of null it'll stop the fallback. 

圖片

This is what it looks like when there are highlights. And in the parser error state, it is just nothing. But you could still trigger the word base highlight when double-clicking on the word so maybe it isn't that big of a deal. I think we can enable this by default. And if enough people find the parser error state annoying we can consider leveraging the word highlight fallback in svelte 4 files.

packages/language-server/src/plugins/PluginHost.ts
```diff
@@ -677,6 +678,25 @@ export class PluginHost implements LSProvider, OnWatchFileChanges {
...
+            throw new Error('Cannot call methods on an unopened document');
+        }
+
+        return (
+            this.execute<DocumentHighlight[] | null>(
+                'findDocumentHighlight',
+                [document, position],
+                ExecuteMode.FirstNonNull,
+                'high'
+            ) ?? []
```
dummdidumm, line 696:
```suggestion
            ) ?? [] // fall back to empty array to prevent fallback to word-based highlighting
```

packages/language-server/src/server.ts
```diff
@@ -322,7 +322,7 @@ export function startServer(options?: LSOptions) {
                 },
                 documentHighlightProvider:
                     evt.initializationOptions?.configuration?.svelte.plugin.svelte
-                        ?.documentHighlight?.enable === false
+                        ?.documentHighlight?.enable === true
```
jasonlyu123, line 325: Oops. Got it wrong. What I actually wanted is `evt.initializationOptions?.configuration?.svelte.plugin.svelte?.enable !== false`. But now that I think about it `evt.initializationOptions?.configuration?.svelte.plugin.svelte?.enable ?? true` is probably clearer.

dummdidumm (approved):
More than 2,5 years in the making, kudos for your persistence 😄

PR #1410, opened 2022-03-12
untyped and `any` should have the same behaviour

ivanhofer:
If a component that contains `<script lang="ts">` exports a prop that is not typed, `TypeScript` treats it as `any` and this is also true for the usage in the component itself. But it shows up as `undefined` in other components.

I think these two props should have the same behaviour:
```svelte
<script lang="ts">
    export let untyped;
    export let typedAsAny: any;
</script>
```

I'm not sure where I can fix this, so I just started by writing tests.
Any suggestion where this needs to be fixed?


dummdidumm:
@jasonlyu123 you mentioned this in #1378 and now it came up through this PR. What do you think about also initializing the prop in this case? Doesn't hurt in my opinion, it also gets rid of some "properly undefined" errors when using an uninitialized prop during component initialization.

jasonlyu123:
Looks ok. I also think it wouldn't hurt. Can't think of other dropbacks.

dummdidumm:
Alright, thanks for second pair of eyes!
Thanks @ivanhofer for getting the ball rolling on this one!

PR #1412, opened 2022-03-14
(feat) make it possible to pass additional file paths to emitDts

dummdidumm:
Needed when someone uses the svelte.JSX typings.
Closes #1405

Other possibilities:
- type this as `string` like in #1405 -> decided against that because it's not flexible enough for other use cases
- use `require.resolve` for this to get the `d.ts` file -> decided against that because people might use this in other environments than HTML (Svelte Native) and they couldn't pass their file in then

### DIFF
--- a/packages/svelte2tsx/index.d.ts
+++ b/packages/svelte2tsx/index.d.ts
@@ -79,6 +79,11 @@ export interface EmitDtsConfig {
      */
     svelteShimsPath: string;
     /**
+     * Additional file paths that need to be present when doing the dts emit.
+     * Example: `[require.resolve('svelte2tsx/svelte-jsx.d.ts')]` for the Svelte HTML ambient type definitions
+     */
+    additionalPaths?: string[];
+    /**
      * If you want to emit types only for part of your project,
      * then set this to the folder for which the types should be emitted.
      * Most of the time you don't need this. For SvelteKit, this is for example
--- a/packages/svelte2tsx/src/emitDts.ts
+++ b/packages/svelte2tsx/src/emitDts.ts
@@ -5,6 +5,7 @@ import { svelte2tsx } from './svelte2tsx';
 export interface EmitDtsConfig {
     declarationDir: string;
     svelteShimsPath: string;
+    additionalPaths?: string[];
     libRoot?: string;
 }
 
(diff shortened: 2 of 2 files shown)


dummdidumm:
Marking this as draft because different thought just occured to me: 
- What if we add the ambient typings to the `@types` namespace? Would this be something we should do in the long run anyway?
- Is this solvable by adding another reference to the `app.d.ts` (`/// <reference types="svelte2tsx/svelte-jsx" />`, @rgossiaux does this work for you?) template in the case of SvelteKit?

cc @jasonlyu123

rgossiaux:
> * Is this solvable by adding another reference to the `app.d.ts` (`/// <reference types="svelte2tsx/svelte-jsx" />`, @rgossiaux does this work for you?) template in the case of SvelteKit?

From what I see the `app.d.ts` is not passed to emitDts, since it lives outside the `lib/` directory, so you'd still need a config change to solve that. But regardless, anything that solves the problem works for me, I'm not picky 😃

dummdidumm:
Oh right you're correct, that wouldn't work for you in this case. You could add a `d.ts` file to the lib folder with the mentioned content and exclude it from being exported, maybe that works.

fallaciousreasoning:
Wondering if there was any movement on this, or did you end up solving this in a different way?

fallaciousreasoning:
FWIW, I'm working on a Svelte component library, and we're compiling the Svelte components to web-components. We're using `svelte2tsx` to generate the type definitions which we use to get intellisense & typings in a React wrapper. This works really well for most cases, but when the component is something primitive (like a button or a link) and we want to expose the props on the underlying HTML element `svelte2tsx` breaks.

Currently, I have a fork of `svelte2tsx` with this change, which I'm vendoring into the repository and it works really nicely. Any chance we could land this? It solves this really nicely :smile:

jasonlyu123:
The plan is to move the element typing to the svelte core. You can see the roadmap here https://github.com/sveltejs/language-tools/issues/1552

fallaciousreasoning:
Okay, got this working without the `svelte2tsx` fork by bumping Svelte to 3.55.0 and using `SvelteHTMLElements['button']` instead of the `svelte2tsx` typings. Thanks @jasonlyu123!

dummdidumm:
Since Svelte 3.55 it's possible to enhance from HTML typings through the `svelte/elements` import. I think this PR might be obsolete because of this.

PR #1434, opened 2022-04-08
Fancy issue templates

gtm-nayan:
Current ones leave artefacts like 
image
if users (*mostly me*) forget to delete them


dummdidumm:
Good idea, thank you!

Issue #1435, opened 2022-04-09
support svelte:element


Zachiah:
Correct me if I'm wrong, but I searched through the issues, and I didn't find any with svelte:element. I updated svelte-check and svelte to latest versions...

I tried updating svelte, svelte-check and installing the latest version of svelte-language-server none of which helped

dummdidumm:
Prettier needed a bump which I needed to wait for, support will land soon, no worries.

PR #1436, opened 2022-04-09
(feat) support <svelte:element>

baseballyama:
fix: https://github.com/sveltejs/language-tools/issues/1435

I don't have much knowledge about language-tools, but I implemented <svelte:element> so I thought I would respond.
Is this PR correct?
I have at least followed the procedure described in the README and the issue is resolved.

### DIFF
--- a/packages/language-server/package.json
+++ b/packages/language-server/package.json
@@ -55,7 +55,7 @@
         "prettier": "2.5.1",
         "prettier-plugin-svelte": "~2.6.0",
         "source-map": "^0.7.3",
-        "svelte": "^3.46.1",
+        "svelte": "^3.47.0",
         "svelte-preprocess": "~4.10.1",
         "svelte2tsx": "~0.5.0",
         "typescript": "*",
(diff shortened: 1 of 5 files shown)


baseballyama:
I didn't think the test would fail, but actually, it did fail in one case, so I'm checking now.

baseballyama:
I found that this is the effect of this PR. We will update the test.
https://github.com/sveltejs/svelte/pull/7232

https://github.com/sveltejs/language-tools/blob/master/packages/language-server/test/plugins/svelte/features/getDiagnostics.test.ts#L468

jasonlyu123:
We still need add the `svelte:element` to [here](https://github.com/sveltejs/language-tools/blob/7d94e237f7b51453e076005ca2750a5ac6bdfb07/packages/language-server/src/plugins/html/dataProvider.ts#L192) for the completion. 

And we still need to support it in the svelte2tsx. I have a quick look and there are multiple places needed to change. Are you interested in adding support for svelte2tsx? If so I can give you some pointers. Or we could also add it later.

dummdidumm:
I have a working branch locally already I couldn't complete because prettier needed a bump, too. Closing this in the meantime - thank you.

Issue #1338, opened 2022-01-16
Support source.sortImports code action


janosh:
**Is your feature request related to a problem? Please describe.**

I like this extension's import organizing feature but prefer it not to auto-delete unused imports, i.e. sorting only. VS Code offers the [`"source.sortImports"` code action](https://code.visualstudio.com/updates/v1_57#_sort-imports-source-action) for this which unlike `"source.organizeImports"` leaves unused imports in place. It doesn't with Svelte files though:

```json5
  "[svelte]": {
    "editor.defaultFormatter": "svelte.svelte-vscode",
    "editor.codeActionsOnSave": [
      "source.fixAll",
      "source.organizeImports", // sorts and removes unused imports
      "source.sortImports" // does nothing
    ]
  },
```

**Describe the solution you'd like**

Add support for `"source.sortImports"`.

PR #1446, opened 2022-04-12
(feat) Sort imports

jasonlyu123:
#1338


gyurielf:
Awesome! Thanks for this nice feature @jasonlyu123 !

packages/language-server/src/plugins/typescript/features/CodeActionsProvider.ts
```diff
@@ -51,6 +56,27 @@ export class CodeActionsProviderImpl implements CodeActionsProvider {
...
+        if (context.only?.[0] === SORT_IMPORT_CODE_ACTION_KIND) {
+            return await this.organizeImports(
+                document,
+                cancellationToken,
+                /**skipDestructiveCodeActions */ true
+            );
+        }
+
+        // for source action command (all source.xxx)
+        // vscode would show different source code action kinds to choose from
```
dummdidumm, line 68: How can I trigger this code path? And what would happen then - would VS Code show a prompt of which action I prefer?

jasonlyu123: type 'Source Action' in the command palette. This is the only command I find that is easy to debug this code action in the UI.

dummdidumm (approved):

Issue #1444, opened 2022-04-12
no error when importing non-existent `.svelte` file


DetachHead:

dummdidumm:
The problem is the 
```ts
declare module "*.svelte";
```
definition inside Svelte core, which tells TS to treat every file ending with `.svelte` as if it exists. It can't be removed because people would get errors for all their Svelte file imports inside TypeScript files if they don't enable the TS plugin. I'm not sure how to make the diagnostic errors appear in another way. If there was a way to somehow ignore/exclude that ambient file definition (we know where inside the Svelte package it is located) that could be a way to do it. We would also need to remove our similar module declaration inside `svelte2tsx`

DetachHead:
> It can't be removed because people would get errors for all their Svelte file imports inside TypeScript files if they don't enable the TS plugin.

is there a good reason to be using typescript without also using the plugin though? that just sounds to me like their project isn't set up correctly

PR #1448, opened 2022-04-13
(feat) detect invalid Svelte import paths

dummdidumm:
Detect that an import of a .svelte file is pointing to a non-existent file by removing the ambient module definitions in svelte and svelte2tsx from the eyes of TS.

If you want the old behavior back, add a `ambient.d.ts` file (or similar; you can add it to `app.d.ts` in a SvelteKit project for example) with the following:
```ts
declare module "*.svelte";
```

#1444

@jasonlyu123 please tell me if this too hacky or actually a good idea 😅

### DIFF
--- a/packages/language-server/src/plugins/typescript/DocumentSnapshot.ts
+++ b/packages/language-server/src/plugins/typescript/DocumentSnapshot.ts
@@ -144,9 +144,15 @@ export namespace DocumentSnapshot {
             originalText = ts.sys.readFile(filePath) || '';
         }
         if (normalizedPath.endsWith('svelte2tsx/svelte-shims.d.ts')) {
-            originalText =
-                originalText.substring(0, originalText.indexOf('// -- start svelte-ls-remove --')) +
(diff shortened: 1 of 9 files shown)


jasonlyu123:
I am ok with this. 

Alternatively, we can also add the error by ourselves. We could use the type-checker to get the symbol of the `ImportDeclartion.moduleSpecifier`. And see if the symbol name is "*.svelte" or if the declaration is from the two files. But to turn this into a diagnostic message. We need to either use typescript internal APIs or copy the error message. So it's also kind of hacky.   And we still need to filter out the `*.svelte` hover or completion.

dummdidumm:
Yeah, I feel that this would be just as hacky and more code - ok, I'll merge this then, thanks for your insights!

DetachHead:
@dummdidumm thanks for fixing this! one question though, is this change supposed to only effect the behavior in vscode's intellisense? as the `ambient.d.ts` module is still present in the `svelte` package so when i run `svelte-check` it still has the old behavior

Issue #1461, opened 2022-04-25
Broken imports after moving file


dummdidumm:
### Describe the bug

Two people have reported that moving a file results in a broken import

### Reproduction

This happens when

- TS plugin is enabled
- a TS file is moved/renamed that is imported in both at least one TS _and_ Svelte file

Then the `$/getEditsForFileRename` in our extension is fired, and afterwards the "update imports?" prompt of VS Code shows up. In some cases this doesn't hurt as the double update is idempotent, but not in all cases, so sometimes it messes up imports.

Not sure what the best way to fix this is. Don't do our import when the moved file is a TS file and it's imported in a Svelte file and the TS plugin is enabled (needs checks)? Somehow move our automatic import to the prompt as well and hope that this is somehow deduplicatable? Enhance the TS plugin to deduplicate on that end?

Edit: When deduplicating the import, the import is still broken, turns out it's the TS plugin which breaks the import, we need to adjust the range. Edit 2: Turns out that's not enough, sometimes the imports will still be broken (because they overwrite the same range which may have changed) - so we need to deduplicate this.

### Expected behaviour

Flawless imports

### System Info

- OS: [e.g. Windows]
- IDE: [e.g. VSCode, Atom]

### Which package is the issue about?

Svelte for VS Code

PR #1466, opened 2022-05-01
(fix) prevent duplicate import updates

dummdidumm:
The TS plugin and the Svelte extension want to update the same locations after a file move/rename in some cases; prevent that
#1461


dummdidumm:
Just realized that the second commit might make things worse. If there's a TS file and both a TS and Svelte file is updated (moving a folder), maybe the TS server then operates on outdated ranges, leading to broken imports again. Need to check that when I have time.

janosh:
@dummdidumm Great work as always on the language tools. They make the Svelte writing experience a lot better!

PR #1477, opened 2022-05-10
(feat) auto-create page endpoints

dummdidumm:
Adds a prompt that shows up when adding a new route to a SvelteKit app, which generates a corresponding page endpoint file.

TODO
- [x] find out if this is really a SvelteKit app (check package.json)
- [ ] read svelte.config.js because people can customize the routes folder (?) (maybe later, not sure how feasible this is since we need to read a JS file)
- [x] finish always/never handling

Questions:
- UX ok? Is a info window in the lower right better?
- Should the page endpoint boiler plate be customizable?

### DIFF
--- a/packages/svelte-vscode/package.json
+++ b/packages/svelte-vscode/package.json
@@ -102,6 +102,16 @@
                     "default": "off",
                     "description": "Traces the communication between VS Code and the Svelte Language Server."
                 },
+                "svelte.kit.createEndpoints": {
+                    "type": "string",
+                    "enum": [
+                        "Ask",
+                        "Always",
+                        "Never"
+                    ],
+                    "default": "Ask",
+                    "description": "Creates a corresponding endpoint when creating a new Svelte page file"
+                },
                 "svelte.plugin.typescript.enable": {
                     "type": "boolean",
                     "default": true,
--- a/packages/svelte-vscode/package.json
+++ b/packages/svelte-vscode/package.json
@@ -102,16 +102,6 @@
                     "default": "off",
                     "description": "Traces the communication between VS Code and the Svelte Language Server."
                 },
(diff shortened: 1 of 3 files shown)


boehs:
What problems does this solve?

david-plugge:
This pr would make life so much easier since the new sveltekit routing requires more files. I´ve started building an extension on my own until i´ve discovered this pr. It isnt ready in any means but maybe its helpful so i just [post it here](https://gitlab.com/david-plugge/vscode-svelte-utils)

dummdidumm:
Oh this looks awesome! I was meaning to pick up this PR again but didn't have time to yet, your solution looks very promising. If you want you can create a new PR in here (doesn't have to be complete, just a starting point) and we can go from there.

Things I imagined:
- The action can be started from a command or a context menu entry in the folder view (right click on a folder -> "SvelteKit: Create Route")
- The action is a multiselect (`window.showQuickPick`) and you select what you want from there
- Ideally the action infers from the project if the files should be TS or JS. Idea: Check for presence of a `.ts` file in the routes folder, or a `tsconfig.json`

david-plugge:
This is the context menu to add a new component.

image

And this is the multiselect if you choose the first option.
I excluded the `+server.ts` option.

image

After choosing the components another prompt pops up asking for the path with a default of `/`.
I thought about using the routes path eg. `src/routes` if the context menu is opened from outside that path, otherwise use the relative path.
1. open the menu at `src/routes/projects`
2. enter path `users`
 -> resolves to components beeing generated at `src/routes/projects/users`.

---

I will create a new PR next week :)
Maybe the discussion about Actions will be resolved by then.

dummdidumm:
Closing as obsolete

Issue #1470, opened 2022-05-03
@const can be inside an if block as well


stephane-vanraes:
### Describe the bug

Since Svelte 3.48 `@const` is allowed to be inside an `#if` as well, the extension however still complains about this.

![image](https://user-images.githubusercontent.com/13446670/166422777-81693054-dfa6-4a81-b0bc-b0e5282ec26f.png)

### Reproduction

```
{#if true}
  {@const test = 123}
{/if}
```

### Expected behaviour

Code above is not marked as an error.

### System Info

- OS: [Windows]
- IDE: [VSCode]
- Svelte for Vs Code: [v105.16.0]

### Which package is the issue about?

Svelte for VS Code

lacherogwu:
Same here 😄
<img width="1631" alt="Screen Shot 2022-05-03 at 10 52 26" src="https://user-images.githubusercontent.com/49835614/166428128-28b0b65f-bcac-4fa5-a9b0-1f07afa2416d.png">

stephane-vanraes:
I suppose this is just a matter of bumping the svelte version in the extension.

jasonlyu123:
We might need some time with this one. It's tricky to implement this into the svelte2tsx transformation for type check.

dummdidumm:
I guess this is easier for the new transformation. Should we only implement it for that one, which also will give us more users for it, which should help get more feedback before switching?

dummdidumm:
Turns out this already works without modifications for the new transformation 👍

jasonlyu123:
I have some ideas about how to do this in the old transformation but haven't finished it yet. I can add a PR later. Don't have to wait for it though I also think we can get the new transformation supported first.

PR #1478, opened 2022-05-10
(feat) support const tag in if block in the old transformation

jasonlyu123:
#1470. Based upon #1474. We could merge that first. 

Supporting const tag in the if block with an IIFE. So typescript would still retain the control flow. Not sure if there's a case where it won't retain the control flow state.


dummdidumm (approved):
Good to know that TS is smart and keeps control flow in IIFEs!

PR #1487, opened 2022-05-17
(chore) add debugger config for ts plugin

jasonlyu123:
https://github.com/microsoft/TypeScript/wiki/Debugging-Language-Service-in-VS-Code#debugging-tsserver-server-side

added an environment variable for the tsserver to launch with inspector enabled when we launch extension for debugging. So we can attach a debugger like what we did for language-server. Do we want to keep the sourcemap in the published plugin? If not, we might need to create another tsconfig to distinguish.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -10,7 +10,10 @@
             "stopOnEntry": false,
             "sourceMaps": true,
             "outFiles": ["${workspaceRoot}/packages/svelte-vscode/dist/**/*.js"],
-            "preLaunchTask": "npm: watch"
+            "preLaunchTask": "npm: watch",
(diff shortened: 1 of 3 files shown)


dummdidumm:
Awesome! Good question about those source maps. I don't think that we need them in production, but it's also a question of whether or not it really hurts. How many additional kilobytes are we talking about?

jasonlyu123:
20KB. About 25%.

dummdidumm:
Just realized we have a `.npmignore` file. Could you add the source maps to that? Then they are not published and we don't need an extra build step.

dummdidumm (approved):
🎉

PR #1488, opened 2022-05-17
(fix) prevent crash on some language client with unknown code action kind

jasonlyu123:
After the sort import version is released. I saw my side project language client for the visual studio crash because of the unknown CodeActionKind. I hesitate to fix it here as the protocol did specify that the language-server client should not panic when there's an unknown `CodeActionKind` but I can't find a way to fix it on my side. 

This adds a initialize option to force the server to filter out code actions the client doesn't support. If another client also has a similar problem they could also pass this option to fix the problem. Didn't choose to only filter out the "sort import". Because it might throw another error once there's a new one that it's not yet supported in the future. 

@dummdidumm What do you think about this?


dummdidumm (approved):
I'm okay with this change 👍

PR #1496, opened 2022-05-26
(chore) bumps typescript to 4.7

jasonlyu123:
Draft for now as it seems like the string literal prop completion is broken. From my debugging, it seems like typescript now looks at the first argument of the component class construtor instead of the `JSX.ElementAttributesProperty`. Since we plan to switch to the new transformation, not sure if it's still worth the effort to ask the typescript team to support it.

### DIFF
--- a/packages/language-server/src/plugins/typescript/features/FindReferencesProvider.ts
+++ b/packages/language-server/src/plugins/typescript/features/FindReferencesProvider.ts
@@ -1,7 +1,7 @@
 import ts from 'typescript';
 import { Location, Position, ReferenceContext } from 'vscode-languageserver';
 import { Document } from '../../../lib/documents';
-import { pathToUrl } from '../../../utils';
+import { flatten, pathToUrl } from '../../../utils';
 import { FindReferencesProvider } from '../../interfaces';
 import { LSAndTSDocResolver } from '../LSAndTSDocResolver';
 import { convertToLocationRange, hasNonZeroRange } from '../utils';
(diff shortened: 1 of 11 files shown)


packages/language-server/src/plugins/typescript/features/FindReferencesProvider.ts
```diff
@@ -18,7 +18,7 @@ export class FindReferencesProviderImpl implements FindReferencesProvider {
         const { lang, tsDoc } = await this.getLSAndTSDoc(document);
         const fragment = tsDoc.getFragment();
 
-        const references = lang.getReferencesAtPosition(
```
jasonlyu123, line 21: Don't know what's the difference between these two methods. But the one we used before no longer returns `isDefinition`

dummdidumm:
I'm ok with it no longer working, it doesn't appear in many places and I want to switch to the new transformation in about a month.

packages/language-server/src/plugins/typescript/features/FindReferencesProvider.ts
dummdidumm: From a quick TS codebase search it seems that `findReferences` is like `getReferencesAtPosition` with an additional step afterwards, at least they both call `getReferencedSymbolsForNode` which does the actual work

jasonlyu123: Found a place where it does make a difference. If you use a go-to reference on the definition.
 It would invoke `includeDeclaration: true`. Previously if there is only one reference it would take you to the reference.  if we use the `findReferences` and remove the check it would not take you to the reference but show both reference and definition.

jasonlyu123:
I ported the string literal completion as a fallback before you commented XD. I think we can use this fallback and if someone finds any difference we can just tell them it won't matter soon. 

It's ready for merge now. Do we want to merge it now? Or do we want to wait a few days until vscode release a new version with 4.7? To see if people have problems with 4.7.

dummdidumm (approved):
Changes look good 👍 although I'm glad we can get rid of all the conditional checks for the new transformation once we remove the old one 😄

PR #1502, opened 2022-05-28
(fix) relax event handler typings

dummdidumm:
null and undefined are valid inputs
sveltejs/svelte#7568

As draft because Svelte itself actually doesn't support it, either.


dummdidumm:
This was implemented in https://github.com/sveltejs/svelte/pull/7863 now

PR #1506, opened 2022-05-31
svelte-check: Ignore files if --ignore and --tsconfig have been passed

### LINKED ISSUE #1495: svelte-check: Ignore diagnostics for imported modules
### Description

We use svelte-check@2.2.6 and Typescript. We also use generated code in our code basis and would like to have the diagnostics for those modules ignored.
We can neither use the ``--ignore`` flag nor an exclude setting in the ``tsconfig.json``  to solve this issue, because it does not work for modules who get imported by modules which are getting diagnosed.

### Proposed solution

Let the diagnostics for modules be ignored/suppressed, also when they get imported by a module which gets diagnosed.
So, e.g. the lines https://github.com/sveltejs/language-tools/blob/master/packages/language-server/src/plugins/PluginHost.ts#L75-L87 implement this feature hard-coded for files under ``node_modules``. It would be nice if this would also work for any file specified by the ``--ignore`` flag (or any newly-introduced flag).

### Alternatives

_No response_

_No response_

cbartz:
Closes #1495

### DIFF
--- a/packages/language-server/src/svelte-check.ts
+++ b/packages/language-server/src/svelte-check.ts
@@ -12,7 +12,8 @@ import {
     TypeScriptPlugin
 } from './plugins';
 import { convertRange, getDiagnosticTag, mapSeverity } from './plugins/typescript/utils';
-import { pathToUrl, urlToPath } from './utils';
+import { normalizePath, pathToUrl, urlToPath } from './utils';
+import * as path from 'path';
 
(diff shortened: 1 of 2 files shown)


dummdidumm:
This looks good, thank you. See my comment in the issue why I'm hesitant to merge this.

dummdidumm:
Closing, since we went with a different route - `--ignore` is no longer allowed together with `tsconfig`

Issue #1210, opened 2021-10-22
Go to References - to where the component is used


lukaszpolowczyk:
**Is your feature request related to a problem? Please describe.**
I have an active tab open with a `*.svelte` component file in VSCode.
I want to jump to all uses of this component.

**Describe the solution you'd like**.
With the tab with the `*.svelte` component file open in VSCode:
* right-click in any empty line of the file
* choose "Go to References" (or the special menuitem "Go to References for this svelte component")
* jumps to each use of this component (of course it bypasses the `import from` sections - it only shows the places of use).

Another option: Menuitem in the context menu of the `*.svelte` file in Expoler and in the context menu of the tab with the `*.svelte` file.

**Describe alternatives you've considered**.
Searching via search engine (Ctrl+Shift+F) and typing component name.
This is less convenient because you have to type it manually, and because it also searches the `import from` section and any other place with the same name as the component, but it should bypass those sections.

Jojoshua:
This is exactly the feature I am needing as well. Anything I can do to help?

Blankeos:
Found: 
🪟 Ctrl (🍏 CMD ) + Shift + P: `Svelte: Find Component References`

PR #1509, opened 2022-06-02
(feat) 1210/where component used

Jojoshua:
Implementation for https://github.com/sveltejs/language-tools/issues/1210

- Finds usages of the svelte component

https://user-images.githubusercontent.com/5598169/171748575-55cda689-44f1-4737-b4bf-bcc76f977446.mp4


packages/language-server/src/plugins/typescript/features/FindComponentUsagesProvider.ts
```diff
@@ -0,0 +1,276 @@
...
+
+    private convertTextSpan(
+        ref: ReferenceEntry,
+        element: Location,
+        fragment: SnapshotFragment,
+        helper: FindReferenceTargetHelper
+    ) {
+        const textSpan = convertToTextSpan(element.range, fragment);
+
+        //For whatever reason the references found in the script section of a .svelte file were off by 1
```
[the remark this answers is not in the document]
dummdidumm: Use https://github.com/sveltejs/language-tools/blob/master/packages/language-server/src/lib/documents/DocumentMapper.ts#L197 instead which handles the off-by-one error. Problem is the generated source mappings. The range ends at the start index of the next character, which often isn't mapped/transformed, so the mapping falls back one character.

dummdidumm:
Thank you for tackling this! My idea for implementing was a little different: The generated code contains a default export which is used in all other places. If we do "find references" on this generated class, the typescript language service should be able to do all the work of finding the correct positions for us. You can find more about what the generated code is and how it fits into the rest here: https://github.com/sveltejs/language-tools/blob/master/docs/internal/overview.md

Jojoshua:
> Thank you for tackling this! My idea for implementing was a little different: The generated code contains a default export which is used in all other places. If we do "find references" on this generated class, the typescript language service should be able to do all the work of finding the correct positions for us. You can find more about what the generated code is and how it fits into the rest here: https://github.com/sveltejs/language-tools/blob/master/docs/internal/overview.md

I will take a look at that. I see there was a link to the readme but it was very easy to miss its importance. This inspired https://github.com/sveltejs/language-tools/pull/1510

Jojoshua:
> If we do "find references" on this generated class, the typescript language service should be able to do all the work of finding the correct positions for us. 

@dummdidumm This sounds like a better way of doing it, but I'm confused...why wasn't the original implementation of find reference one this way? Should FindReferencesProvider.ts be changed to make this happen?

Jojoshua:
What is the preferred method of getting the position of the default export from the generated class?

Jojoshua:
I could be missing something, but I looked at the generated code today and I am not seeing how this job is any easier or cleaner. The generated class from the 'child' is not mapped into the generated class of the 'parent'.

````
"///<reference types=\"svelte\" />;

import Component1 from \"./find-component-usages-child.svelte\";

function render() {    test(); 
const theModule2 = import(\"./find-component-usages-child.svelte\");

import(\"./find-component-usages-child.svelte\").then((module) => {    new module.default({      target: document.body as HTMLElement,    });  }); 

async function test() {    
    const theModule = await import(\"./find-component-usages-child.svelte\");
};
    
async () => { 
    { const $$_Component10C = __sveltets_2_ensureComponent(Component1); new $$_Component10C({ target: __sveltets_2_any(), props: {}});} 
    { const $$_Component10C = __sveltets_2_ensureComponent(Component1); new $$_Component10C({ target: __sveltets_2_any(), props: {}});
    { svelteHTML.createElement(\"p\", {});  }   { svelteHTML.createElement(\"p\", {});  } Component1}};
    return { props: {}, slots: {}, getters: {}, events: {} }}

export default class FindComponentUsagesParent__SvelteComponent_ extends __sveltets_1_createSvelte2TsxComponent(__sveltets_1_with_any_event(render())) {}"
````

**Comment by dummdidumm** [comment]
The idea is the following:
- someone requests component usages of component X
- load the TS document
- find the offset of `FindComponentUsagesParent__SvelteComponent_ ` (it will be in the generated code, there's no mapping to the original, but that's irrelevant)
- do "find references" like you normally would

Since it's very close to the normal "find references" code, I would add the code as another method to https://github.com/sveltejs/language-tools/blob/master/packages/language-server/src/plugins/typescript/features/FindReferencesProvider.ts and refactor the existing code a little so all code starting at https://github.com/sveltejs/language-tools/blob/master/packages/language-server/src/plugins/typescript/features/FindReferencesProvider.ts#L21 can be reused.

The advantage is that TS does all the work of finding references for us - not only in other Svelte files, but TS files, too.

File: packages/language-server/src/plugins/typescript/features/FindComponentUsagesProvider.ts [hunk 2]
```diff
@@ -1,105 +1,79 @@
...
     constructor(private readonly lsAndTsDocResolver: LSAndTSDocResolver) {}
 
     async findComponentUsages(uri: string): Promise<Location[] | null> {
         const u = URI.parse(uri);
         const fileName = u.fsPath;
-        const lang = await this.getLSForPath(fileName);
+        const document = await this.getDocument(fileName);
+        const { lang, tsDoc } = await this.getLSAndTSDoc(document);
+        const fragment = tsDoc.getFragment();
+        const ignoreImports =
```
Jojoshua, line 21: Is this the right way to pull user configuration? It didn't seem like it was switching on/off when I changed my vscode setting for it

packages/language-server/src/plugins/typescript/features/FindComponentUsagesProvider.ts
```diff
@@ -1,105 +1,79 @@
...
+            lsConfig.getConfig().typescript.findComponentUsagesIgnoresImports.enable;
```
Jojoshua, line 22: Is this the right way to pull user configuration? It didn't seem like it was switching on/off when I changed my vscode setting for it

Jojoshua:
I reworked to use the find references concept but decided to keep it in its own file because there were some special cases such as ignoring end tags and the option to ignore imports that probably shouldn't be part of the standard file references code.

packages/language-server/src/plugins/typescript/features/FindComponentUsagesProvider.ts
```diff
@@ -0,0 +1,127 @@
...
+import { isNoTextSpanInGeneratedCode, SnapshotFragmentMap } from './utils';
+import { SvelteSnapshotFragment } from '../DocumentSnapshot';
+import { lsConfig } from '../../../ls-config';
+export class FindComponentUsagesProviderImpl implements FindComponentUsagesProvider {
+    constructor(private readonly lsAndTsDocResolver: LSAndTSDocResolver) {}
+
+    async findComponentUsages(uri: string): Promise<Location[] | null> {
+        const u = URI.parse(uri);
+        const fileName = u.fsPath;
+        const document = await this.getDocument(fileName);
```
dummdidumm, line 18: I think it would be better if resolving to a document would be done in PluginHost already, so you don't need to duplicate the logic here and so that the documents state is consistent.

packages/language-server/src/ls-config.ts
```diff
@@ -15,6 +15,8 @@ const defaultLSConfig: LSConfig = {
         definitions: { enable: true },
         findReferences: { enable: true },
         fileReferences: { enable: true },
+        findComponentUsages: { enable: true },
+        findComponentUsagesIgnoresImports: { enable: true },
```
dummdidumm, line 19: It was requested in the referenced issue but I'm against adding such a functionality (see comments) for consistency. It also makes the implementation a little easier and makes your question obsolete I think.

packages/language-server/src/plugins/typescript/features/FindComponentUsagesProvider.ts
```diff
@@ -0,0 +1,127 @@
...
+
+        // Some references are in generated code but not wrapped with explicit ignore comments.
+        // These show up as zero-length ranges, so filter them out.
+        return locations.filter(hasNonZeroRange);
+    }
+
+    //Return the position of the class in the generated code as the reference target
+    private GetClassPosition(fragment: SvelteSnapshotFragment) {
+        return fragment.positionAt(
+            fragment.getLineOffsets()[fragment.getLineOffsets().length - 2] + 23
```
dummdidumm, line 69: This won't work in all cases, there are situations where getters are added to the class, so it would have a different line offset. I think it's better to do `lastIndexOf('__SvelteComponent__')` (appended to every component export name).

packages/language-server/src/plugins/typescript/features/FindComponentUsagesProvider.ts
```diff
@@ -0,0 +1,127 @@
...
+        const docs = new SnapshotFragmentMap(this.lsAndTsDocResolver);
+        docs.set(tsDoc.filePath, { fragment, snapshot: tsDoc });
+
+        const locations = await Promise.all(
+            flatten(references.map((ref) => ref.references))
+                .filter((ref) => !ref.isDefinition)
+                .filter(this.notInGeneratedCode(tsDoc.getFullText()))
+                .map(async (ref) => {
+                    const defDoc = await docs.retrieveFragment(ref.fileName);
+                    const document = await this.getDocument(ref.fileName);
```
dummdidumm, line 41: You can't assume it's a Svelte document here (DocumentManager only handles these, not very clear from the name I realize now), you need to use fragment/snapshot similar to other code here - but it will be irrelevant anyway if you delete the "filter out imports" code I assume

packages/language-server/src/plugins/typescript/features/FindComponentUsagesProvider.ts
```diff
@@ -0,0 +1,127 @@
...
+                })
+        );
+
+        // Some references are in generated code but not wrapped with explicit ignore comments.
+        // These show up as zero-length ranges, so filter them out.
+        return locations.filter(hasNonZeroRange);
+    }
+
+    //Return the position of the class in the generated code as the reference target
+    private GetClassPosition(fragment: SvelteSnapshotFragment) {
```
dummdidumm, line 67: First letter should be lowercase

dummdidumm (changes requested):
Thank you, looks much more how I imagined it already! I left some comments, the biggest being the request to remove the "filter out imports" logic which should also make the code a little easier.

packages/language-server/src/ls-config.ts
Jojoshua: I was back and forth on it. I ended up with an option because this overlapped some of the file reference functionality. I don't mind to remove the option if we don't care about overlap.

packages/language-server/src/plugins/typescript/features/FindComponentUsagesProvider.ts
Jojoshua: Is this what you had in mind? https://github.com/sveltejs/language-tools/pull/1509/commits/ddba005203791ef5f840848d26f645654c1f4d60

Jojoshua: > Is this what you had in mind? [ddba005](https://github.com/sveltejs/language-tools/commit/ddba005203791ef5f840848d26f645654c1f4d60)

Actually this is causing an issue because it would be pushing the getDocument all the way back to server.ts . I was originally modeling after fileReferences where it simply passes the URI around.

packages/language-server/src/plugins/typescript/features/FindComponentUsagesProvider.ts
Jojoshua: I would still need this to filter out the end tag(which should only be svelte docs) but this had been working for .TS files just fine. Did you expect it not to work? Also, I dont understand the difference between the different flavors of "document" 

What would be a good explanation of...
- Document
- DocumentSnapshot
- SvelteDocumentSnaphot
- SvelteSnapshotFragment
- TextDocumentItem
- JSOrTSDocumentSnapshot
- SnapshotFragment

packages/language-server/src/plugins/typescript/features/FindComponentUsagesProvider.ts
dummdidumm: `lastIndexOf` called on the full text of the fragment will return the offset, no further steps needed

Jojoshua: Excellent, got it

packages/language-server/src/ls-config.ts
Jojoshua: Done

packages/language-server/package.json
```diff
@@ -6,6 +6,7 @@
     "typings": "dist/src/index",
     "scripts": {
         "test": "cross-env TS_NODE_TRANSPILE_ONLY=true mocha --require ts-node/register \"test/**/*.ts\" --exclude \"test/**/*.d.ts\"",
+        "FindComponentUsagesProvider": "cross-env TS_NODE_TRANSPILE_ONLY=true mocha --require ts-node/register \"test/**/FindComponentUsagesProvider.test.ts\" --exclude \"test/**/*.d.ts\"",
```
Jojoshua, line 9: Remove before merge

packages/language-server/src/plugins/typescript/features/FindComponentUsagesProvider.ts
Jojoshua: Looked at this again and still having issues. Trying to use getDocument in pluginhost produce an error saying it was unopened, not sure what's going on or if this really needs to be changed

dummdidumm:
This is all a little hard to explain so I adjusted the code myself.

> What's all these different document/snapshots about?

`Document` is used for Svelte documents. Everything with snapshot in it is used in the TS plugin only, and it's there because we use the TS language service which requires us to implement their snapshot API, which is essentially an immutable state of a document at a certain point in time.

> `const theModule2 = import("./find-component-usages-child.svelte");` is not found

That's because it's not a component usage. `const theModule2 = (await import("./find-component-usages-child.svelte")).default;` would be.

Jojoshua:
> I'm also thinking about whether or not we should rename this to "Svelte: Find Component References" (usages -> references) to align with the nomenclature of VS Code - @jasonlyu123 thoughts?

I was thinking about this also. It does end up in the References window

image

packages/language-server/src/plugins/typescript/features/FindComponentUsagesProvider.ts
```diff
@@ -0,0 +1,87 @@
...
+        );
+
+        return locations.filter(isNotNullOrUndefined);
+    }
+
+    private offsetOfComponentExport(fragment: SvelteSnapshotFragment) {
+        return fragment.text.lastIndexOf(COMPONENT_SUFFIX);
+    }
+
+    private isEndTag(span: ts.TextSpan, document: DocumentSnapshot) {
```
Jojoshua, line 83: I'm not sure if there is a problem here or elsewhere but end tags are showing up again

Jojoshua: Fixed

Jojoshua:
> This is all a little hard to explain so I adjusted the code myself. Hope this is ok - let me know if anything is unclear.

👍 


> > What's all these different document/snapshots about?
> 
> `Document` is used for Svelte documents. Everything with snapshot in it is used in the TS plugin only, and it's there because we use the TS language service which requires us to implement their snapshot API, which is essentially an immutable state of a document at a certain point in time.

Ahh ok that helps a bit with terminology. I'll need to read up on the TS language service API

dummdidumm:
Thank you so much for driving this forward! I adjusted the name (usages -> references), made the documentmanager thing obsolete again by exposing the text on the snapshot, and removed the "force references view" part in the extension, which means it will open this in a peek view if the user has configured it like that. I made this to a) respect the user config and b) so we don't create these empty vs code files

packages/language-server/src/plugins/typescript/DocumentSnapshot.ts
```diff
@@ -273,6 +273,10 @@ export class SvelteDocumentSnapshot implements DocumentSnapshot {
         return this.parent.getFilePath() || '';
     }
 
+    getOriginalText(range?: Range) {
```
Jojoshua, line 276: cool

Jojoshua:
> made the documentmanager thing obsolete again by exposing the text on the snapshot,
Perfect!

Blankeos:
Thanks for this. I'm more used to the Alt + Shift + F12 approach in TSX components.
For the ones that want to only use their keyboard, I guess Ctrl + Shift + P: `Svelte: Find Component References` works as well.

Let me know if there's a faster hotkey for this without typing.

PR #1510, opened 2022-06-03
Diagram for system overview

Jojoshua:
It's difficult to get a grasp of everything in the tooling arena. I think a diagram of the architecture would help newcomers (and myself). This is what I've tried to come up with on my own and now I need @jasonlyu123 @dummdidumm to review & complete. 

During this process it may help you think of new ways to organize/split up certain packages.

https://marketplace.visualstudio.com/items?itemName=bierner.markdown-mermaid helps to preview the readme w/the diagram but ive attached the initial preview

mermaid-diagram-20220603130036

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,68 +74,47 @@ Converts a .svelte file into a legal TypeScript file. Built from [halfnelson/sve
 ```mermaid
 flowchart LR
     %% IDEs
-    VSC[VSCode]
-
+    VSC[IDE: VSCode + Svelte for VS Code extension]
+    click VSC "https://github.com/sveltejs/language-tools/tree/master/packages/svelte-vscode" "Svelte for VSCode extension"
     %% Tools
-    CLI[CLI]
-
+    CLI[CLI: svelte-check]
+    click CLI "https://github.com/sveltejs/language-tools/tree/master/packages/svelte-check" "A command line tool to get diagnostics for Svelte code"
     %% Svelte - Extensions
-    VSC_TSSP[vscode-typescript-svelte-plugin]
-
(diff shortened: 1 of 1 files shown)


dummdidumm:
Thanks, that's a great idea! I adjusted the graph:
- reworded some stuff so they match the npm package names
- reordered graph a little

Let me know if this makes sense to you

Jojoshua:
> Thanks, that's a great idea! I adjusted the graph:
> 
> * reworded some stuff so they match the npm package names
> * reordered graph a little
> 
> Let me know if this makes sense to you

Thank you for this update @dummdidumm , it does make sense. 

In the initial draft I was having trouble with the svelte-check->language-server piece so I'm glad you addressed that. I was assuming that anything talking to a "server" must be a "client".  Is svelte-check's communication to the svelte-language-server being done differently and you don't really consider it an LSP client?

I also like the additional information in the description for the embedded language services, that helps quickly grasp its purpose.

dummdidumm:
svelte-check bypasses the LSP entry point completely and uses a different entry point specifically made for it. This trick allows us to reuse the diagnostics functionality of the language server without extracting it out to some other package

Jojoshua:
> svelte-check bypasses the LSP entry point completely and uses a different entry point specifically made for it. This trick allows us to reuse the diagnostics functionality of the language server without extracting it out to some other package

Does this entry point allow it to access everything the language server can do such as access all the embedded services on the far right side?

dummdidumm:
Yes, all these are accessed (but only their diagnostics part of course)

Jojoshua:
Thanks, I'm good with this

Issue #1508, opened 2022-06-02
Import statements moved to context="module" on save


billmcknight1953:
When I save a file that has both  an ordinary <script language="ts"> tag, and a <script context="module" language="ts"> tag,  all the import statements are moved to the script context="module" tag.  If I disable the Svelte for VS Code plugin, the save operation doesn't move the import statements.  Can I disable this move operation without disabling the Svelte for VS Code plugin?

I am using version v.105.16.1

dummdidumm:
Out of curiousity: Why is that a problem for you? It doesn't make a difference where the import appears, it works either way.

billmcknight1953:
If it makes no difference where the import statements exist, then I will close this request

billmcknight1953:
If it makes no difference then I will close this request

Issue #1486, opened 2022-05-16
organizeImports should use newlines between groups


seanlail:
### Description

I had a look through https://github.com/sveltejs/language-tools/issues/83 and I think this has not been mentioned yet.

Currently when I save a file eslint and the svelte plugin are fighting each other.
My non-svelte files are using this plugin to sort imports:
https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/order.md

These are the rules I have:
```javascript
"import/newline-after-import": ["error", { count: 1 }],
"import/order": [
  "warn",
  {
    "newlines-between": "always",
    alphabetize: {
      order: "asc",
      caseInsensitive: true,
    },
  },
],
```

<table>
<tr>
<th>Expected</th>
<th>Actual</th>
</tr>
<tr>
<td>

```svelte
<script lang="ts">
  import { onMount } from "svelte";
  import { writable } from "svelte/store";

  import ALib from "$lib/ALib.svelte";
  import BLib from "$lib/BLib.svelte";

  import ARelative from "./ARelative.svelte";
  import BRelative from "./BRelative.svelte";
</script>
```
</td>
<td>

```svelte
<script lang="ts">
  import ALib from "$lib/ALib.svelte";
  import BLib from "$lib/BLib.svelte";
  import { onMount } from "svelte";
  import { writable } from "svelte/store";
  import ARelative from "./ARelative.svelte";
  import BRelative from "./BRelative.svelte";
</script>
```
</td>
</tr>
</table>

### Proposed solution

Allow a way to configure this or use the eslint rules

### Alternatives

live with the Svelte files being formatted differently.

dummdidumm:
If you say "conflict" do you mean that the formatter (a prettier plugin) conflicts with it. Does that mean you have "formatOnSave" or "organizeImports on save" turned on? What happens if you turn it off?

Btw the eslint-plugin-svelte3 is known to not play nice with some of the rules of that import plugin (but the one you mention is not listed there, who knows) https://github.com/sveltejs/eslint-plugin-svelte3/blob/master/OTHER_PLUGINS.md

seanlail:
@dummdidumm Yes, sorry for not being clearer. I have Code set to format on save with eslint. 

So yes, the eslint-plugin-svelte3 is conflicting with something and whatever is running last is resulting in the above format. 

I'll create a repo as soon as I can to show the issue.

seanlail:
[...]

Issue #1269, opened 2021-12-08
Don't send text changes if updateImports changes nothing


dummdidumm:
This came out of https://github.com/sveltejs/language-tools/pull/1261#issuecomment-988915940 : When organizing imports but these don't change anything, we should somehow detect this to not send a needless update to the client.

PR #1515, opened 2022-06-10
(feat) support ts 4.7 group aware organize import

jasonlyu123:
#1486

Added a check in the svelte2tsx to add a blank line if the import matches the import organization group in the ts 4.7. Also fixed a bug introduced when updated to ts4.7 where an empty indent would be added sometimes. Not sure if there's another way to fix it. 

We could probably use the group-aware organize import for #1508 if we add a blank line to separate module script import with instance import. Is this something we want to support?


jasonlyu123:
Side note. Because of the new group-aware organize import I found some behaviour differences between Linux / Mac and Windows. This caused the test to fail when the group didn't need organizing. Maybe it's the cause of  #1269. Because we always use `\n` between the moved import. Maybe that is the reason why Typescript always think it needs organizing. 

I removed the semicolon in some tests so that it always has the edit. But maybe we should find a way to filter it out.

dummdidumm:
Mhhm yeah how to test this? 😅  Are we able to somehow hijack reading the file during tests and insert the line breaks in a way we want?

jasonlyu123:
Wondering if we can check the line ending on the original file. If the file is all CRLF or all LF then we use it. If the file has a mixed line ending or no line break we use the system line ending. This should at least help folks using LF in Windows and make our test more reliable. 

And then we can try to #1269 later. I tried it on a ts file and it seems like the ts extension also just applies the edit even if nothing changes. So maybe that's not that important. 

If we want to go this route do we also apply it to other ts features that accept format options? If so we probably can work around the failed test here. And do it in another PR.

dummdidumm:
> If we want to go this route do we also apply it to other ts features that accept format options? If so we probably can work around the failed test here. And do it in another PR.

Sounds good 👍 What are other areas where we need this in? If we do need this elsewhere, we should extract the "what kind of line ending to use?"-logic into a small function (in another PR then, doesnt have to be now).

Is this PR good to merge now? Seems like you added the logic and the tests look good.

jasonlyu123:
Yeah. We probably can extract it out later. I think it's ready to merge.

Issue #83, opened 2020-05-12
Overview of current features of language server


dummdidumm:
This is an overview of the current state of the `svelte-language-server` - what it already can do, what it cannot, yet. If you feel a feature is missing or something that should work according to this list does not work, open a new issue.

#### autocompletion / auto imports
- [x] autocompletion for normal html tags
- [x] autocompletion for special svelte tags like `<svelte:window>`
- [x] autocompletion for special svelte tags like `#await`

#### hover info
- [x] hover info for normal html tags
- [x] hover info for special svelte tags like `<svelte:window>`
- [x] hover info for special svelte tags like `#await`

## CSS

- [x] support for CSS/Less/SCSS

#### autocompletion / auto imports
- [x] autocompletion for css tags
- [x] autocompletion for `:global` tag

#### hover info
- [x] hover info for css tags

#### go to definition
- [ ] go to definition from class in template to css
- [ ] go to definition from css to template

## Javascript / Typescript

- [ ] semantic highlighting (#71)
- [x] show svelte parser/compiler warnings/errors
- [ ] typescript-support for $-syntax (`$: ...` and `$variableSubcribeShortHand`) (#57)

#### go to definition
- [x] go to definition within script
- [x] go to definition to non-svelte file
- [ ] go to definition from template to script (#57)
- [ ] go to definition from component-tag to component (#57)
- [ ] go to definition from component-property in template to component (#57)

#### typechecking
- [x] typechecking capabilities for script
- [ ] typechecking capabilities for component props (#57)
- [ ] typechecking capabilities for component events
- [x] support for `//@ts-check` / `//@ts-nocheck`
- [x] JSDoc support

#### autocompletion / auto imports
- [x] autocompletion/autoimport inside script
- [ ] autocomplete/autoimport of svelte component inside template (#57)
- [ ] autocomplete of svelte component props (#57)
- [ ] autocomplete of svelte component events

#### code actions
- [ ] organize imports support (#72)
- [ ] rename function/variable
- [ ] rename svelte component default import
- [ ] rename svelte component prop

PR #1518, opened 2022-06-11
(feat)Go to definition from class in template to css

Jojoshua:
Implementation help for go to css definition from template in https://github.com/sveltejs/language-tools/issues/83

Works for many different class formats:
- Standard 
  - class="test test1"
  
- Conditional Expression
  - class="{current === 'baz' ? 'selected' : ''

- Class Directive
  - class:active="{current === 'foo'}"

- HTML Element Tag
```
<main>

<style>
 main {
 ...
 }
</style>
```


Jojoshua:
@dummdidumm Side-note https://github.com/sveltejs/language-tools/issues/83 regarding go to definition from css to template, I don't feel this is a "**go to definition**" concern but rather a **find references** because the class can be used many places within the template

packages/language-server/src/plugins/css/CSSClassDefinitionLocator.ts
```diff
@@ -0,0 +1,167 @@
...
+                if (!this.IsExactClassMatch(targetRange)) {
+                    continue;
+                }
+
+                return targetRange;
+            }
+        }
+    }
+
+    public IsOffsetWithinStyleSection(offsetPosition: number) {
```
dummdidumm, line 130: This and other methods should probably be private since they are not used outside of the class.

packages/language-server/src/plugins/css/CSSClassDefinitionLocator.ts
```diff
@@ -0,0 +1,167 @@
...
+    initialNodeAt: SvelteNode;
+    constructor(
+        public tsDoc: SvelteDocumentSnapshot,
+        public position: Position,
+        public document: Document
+    ) {
+        this.initialNodeAt = this.tsDoc.svelteNodeAt(this.position) as SvelteNode;
+    }
+
+    public GetCSSClassDefinition() {
```
dummdidumm, line 15: The public methods don't need `public` in front, at least that's the code style in this repo. Also please use lowercase-first method names.

packages/language-server/src/plugins/css/CSSClassDefinitionLocator.ts
```diff
@@ -0,0 +1,167 @@
...
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    public IsExactClassMatch(testRange: Range) {
+        //Check nothing before the test position
+        if (testRange.start.character > 0) {
```
dummdidumm, line 145: This would fail for something like `.foo.bar { .. }` or `.foo .bar { .. }`. In general it's a little brittle to only use text matching here. The vscode-css-language-service which is used in the CSSPlugin provides an AST I think, but it's private. This puts us at risk of breaking changes but I feel it's more robust to use that one instead and fall back to this text-based matching only if parsing fails (because a different preprocessor is used for example).

packages/language-server/src/plugins/typescript/TypeScriptPlugin.ts
```diff
@@ -329,6 +330,31 @@ export class TypeScriptPlugin
         const { lang, tsDoc } = await this.getLSAndTSDoc(document);
         const mainFragment = tsDoc.getFragment();
 
+        const cssClassHelper = new CSSClassDefinitionLocator(tsDoc, position, document);
```
dummdidumm, line 333: The idea of the plugin folders is that they are mostly independent of each other, so it would be better to put this logic into the CSSPlugin or move the class into the TypeScript plugin folder - but I feel like the best place would be `SveltePlugin.ts`. This brings us to a question to which I don't have a good answer yet: What is the best place for code like this? It's cross-cutting so it's not obvious to me in which plugin this does belong. Maybe this hints at that the plugin system needs a redesign / needs to be replaced with something new, but that's for another day.

dummdidumm (changes requested):
Thank you for tackling this! The idea is good, I left some comments on the implementation, where code should go in my opinion and what a more robust class lookup might look like

packages/language-server/src/plugins/css/CSSClassDefinitionLocator.ts
Jojoshua: got it

packages/language-server/src/plugins/css/CSSClassDefinitionLocator.ts
Jojoshua: habit from another language, thanks for noting

packages/language-server/src/plugins/css/CSSClassDefinitionLocator.ts
Jojoshua: I had a feeling about that one. The case I thought would be the most straightforward turned out, not. I will try to improve it. I was looking for a way to pull the text for an entire line but couldn't find it

Jojoshua: Note that svelte doesn't recognize the .foo .bar case either

image

Jojoshua: @dummdidumm So this actually did already work for **.foo .bar** but did not for **.foo.bar**

Try the latest, it handles all sorts of combinations. I feel this is a good compromise especially since I really don't feel that style's inside a svelte component would/should be very complex. If after testing you still feel like we need more horsepower I can look into the CSS AST(just point me in the right direction for that)

packages/language-server/src/plugins/typescript/TypeScriptPlugin.ts
Jojoshua: Feel free to relocate wherever you feel is best

Jojoshua:
Had a chance to look at again @dummdidumm?

dummdidumm:
Sorry, have some other things on my plate currently, but I'll get back to this in the next weeks!

Jojoshua:
@dummdidumm Just checking in -- didn't want to lose too much context on this as time passed

packages/language-server/src/plugins/typescript/CSSClassDefinitionLocator.ts
```diff
@@ -0,0 +1,168 @@
...
+    cssClassTerminators = ['', '{', '.', '>', '~', '>', '[', ':', '#', '+'];
+    constructor(
+        public tsDoc: SvelteDocumentSnapshot,
+        public position: Position,
+        public document: Document
+    ) {
+        this.initialNodeAt = this.tsDoc.svelteNodeAt(this.position) as SvelteNode;
+    }
+
+    getCSSClassDefinition() {
```
dummdidumm, line 16: I believe it could be more promising/faster to defer CSS parsing to `vscode-css-languageservice` which has a parse method. The AST that is returned doesn't have type definitions because it's private, but I'm honestly not worried too much about that since it's unlikely to change for this. That AST could be traversed to more easily find the correct node to link to.

packages/language-server/src/plugins/typescript/TypeScriptPlugin.ts
```diff
@@ -329,6 +330,31 @@ export class TypeScriptPlugin
...
+                LocationLink.create(
+                    pathToUrl(document.getFilePath() as string),
+                    cssDefinitionRange,
+                    cssDefinitionRange,
+                    originRange
+                )
+            );
+
+            if (results) {
+                return results;
```
dummdidumm, line 354: The results should probably be merged with what else comes up. In the case of `class:foo` both `.foo { ... }` in `<style>` and `let foo = ...` in `script` are definitons, and we would lose the latter when returning early.

dummdidumm (commented):
Sorry for leaving you hanging for so long, my primary focus is on SvelteKit right now. I left two comments from what I think could be improved, hope this helps.

lukaszpolowczyk:
Definitely this function should also take into account the name of the element (tagName)
As far as the proposal doesn't take it into account.

Just referring to classes is not enough.

Also the recognition of fixed attributes can be taken into account.

packages/language-server/src/plugins/typescript/CSSClassDefinitionLocator.ts
Jojoshua: Looking into this again. Using svelteNodeAt is giving back a lot of good Svelte information that I wouldn't think the vscode-css-languageservice would know about. This method will tell us all the other variations of Svelte css such as whether it is a ConditionalExpression or Class Directive Format.

packages/language-server/src/plugins/typescript/TypeScriptPlugin.ts
Jojoshua: I'm not sure I follow. In all the tests and variations I've used it seems to work. When you hit F12 to go to definition there can only be 1 result (if a definition can be found).

Jojoshua:
> Definitely this function should also take into account the name of the element (tagName) As far as the proposal doesn't take it into account.
> 
> Just referring to classes is not enough.
> 
> Also the recognition of fixed attributes can be taken into account.

Done via https://github.com/sveltejs/language-tools/pull/1518/commits/601023610bca627a07c52dfcca877921ddb5cb08

packages/language-server/src/plugins/typescript/CSSClassDefinitionLocator.ts
dummdidumm: I didn't mean the HTML part, I meant the CSS part. The code does string traversal of the CSS to find the matching definition, but it could be more promising to use the CSS AST output of `vscode-css-languageservice`

Jojoshua: Are you referring to the while loop in getStandardFormatClassName() ? This is actually not traversing css but the class attribute string itself to figure out which classname the position is at  

`class="test te|st1 bar foo"` so in this example the cursor is in the middle of test1 so that's what we're going to look for in the style section

packages/language-server/src/plugins/typescript/CSSClassDefinitionLocator.ts
```diff
@@ -0,0 +1,203 @@
...
+            result.start.character -= 1;
+            if (this.initialNodeAt.name) {
+                result.end.character = originalStart + this.initialNodeAt.name.length;
+            }
+
+            return result;
+        }
+    }
+
+    private getDefinitionRangeWithinStyleSection(targetClass: string) {
```
dummdidumm, line 137: Apologies, the comment on line `getCSSClassDefinition` was actually meant to appear here. That code does a somewhat unperformant string search. I believe it could be more promising/faster to defer CSS parsing to vscode-css-languageservice which has a parse method. The AST that is returned doesn't have type definitions because it's private, but I'm honestly not worried too much about that since it's unlikely to change for this. That AST could be traversed to more easily find the correct node to link to.

PR #1520, opened 2022-06-14
(fix) remove some options

dummdidumm:
These options don't make sense to be toggleable, because the user has to perform an explicit action to use them, and because we can't hide/disable the corresponding menu entries/shortcuts if the user choses to disable them

@jasonlyu123 thoughts? Something else that should be enabled always? Something I removed which should still be an option?


jasonlyu123:
I think `selectionRange` might make sense so it can be disabled in a specific plugin. Others I am ok with it.

PR #1521, opened 2022-06-14
(refactor) move fragment into snapshot

dummdidumm:
It was necessary to have it seperately because the fragment was initialized asynchronously because of the source map library. With the new source map library this is synchronous, and therefore unnecessarily complex. This moves the logic into the snapshot. @jasonlyu123 let me know what you think of these changes 😃


jasonlyu123 (approved):
Looks good. It does look simpler.

Issue #1522, opened 2022-06-15
Support TypeScript `"moduleResolution": "node16"`


JuanM04:
### Description

In [TypeScript 4.7](https://devblogs.microsoft.com/typescript/announcing-typescript-4-7), along the property `"module": "node16"` that enables ESM features in TypeScript, there are new values for the property `"moduleResolution"`: `"node16"` and `"nodenext"`. Both values allows importing from dependencies with [package.json exports, imports, and self-referencing](https://devblogs.microsoft.com/typescript/announcing-typescript-4-7/#package-json-exports-imports-and-self-referencing). Right now, this property is hard-coded to `"node"`.

### Proposed solution

The svelte-language-server shouldn't hard-code that property and instead read it from the `tsconfig.json`

dan-peterson:
Hey @dummdidumm does this update for the language server need to be manually published to npm? Eager to try it out, but noticed NPM is listing it as 0.14.28 / updated 16 days ago.

dummdidumm:
The fixed label means it's fixed on master but not released yet. when the issue is closed it's released. You could try it out yourself by building from sourcey the getting started instructions in the main readme if this repo should help with that

dan-peterson:
Okay cool I tried it out but encountered an issue.

It seems to honor my module resolution settings now, and that fixes the issue importing from a module that uses the new `exports` field in `package.json`. The language server no longer reports the error: "Cannot find module x or its corresponding type declarations".

However, it seems to no longer understand svelte file imports.

The problem may be unrelated / out of the scope of svelte-language-server, or even an issue with how I built it locally, I'm not sure, but figured it was worth noting here.

---

As a base case I did a clean install of SvelteKit with the demo code and updated the tsconfig:

```json
"moduleResolution": "node16",
``` 

Now in `index.svelte` the language server reports the error:

```js
import Counter from '$lib/Counter.svelte';
```
`Cannot find module '$lib/Counter.svelte' or its corresponding type declarations.`

And dropping the alias didn't solve the issue:

```js
import Counter from '../lib/Counter.svelte';
```

`Cannot find module '../lib/Counter.svelte' or its corresponding type declarations.`


**Note that in all cases the svelte project builds and runs fine. It's just the language server that's reporting errors.**

dummdidumm:
Thank you, not sure how that slipped through. Should be fixed now.

dan-peterson:
> Should be fixed now.

Confirmed on my end.

PR #1526, opened 2022-06-16
(feat) support node16/nodenext

dummdidumm:
#1522

Not sure how to test this, I thought that this would error if I leave out `.js` in imports but that's not the case, so I'm not sure if this is actually working. @jasonlyu123 @JuanM04 do you know more? Edit: Turns out that was an error on our part, we need to pass a new parameter to resolveModuleNames in the TypeScript plugin.

@JuanM04 would be great if you could test this out:
1. Check out this branch
2. Make sure you have yarn 1.x
3. yarn install
4. yarn bootstrap
5. go in language-server and run yarn build
6. go in svelte-check and run yarn build
7. copy over dist folder into your project (replace the dist folder of your existing svelte-check in your node_modules)

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
         "lint": "prettier --check . && eslint \"packages/**/*.{ts,js}\""
     },
     "dependencies": {
-        "typescript": "^4.7.2"
+        "typescript": "^4.7.3"
     },
     "devDependencies": {
         "@sveltejs/eslint-config": "github:sveltejs/eslint-config#v5.2.0",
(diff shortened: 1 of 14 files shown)


JuanM04:
I tried it in my project where I first saw the bug and it seems to work!

dummdidumm:
Great to hear that! Was the bug around not respecting the exports map of a package? What I'm still confused about is that I don't seem to get errors from TS about missing file endings (.js) from relative imports

JuanM04:
> Was the bug around not respecting the exports map of a package?

Yes, that's how I've spotted it. In fact, I have a dependency with broken exports and now svelte-check detects it.

> What I'm still confused about is that I don't seem to get errors from TS about missing file endings (.js) from relative imports

Maybe svelte2tsx has something to do with it?

dummdidumm:
Mhm we might need to also check the resolution mode, as implemented here https://github.com/TypeStrong/ts-node/issues/1778 and exposed since 4.7 (https://github.com/microsoft/TypeScript/pull/49360)

Issue #1475, opened 2022-05-07
Guard for non-nullish value nested in store is reset after `#each`


fehnomenal:
### Describe the bug

Strange title I know...

Minimum reproducible example is:

![image](https://user-images.githubusercontent.com/9959940/167240736-bd861657-3e91-4e5b-9892-f1a5f42fcd74.png)

The error only occurs when using a store that contains an object with a field that is optional or `| null` or `| undefined` and this field is an object.
After `#each`ing over the `foo` strings (or even writing `{#each [$data.item.foo] as str}` when having only a `string`) the type safety of `$data.item` is reset and I get an error like in the screenshot. This is even the case when testing like `{#if $data?.item?.foo}`.

Without the `each` I get no error.

### Reproduction

```svelte
<script lang="ts">
  import type { Readable } from 'svelte/store';

  let data: Readable<{
    item?: {
      foo: string[];
    };
  }>;
</script>

{#if $data?.item}
  {#each $data.item.foo as str}
    <li>{str}</li>
  {/each}

  <pre>{JSON.stringify($data.item.foo)}</pre>
{/if}
```

### Expected behaviour

Well, typescript should know that the value is still non-nuliish.

### System Info

- OS: NixOS
- IDE: VSCodium with official extension Svelte for VS Code (both latest version)

### Which package is the issue about?

svelte2tsx

### Additional Information, eg. Screenshots

I'm not sure if the problem is with `svelte2tsx` and I don't know how to verify this.

The problem seems related to #1341

jasonlyu123:
I can't reproduce this on both VSCode and VSCodium. What exactly is the version of  Svelte for VS Code and VSCodium you're using? The latest version is 105.16.0 which works with VSCode 1.52.0 or later. If you're indeed in the latest version can you provide a reproduction in a repository form?

fehnomenal:
Extension is at v105.16.0

VSCodium (I noticed there was a new release yesterday):
```
Version: 1.66.2
Commit: dfd34e8260c270da74b5c2d86d61aee4b6d56977
Date: 2022-04-13T00:16:29.523Z
Electron: 17.2.0
Chromium: 98.0.4758.109
Node.js: 16.13.0
V8: 9.8.177.11-electron.0
OS: Linux x64 5.15.34
```

Do you have `"strict": true` in your `tsconfig.json`? Only with `true` I get these errors...

I will try to reproduce it in a simple repo.

jasonlyu123:
[...]

PR #1531, opened 2022-06-19
(fix) store rework

dummdidumm:
Reworks $store handling in the code base. Previously, svelte2tsx would transform `$store` access to something like `(__get(store), $store)`. This is now no longer the case. This means we need to add some extra code to some of the intellisense features:
- if someone renames a $store, store also needs to be renamed and vice versa
- if someone wants to find references of a $store, we also need to find store references and vice versa
- if someone wants to go to the definition of a $store, we need to reroute that to store
- if someone wants to go to the implementation of a $store, we need to reroute that to store
- diagnostics of wrong $store usages need to be handled

This means some more code overall, but I feel it's the better choice since we had so many edge cases with our transformation other time, and #1475 being an upstream bug we cannot solve ourselves but need to solve for the new transformation.


jasonlyu123:
Maybe we also need to check the $store variable in the `go-to-implementation. Probably only need to check for `store` for `$store` and don't need the other way around.

dummdidumm:
Do you mean  "right-click on `$store` -> Go to implementations -> get taken to `store`"?

dummdidumm:
Added that, thanks for noticing. Does this look good to otherwise or is there something that needs changing (codestyle, missing cases)?

jasonlyu123 (approved):

Issue #1505, opened 2022-05-30
Autocomplete not working for sass `@use` statements


Dan503:
### Describe the bug

Autocomplete suggestions are not working for sass `@use` statements.

### Reproduction

Use [the Svelte template](https://github.com/sveltejs/template) as a starting point, open it in VSCode.

Install (dart) sass `npm install sass`

Create a `test.scss` scss file to import.

Add an scss style block to a .svelte file with an `@use` statement pointing to the test scss file

```
<style lang="scss">
    @use './test.scss';
</style>
```

### Expected behaviour

Autocomplete for `@use` should work the same as it does for script imports.

### System Info

- OS: Windows 10
- IDE: VSCode, Atom

### Which package is the issue about?

Svelte for VS Code

### Additional Information, eg. Screenshots

![image](https://user-images.githubusercontent.com/10610368/170975761-5f7ba9fc-47f3-452f-99cf-f241c62b0af4.png)

![image](https://user-images.githubusercontent.com/10610368/170975810-2fcc264a-47a0-4bb1-986b-db54a43f0201.png)

PR #1533, opened 2022-06-20
(feat) css path completion

jasonlyu123:
#1505 

implemented a `FileSystemProvider` the `vscode-css-languageervice` expected to support path completion. The `DocuementContext` is adopted from the vscode source. And the `FileSystemProvider` has some differences with the implementation of vscode but I adopted the error handling.


packages/language-server/src/lib/FileSystemProvider.ts
```diff
@@ -0,0 +1,94 @@
...
+} from 'vscode-css-languageservice';
+import { urlToPath } from '../utils';
+
+interface StatLike {
+    isDirectory(): boolean;
+    isFile(): boolean;
+    isSymbolicLink(): boolean;
+}
+
+export class FileSystemProvider implements CSSFileSystemProvider {
```
dummdidumm, line 16: You added this to lib and it's only used inside `plugins/css`. What's the reason for this? Do you think this could be useful outside the CSS plugin, too, or that it's not directly connected to the CSS plugin?

packages/language-server/src/lib/FileSystemProvider.ts
```diff
@@ -0,0 +1,94 @@
...
+
+        return {
+            ctime: stat.ctimeMs,
+            mtime: stat.mtimeMs,
+            size: stat.size,
+            type: this.getFileType(stat)
+        };
+    }
+
+    private UnknownStat(): FileStat {
```
dummdidumm, line 62: lowercase first

packages/language-server/src/lib/documents/documentContext.ts
```diff
@@ -0,0 +1,42 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+// adopted from https://github.com/microsoft/vscode/blob/5ffcfde11d8b1b57634627f5094907789db09776/extensions/css-language-features/server/src/utils/documentContext.ts
+
+import { DocumentContext } from 'vscode-css-languageservice';
```
dummdidumm, line 8: Same question like in `FileSystemProvider`

packages/language-server/src/plugins/css/service.ts
```diff
@@ -59,7 +44,28 @@ export function getLanguage(kind?: string) {
...
 
-export function getLanguageService(kind?: string): LanguageService {
+export type CSSLanguageServices = Record<'css' | 'less' | 'scss', LanguageService>;
+
+export function getLanguageService(langs: CSSLanguageServices, kind?: string): LanguageService {
     const lang = getLanguage(kind);
     return langs[lang];
 }
+
+export function getLanguageServices(options?: LanguageServiceOptions): CSSLanguageServices {
```
dummdidumm, line 54: `createLanguageServices` is a better name I think

dummdidumm (commented):
Nice! Left some minor comments/questions

packages/language-server/src/lib/FileSystemProvider.ts
jasonlyu123: `vscode-html-languageservice` has the same API for path completion. Originally I was thinking maybe we can add it too. But I am not sure is useful inside a svelte file. Since most of the time, the relative path to an asset does work like plain HTML. I think I'll move it to `plugins/css` for now. We could move it out if we want to use it in HTML later.

dummdidumm (approved):
👍

Issue #1529, opened 2022-06-18
Watch tsconfig extends for changes


dummdidumm:
### Description

In SvelteKit, a generated tsconfig lives in `.svelte-kit/tsconfig.json`. This file doesn't exist initially (before you run dev or prepare for the first time) so you see errors in your code. These go away once the extended tsconfig exists, but we don't watch for changes like this, so people have to restart the language server in order to see that.

The quick fix is to handle this specific case, but it would be great if we could generalize this to watch tsconfig files that others extend from and react to changes there.

### Proposed solution

Dig into the TS code and see how they do it, and see if this can be applied at our code, too.

jasonlyu123:
There are two ways I can find to get the extended config file path. 

1. The `configFile` property in `compilerOptions` have the extended config files info. But it's internal. 
2. Or we could leverage the `extendedConfigCache` parameter that can be passed into `parseJsonConfigFileContent`. If we implement the cache, typescript would cache the extended config files into it. 

After we obtain the config file paths, we can watch the files and restart the typescript language service. I think we could use `ts.sys.watchFile` for it. VSCode default to not watching files that are in the gitignore and there's probably not a lot of tsconfig to watch. 

If we are watching the extended config files in the server. We could also change to watch the main tsconfig in the server. It could probably improve the performance a bit.

dummdidumm:
Option 2 sounds good! From the map we get the paths to watch and we also can compare the new config to the old to not restart when the content is the same - I like it! Do you want to implement this or shall I do it?

jasonlyu123:
I can try to implement it this week

PR #1535, opened 2022-06-22
(feat) watch tsconfig and extended tsconfig

jasonlyu123:
#1529

Steps:
1. use a proxy of `extendedConfigCache` to monitor referenced `extendedConfig`
2. watch the tsconfig and the extended config file. 
3. if tsconfig is updated, delete the service.
4. if extended tsconfig is updated, delete the services extending the config file
5. update diagnostic and restart service if the deleted one is still used.

todo:
find a way to test this. currently thinking about adding another property to `LanguageServiceDocumentContext`. To allow mocking the `ts.sys` object so we can more easily test it without needing to `try finally` on the file system

### DIFF
--- a/packages/language-server/src/plugins/typescript/service.ts
+++ b/packages/language-server/src/plugins/typescript/service.ts
@@ -18,6 +18,7 @@ import { ensureRealSvelteFilePath, findTsConfigPath, hasTsExtensions } from './u
 export interface LanguageServiceContainer {
     readonly tsconfigPath: string;
     readonly compilerOptions: ts.CompilerOptions;
+    readonly extendedConfigPaths: Set<string>;
     /**
      * @internal Public for tests only
      */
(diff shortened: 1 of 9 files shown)


packages/language-server/src/plugins/typescript/service.ts
```diff
@@ -91,22 +101,107 @@ export async function getServiceForTsconfig(
...
+async function watchExtendedConfigCallback(
+    extendedConfig: string,
+    kind: ts.FileWatcherEventKind,
+    docContext: LanguageServiceDocumentContext
+) {
+    docContext.extendedConfigCache.delete(extendedConfig);
+
+    extendedConfigToTsConfigPath.get(extendedConfig)?.forEach((config) => {
+        services.delete(config);
+        pendingReloads.add(config);
```
dummdidumm, line 172: I was at first confused why this is enough, until I realized that you don't need to explicitly restart the server because the next `getService` call will do this. Could you add a comment here and in `watchConfigCallback` about this?

dummdidumm:
Is this ready for a final review and merge? It doesn't have tests now but I'm okay with this being untested for now

jasonlyu123:
I added some tests with the virtual `ts.sys`. And also find that I never call the `onProjectReloaded` 😆. It only works because svelte-kit sync also creates `.d.ts` files which trigger `onDidChangeWatchedFiles`

dummdidumm (approved):
Nice!

I feel like we should revisit `service.ts` at some point, it feels like it has grown a little too much over the years, but I don't have a good idea how to refactor this into something that is easier to understand and more testable.

Issue #1094, opened 2021-07-13
VSCode plugin: Syntax highlighting not correctly applied to embedded svelte component in markdown


wlach:
**Describe the bug**
Syntax highlighting not correctly applied to embedded svelte component in markdown when using "Svelte for VSCode"

**To Reproduce**

Put the following inside a markdown document:

````md
```svelte
<script>
  import { onMount } from "svelte";

  export let data = undefined;
  let dom_node;

  onMount(() => {
    Plotly.newPlot(dom_node, data, {barmode: 'stack'});
  });
</script>

<div id="plotDiv" bind:this={dom_node}></div>
```
````
For example a code snippet that is treated in a way you don't expect.

**Expected behavior**

Expected similar syntax highlighting to what I see when viewing a svelte file with that content.

**Actual behaviour**

Get somewhat incorrect / inconsistent formatting (see screenshots)

**Screenshots**

View for a svelte component (correct):

![image](https://user-images.githubusercontent.com/20569/125509203-c043194f-bed0-45b5-b059-5889b8179b59.png)

View for a svelte component embedded in markdown (incorrect):

![image](https://user-images.githubusercontent.com/20569/125509521-334a0e8b-2d73-4be1-8b2d-f00143efe4b8.png)

**System (please complete the following information):**
 - OS: Mac and Windows 10
 - IDE: VSCode
 - Plugin/Package: "Svelte for VSCode"

**Additional context**

**Comment by jasonlyu123** [comment]
Seem like the [injection rules](https://github.com/sveltejs/language-tools/blob/6e0396ca18ea5e7da801468eab35cdef43b3c979/packages/svelte-vscode/syntaxes/svelte.tmLanguage.src.yaml#L7) here doesn't work in the markdown code block for some reason. We'll see whether that's possible to fix.

**Comment by wlach** [comment]
I spent a bit of time today trying to debug this. I used this toy example to test:

````md
```svelte
<script>
  var x = 5;
</script>
```
````

After giving things slightly more unique names, I found out that vscode's textmate parser seems to be "stuck" in this directive when in markdown:

https://github.com/sveltejs/language-tools/blob/6e0396ca18ea5e7da801468eab35cdef43b3c979/packages/svelte-vscode/syntaxes/svelte.tmLanguage.src.yaml#L467

[...]

PR #1537, opened 2022-06-22
(fix) Handle syntax highlighting for most markdown files

wlach:
Mostly fixes_ #1094. Do this by injecting rules to parse typescript and
css for svelte script/style blocks inside markdown only. This
should cover *most* cases of embedded svelte blocks inside
markdown.

It would not be *too* difficult to extend this further to handle other
types of embedded script/style tags (as is done in the original grammar),
though it would be somewhat tedious.

### DIFF
--- a/packages/svelte-vscode/package.json
+++ b/packages/svelte-vscode/package.json
@@ -475,6 +475,20 @@
                 }
             },
             {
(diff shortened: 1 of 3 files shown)


dummdidumm:
Smart solution for the 90% use case - thank you so much!

Issue #1175, opened 2021-09-13
Emmet only works some of the time


kswope:
**Describe the bug**
completions only sporadically available

https://user-images.githubusercontent.com/321586/133140763-313199c7-eee3-419f-9087-cef8ff2b696d.mov

**System (please complete the following information):**
 - OS: mac
 - IDE: VScode
 - Plugin/Package: Svelte for VS Code
 - Vim emulation mode (maybe relevant)

heisenbug was spotted here but never resolved

https://github.com/sveltejs/language-tools/issues/651

jasonlyu123:
I can't reproduce it. Can you try with this config `"svelte.trace.server": "messages"` and post the log in the output channel? It's near the integrated terminal, chose svelte in the dropdown.

kswope:
Starting on a empty buffer this is the output from one successful emmet "div" completion and one missing emmet "div" completion.  Something is not happy.

```
[Trace - 3:07:24 AM] Sending notification '$/setTrace'.
[Trace - 3:07:24 AM] Sending notification 'textDocument/didChange'.
[Trace - 3:07:24 AM] Sending request 'textDocument/completion - (177)'.
[Trace - 3:07:24 AM] Received response 'textDocument/completion - (177)' in 29ms.
[Trace - 3:07:24 AM] Sending request 'completionItem/resolve - (178)'.
[Trace - 3:07:24 AM] Received response 'completionItem/resolve - (178)' in 0ms.
[Trace - 3:07:24 AM] Sending notification 'textDocument/didChange'.
[Trace - 3:07:24 AM] Sending request 'textDocument/completion - (179)'.
[Trace - 3:07:25 AM] Received response 'textDocument/completion - (179)' in 21ms.
[Trace - 3:07:25 AM] Sending request 'completionItem/resolve - (180)'.
[Trace - 3:07:25 AM] Received response 'completionItem/resolve - (180)' in 0ms.
[Trace - 3:07:25 AM] Sending notification 'textDocument/didChange'.
[Trace - 3:07:25 AM] Sending request 'textDocument/completion - (181)'.
[Trace - 3:07:25 AM] Received response 'textDocument/completion - (181)' in 15ms.
[Trace - 3:07:25 AM] Sending request 'completionItem/resolve - (182)'.
[Trace - 3:07:25 AM] Received response 'completionItem/resolve - (182)' in 0ms.
[Trace - 3:07:25 AM] Sending request 'textDocument/codeAction - (183)'.
[...]

PR #1543, opened 2022-07-01
(fix) auto import in workspace without tsconfig/jsconfig

jasonlyu123:
#1175. Not 100% sure it's the problem but the error is the same. It also happens in the [rollup template](https://github.com/sveltejs/template) that what I get the reproduction.  

This one is very tricky to write a test for. So I use a virtual system to write an isolated test. It only happens because of some declaration files that are referenced by the `/// <reference type="">` [triple-slash directives](https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html). And it needs the `node_modules` to be at the root of the workspace. 

What happens is that typescript resolved a module by a relative path to `node_module`. And try to resolve the absolute path to the module so it can proceed with the auto import. But the provided `currentDirectory` is an empty string. I changed it to use the target file path to find the nearest workspace root. So we would always have a current directory.

### DIFF
--- a/packages/language-server/src/plugins/typescript/DocumentSnapshot.ts
+++ b/packages/language-server/src/plugins/typescript/DocumentSnapshot.ts
@@ -96,12 +96,13 @@ export namespace DocumentSnapshot {
     export function fromFilePath(
         filePath: string,
         createDocument: (filePath: string, text: string) => Document,
-        options: SvelteSnapshotOptions
+        options: SvelteSnapshotOptions,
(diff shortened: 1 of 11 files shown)


jasonlyu123:
Yeah, I mean the error. I guess sometimes it triggers an auto import and errors out. Maybe it's just a coincidence that it happens in his attempts. Let's treat it as unrelated then but we could ask him if it improves.

dummdidumm (approved):

PR #1548, opened 2022-07-05
(feat) use SvelteComponentTyped if exists

dummdidumm:
Check in svelteshims if we can use SvelteComponentTyped for more aligned typings

This means invalid accessor calls are no longer caught, which is edge-case enough in my opinion so the benefits outweigh the drawbacks. In Svelte 4 this could be restricted in Svelte. @jasonlyu123 what's your opinion on this?


jasonlyu123 (approved):
I am ok with it not being caught

PR #1549, opened 2022-07-05
(fix) keep colon while transforming events

dummdidumm:
on:click -> "on:click" instead of on:click -> "onclick"

Related to https://github.com/sveltejs/svelte/pull/7649 - @jasonlyu123 this is what you had in mind, right?


jasonlyu123:
Should we keep it extend the `svelte.JSX.DOMAttributes` for a little while? Because this would probably break the custom DOM typedef in the new transformation. Or we can create some migration guide somewhere and do it now?

dummdidumm:
Good point. I think we need some migration guide either way because if we go with `on:click` etc people defining custom events need to add the colon to their type definitions even if we extend from the previous definitions. Damn this situation sucks.. I don't see a way how we can magically find out whether or not to add the colon for a unknown event.

Issue #220, opened 2020-06-21
Property 'files' does not exist on type 'SvelteInputProps'


PatrickG:
**Describe the bug**
I'm getting the following typescript error on an `input` element with `type="file"`.
`Property 'files' does not exist on type 'SvelteInputProps'.`

**To Reproduce**
```html
<script>
  let files;
</script>

<input type="file" bind:files />
```

**Expected behavior**
No error.

dummdidumm:
Fix available in about 22 hours.

martingraf4:
Same problem, this time for the `focusout` event: `Property 'onfocusout' does
not exist on type 'SvelteInputProps'`. 

**To reproduce**

```svelte
<input on:focusout={() => {}} />
```

dummdidumm:
Fix available soon. Next time please open a separate issue. The problem is that [this files](https://github.com/sveltejs/language-tools/blob/master/packages/svelte2tsx/svelte-jsx.d.ts) is incomplete.

PR #1550, opened 2022-07-05
(fix) missing search event

jameslieu:
Hello,

I was getting a TypeScript error when listening to the "search" DOM event. This event is triggered when you click the "X" that appears in an input `type=search` field.

image

Here is an article about the "X" or clear button for more context: https://www.freecodecamp.org/news/targeting-click-of-clear-button-x-on-input-field/

I had a look and there was a similar fix done in this raised issue: https://github.com/sveltejs/language-tools/issues/220
Adding the `onsearch` property here has made the error go away. Hope this helps

### DIFF
--- a/packages/svelte2tsx/svelte-jsx.d.ts
+++ b/packages/svelte2tsx/svelte-jsx.d.ts
@@ -465,6 +465,7 @@ declare namespace svelte.JSX {
       onmouseout?: MouseEventHandler<T> | undefined | null;
       onmouseover?: MouseEventHandler<T> | undefined | null;
       onmouseup?: MouseEventHandler<T> | undefined | null;
+      onsearch?: MouseEventHandler<T> | undefined | null;
 
       // Selection Events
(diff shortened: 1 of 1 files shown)


jasonlyu123:
Thanks for the PR. But according to MDN, this event is non-standard that is not available in some of the browsers. And typescript's `dom.d.ts` also doesn't have this event. So I think we won't add this. 

But if you do want to use it you can enhance the typing by yourself. You can create a `.d.ts` file and make sure this file is [included](https://www.typescriptlang.org/tsconfig#include) in your tsconfig/jsconfig 

```ts
declare namespace svelte.JSX {
    interface DOMAttributes<T extends EventTarget> {
        onsearch?: EventHandler<Event, T> | undefined | null;
    }
}

```

jameslieu:
Thank you for looking at this and your code example to enhance that event myself @jasonlyu123. 

I'll close the PR 👍

Issue #1552, opened 2022-07-06
Roadmap to new transformation


dummdidumm:
The new transformation which transforms Svelte to TS code instead of TSX code for intellisense purposes has been in beta for a while now, and I think it's stable enough to switch to it. This is a breaking change which needs some coordination, so I'm writing down the steps here:

1. [x] We want the new transformation to be more fine grained with respect to the allowed attributes on elements - for example only allow the `href` on `a` tags. At the same time we want to minimize the reliance on `svelte2tsx`. Therefore we need these transformation in Svelte core. https://github.com/sveltejs/svelte/pull/7649 adds these. Also https://github.com/sveltejs/language-tools/pull/1553 for better action typings. (Update: Not merging the HTML PR for now)
2. [x] Try out typings in https://github.com/sveltejs/svelte/pull/7649 to see how we can incorporate them in `svelte2tsx`
3. [x] Test out things with SvelteNative (cc @halfnelson) and Svelte NodeGUI (cc @shirakaba ) - both likely need adjustments (adding the new required functions in the template and adding support for `on:X` events). Depending on the outcome it may be better to bury our hopes of doing `on:X` event typings and instead keep `onX` typings.
4. [x] If experiments succeed, merge https://github.com/sveltejs/svelte/pull/7649 and do a new release of Svelte
5. [x] Do one last release with the old transformation with any bug fixes that may have been merged in the meantime
6. [x] Merge https://github.com/sveltejs/language-tools/pull/1549 , adjust `svelte-jsx.d.ts` so that the old typings are removed - only keep some of the interfaces that people might have used to enhance their typings. Add the typing mentioned in https://github.com/sveltejs/language-tools/pull/1549 so breakage on events is minimal for peole
7. [x] Release a new version of the Svelte extension with the new transformation turned on by default
8. [x] Wait for a few weeks for bug reports or things we might have missed.
9. [ ] Release `svelte-check` 3.0 with the new transformation only. This requires peer dependency bumps for Svelte and TypeScript. We should also bump the node version to latest 14 so we can bump the LSP version under the hood later without it being another breaking change 

@jasonlyu123 what do you think about these steps? Did I miss anything?

Issue #1243, opened 2021-11-16
How do I type custom events for a package?


ghostdevv:
I was wondering what would be the best way to type custom events for a package, puru was recently doing this and now I am too - while you can type the events like this (https://discord.com/channels/457912077277855764/728292755087818924/908761587660255313) I don't know of a way to export those types out of a package and have the language server pick up on that

Any ideas?

dummdidumm:
I guess you talking about custom events in the context of actions. If not, the solution below doesn't apply.

This came up in Discord, too, recently, here is the thread: https://discord.com/channels/457912077277855764/897113912422580274 . It didn't seem to work for that person but I didn't look into why exactly. If you could try the steps and report back if it works that would be great. The steps (roughly):
1. create a `types.d.ts` (or whatever you call it) inside your package, enhance the typings like in your Discord link - make sure to keep this an ambient type, so no top level imports/exports allowed
2. import that file in your `index.d.ts` which is the default types export like `import './types.d.ts';`

If you have your package publicly somewhere, I could take a look.

I also plan on introducing support for typing the events of an action on the action function itself. So basically you would be able to hint at typescript that by applying an action to a DOM element that element now has additional attributes/events that can be passed to it. This depends on changes to Svelte core and a pending rewrite of svelte2tsx, so this will take a little while.

ghostdevv:
I wrote this package as a test, it has a `testAction` which emits a custom event called `test`, I don't get intellisense suggesting that test is a event (like I do with [other events](https://i.imgur.com/1fTpIda.png)) but the actual typings appear to be working which is awesome:

![Screenshot from 2021-11-19 09-39-10](https://user-images.githubusercontent.com/47755378/142600779-9e245555-d9fb-4776-8b30-9127200288f4.jpg)

dummdidumm:
Yes, autocomplete intellisense is missing because that needs some more work on our side to make it show up (not sure how we can get there without showing all other redundant completions).
Good to hear that this works, I guess we need to document the way how to do this once the rewrite happened and we are in a more stable state to officially recommend the best way to do this.

ghostdevv:
Yea would be awesome to be able to do this but glad to know that this does work just missing intellisense - how would you recommend doing this with sveltekit package for example?

ghostdevv:
This was solved a couple years back! https://svelte.dev/docs/svelte/svelte-action#Action

PR #1553, opened 2022-07-06
(feat) enable actions to enhance typings on applied element

dummdidumm:
#1243
#431

This assumes we also enhance the typing of the action type in Svelte core so they match. Assuming we have this:

```ts
export interface ActionReturn<Parameter = any, Attributes extends Record<string, any> = Record<never, any>> {
	update?: (parameter: Parameter) => void;
	destroy?: () => void;
	$$attributes?: Attributes;
}
```

You could define an action like this:

```ts
export function action(node: HTMLElement): ActionReturn<never, { foo: boolean; 'on:bar': EventHandler<Event> }> {
   // ..
}
```

and it would enhance the DOM element with the new required attribute `foo` and the optional `on:bar` event.

This requires newer TS types so we can only merge that right before the new transformation (also see https://github.com/sveltejs/language-tools/issues/1552)

Question (cc @jasonlyu123 @ignatiusmb @ivanhofer ):
Is the assumed `ActionReturn` good like this? Should we put both event handler and attributes into one parameter (people would have to type `onbar` or `on:bar` then (undecided yet what we do here, see other issues)? If not, are the typings for Events good? Should it be `bar: Event` instead and we add the `EventHandler` typing ourselves (but how to get currentTarget of the correct DOM element into that then?)

### DIFF
--- a/packages/svelte2tsx/src/htmlxtojsx_v2/nodes/Element.ts
+++ b/packages/svelte2tsx/src/htmlxtojsx_v2/nodes/Element.ts
@@ -104,10 +104,6 @@ export class Element {
                 break;
             }
             case 'slot': {
-                // If the element is a <slot> tag, create the element with the createSlot-function
-                // which is created inside createRenderFunction.ts to check that the name and attributes
-                // of the slot tag are correct. The check will error if the user defined $$Slots
-                // and the slot definition or its attributes contradict that type definition.
                 this._name = '$$_slot' + this.computeDepth();
                 break;
             }
@@ -248,6 +244,7 @@ export class Element {
             }
         };
 
+        let createElementStatement: TransformationArray;
         switch (this.node.name) {
             // Although not everything that is possible to add to Element
             // is valid on the special svelte elements,
@@ -258,15 +255,8 @@ export class Element {
             case 'svelte:window':
             case 'svelte:body':
             case 'svelte:fragment': {
-                if (!this.referencedName) {
-                    return [`{ ${createElement}("${this.node.name}"${addActions()}, {`];
-                } else {
-                    return [
-                        `{ const ${this._name} = ${createElement}("${
-                            this.node.name
-                        }"${addActions()}, {`
-                    ];
-                }
+                createElementStatement = [`${createElement}("${this.node.name}"${addActions()}, {`];
+                break;
             }
             case 'svelte:element': {
                 const nodeName = this.node.tag
@@ -274,15 +264,8 @@ export class Element {
                         ? ([this.node.tag.start, this.node.tag.end] as [number, number])
                         : `"${this.node.tag}"`
                     : '""';
-                if (!this.referencedName) {
-                    return [`{ ${createElement}(`, nodeName, `${addActions()}, {`];
-                } else {
-                    return [
-                        `{ const ${this._name} = ${createElement}(`,
-                        nodeName,
-                        `${addActions()}, {`
-                    ];
-                }
+                createElementStatement = [`${createElement}(`, nodeName, `${addActions()}, {`];
+                break;
             }
             case 'slot': {
                 // If the element is a <slot> tag, create the element with the createSlot-function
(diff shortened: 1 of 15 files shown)


packages/svelte2tsx/src/htmlxtojsx_v2/nodes/Element.ts
```diff
@@ -244,6 +238,96 @@ export class Element {
...
+        const createElement = `${this.typingsNamespace}.createElement`;
+        const addActions = () => {
+            if (this.actionIdentifiers.length) {
+                return `, __sveltets_2_union(${this.actionIdentifiers.join(',')})`;
+            } else {
+                return '';
+            }
+        };
+
+        switch (this.node.name) {
```
ivanhofer, line 251: From what I can see the `if-else` statements are quite repetitive and error-prone on changes.
I would write it like this:

```ts
const getCreateElementStatement = (name: string): string => {
   // switch-case statement with just the else branches
   // as a single string and not as an array (from what I see it is always 1 element)
   // and without the leading '{'
}

const createElementStatement = getCreateElementStatement(this.node.name)
const variable = this.referencedName ? `const ${this._name} = ` : ''

return [`{ ${variable}${createElementStatement}`]
```

packages/svelte2tsx/test/svelte2tsx/samples/$store-as-directive/expectedv2.ts
```diff
@@ -9,8 +9,8 @@
 ;
 async () => {
 
- { svelteHTML.createElement("div", {      });__sveltets_2_ensureTransition($transitionStore(svelteHTML.mapElementTag('div'),({ y: 100 })));__sveltets_2_ensureAction($actionStore(svelteHTML.mapElementTag('div')));__sveltets_2_ensureTransition($inStore(svelteHTML.mapElementTag('div')));__sveltets_2_ensureTransition($outStore(svelteHTML.mapElementTag('div')));__sveltets_2_ensureAnimation($animateStore(svelteHTML.mapElementTag('div'),__sveltets_2_AnimationMove));
- }};
+ {const $$action_0 = __sveltets_2_ensureAction($actionStore(svelteHTML.mapElementTag('div')));{ svelteHTML.createElement("div", __sveltets_2_union($$action_0), {      });__sveltets_2_ensureTransition($transitionStore(svelteHTML.mapElementTag('div'),({ y: 100 })));__sveltets_2_ensureTransition($inStore(svelteHTML.mapElementTag('div')));__sveltets_2_ensureTransition($outStore(svelteHTML.mapElementTag('div')));__sveltets_2_ensureAnimation($animateStore(svelteHTML.mapElementTag('div'),__sveltets_2_AnimationMove));
```
ivanhofer, line 12: I can only see tests for `$$action_0`. I think a test with multiple actions applied to a dom element would be good.

What happens if two actions expect the same `attribute` with a different type?

ivanhofer (commented):
The idea behind this PR is great!

I didn't dug deep into how the language-tools work so I can't give the best review. I tried to just look at the new additions and the big picture make sense to me.

You mentioned you want to enhance the `ActionReturn` type from `'svelte/action'`. Please also consider enhancing the `Action` type with the same additions.

Because I don't fully know how the language tools work, I'm wondering why a lot of (or all) types get a `__svelte...` name. When I set something a wrong param on an element I sometimes get an error inside VS Code like "x is not assignable to __svelte_xyz" where it actually could say "x is not assignable to Xyz". Is the prefix really needed?

dummdidumm:
Thanks for the review! I wasn't expecting a review of the code (although I greatly appreciate it), more of the general idea and thoughts on the questions I mentioned - if you prefer attributes and events to be split or not etc.

packages/svelte2tsx/test/svelte2tsx/samples/$store-as-directive/expectedv2.ts
dummdidumm: A test with multiple actions is a good idea, yes. If actions conflict then, well, nothing we can do about that 😄 If they types conflict that likely means these actions aren't compatible at runtime, too.

packages/svelte2tsx/src/htmlxtojsx_v2/nodes/Element.ts
dummdidumm: Yeah I was thinking about cleaning this up, too, this is a good outline of the solution

ivanhofer:
> Thanks for the review! I wasn't expecting a review of the code (although I greatly appreciate it), more of the general idea and thoughts on the questions I mentioned - if you prefer attributes and events to be split or not etc.

Sorry, I completely forgot to answer the questions from the description 😅

I prefer to have `Attributes` and `Events` split. If not possible/wanted, I would prefer `on:event` instead of the `jsx`-like `onEvent`, since it is 1:1 what you write on the element when listening to the event. And `onEvent` could also confilct with a prop that is called the same.

The `ActionReturn` interface should probably restrict `Attributes` and `Events` to be `Records`:
```ts
export interface ActionReturn<Parameter = any, Attributes extends Record<string, any> = {}, Events extends Record<string, any> = {}> {
	...
}
```

Defining `Events` should be handled the same way like in `createEventDispatcher`.

> About the prefixes: They exist so it's easier to filter them out of completions, and because this minimizes the probability that they clash with other definitions - they are global/ambient so everyone could use the types by accident.

By accident or as a workaround. And the naming can change. I today realized that [this code](https://github.com/ivanhofer/sveltekit-typescript-showcase/blob/main/src/01-props/09-svelte-component/Component.svelte#L12) does not work anymore because the types have changed in a recent version. (But I can soon switch to the new types from https://github.com/sveltejs/svelte/pull/6770 to get rid of those types)

dummdidumm:
> By accident or as a workaround. And the naming can change. I today realized that [this code](https://github.com/ivanhofer/sveltekit-typescript-showcase/blob/main/src/01-props/09-svelte-component/Component.svelte#L12) does not work anymore because the types have changed in a recent version. (But I can soon switch to the new types from https://github.com/sveltejs/svelte/pull/6770 to get rid of those types)

Yeah that's an unfortunate outcome of these types being global. People come to the idea that they can use that but that wasn't the intention, and then they get broken code after bumps. Maybe as part of the new transformation this all should live in a namespace which we hide from autocompletion.

ignatiusmb:
> Is the assumed `ActionReturn` good like this? Should we put both event handler and attributes into one parameter (people would have to type `onbar` or `on:bar` then (undecided yet what we do here, see other issues)?

Ideally they're specified in one place, I would like to avoid the need to skip a parameter as much as possible. I think we should be able to make it so that properties that starts with `on:` would be treated as events, otherwise an attribute by default.

> are the typings for Events good? Should it be `bar: Event` instead and we add the `EventHandler` typing ourselves (but how to get currentTarget of the correct DOM element into that then?)

I'd say it's probably better to be as explicit as possible, especially when working with TS. More so if it makes the implementation on the language-tools side more complicated.

dummdidumm:
Since the event HTML typings in https://github.com/sveltejs/svelte/pull/7649 are now of the form `on:X` I agree with @ignatiusmb that we should make this one generic where you type both names and events. Will open a PR over at Svelte core shortly.

PR #1566, opened 2022-07-23
(fix) don't ignore .svelte-kit/types in file watch

jasonlyu123:
#1568 For the generated types for svelte-kit. 

Another alternative is to check all the .d.ts files in the ignored build folder. @dummdidumm What do you think is better?


dummdidumm:
Right now it doesn't make a difference, but I like your alternative proposal to instead always add snapshots of d.ts files since we probably want to load them - let's change it 👍

tcc-sejohnson:
Just glanced -- you may want to check the issue I just opened. The type file we're currently generating is `.svelte-kit/ambient.d.ts`. I don't _think_ this update would pick it up.

jasonlyu123:
I updated to check all the declaration files. The reason I proposed the declaration file alternative is it would be more generic so we won't need to update again if kit change the file location. Plus the declaration files usually not that big so it's safer for it to slip.

dummdidumm (approved):

Issue #1558, opened 2022-07-12
Using "useVitePreprocess: true" and "preprocess({ typescript: false })" causes error for type imports.


perfStack:
### Describe the bug

Turning off  `preprocess({ typescript: false, scss: false })` and turning on `experimental: { useVitePreprocess: true, }` causes type imports in .svelte files to throw "Unexpected token" error.

### Reproduction

Turning on `experimental: { useVitePreprocess: true, }`  and turning off  `preprocess({ typescript: false, scss: false })` 

![useExperimental](https://user-images.githubusercontent.com/95343750/178461970-add9d750-c02a-4275-897e-16e44df493e5.jpeg)
![experimental_error](https://user-images.githubusercontent.com/95343750/178462181-d932ef58-02dd-43f4-87a3-5c3d2070abb4.jpeg)

Turning off `experimental: { useVitePreprocess: false }`  and turning on `preprocess({ typescript: true, scss: true })` behaves correctly.

![no_problem_config](https://user-images.githubusercontent.com/95343750/178462801-cf9b186e-cab0-4c87-9d77-0814ff06fab3.jpeg)
![no_error](https://user-images.githubusercontent.com/95343750/178462610-8294f53e-0928-470e-9454-752285d9d803.jpeg)

### Expected behaviour

Expected similar behavior between `preprocess({ typescript: true })`  and `experimental: { useVitePreprocess: true, }` 

### System Info

- OS: Arch linux
- IDE: VSCode - 1.69.0
- Svelte for VS Code - v105.18.1

### Which package is the issue about?

Svelte for VS Code extension

Issue #1479, opened 2022-05-10
Make .svelte files style tag compatible with useVitePreprocess (not only svelte-preprocess)


felixsanz:
### Description

I'm currently using useVitePreprocess and it works wonderful. It's faster because of esbuild, you can use $aliases inside style tags and everything is awesome. Unused CSS also gets removed and classes contains the scoped random symbols. In fact, it produces the same CSS as svelte-preprocess.

The issue is... the vscode extension doesn't handle `<style lang="postcss">` when using useVitePreprocess. It breaks on things like nesting.

I know useVitePreprocess is **experimental**, but can we get this working please?

### Proposed solution

useVitePreprocess able to work with .svelte files

### Alternatives

I can use svelte-preprocess but it duplicates CSS in dev mode. Also.. why use another tool when we are already using vite?

pevey:
When following the instructions at https://tailwindcss.com/docs/guides/sveltekit, the plugin highlights an "error."  I think it may be related to this issue.  

The hello world test page in the instructions use this style block:

`+page.svelte`
```
<h1 class="text-3xl font-bold underline">
	Hello world!
 </h1>
 
<style lang="postcss">
	:global(html) {
	  background-color: theme(colors.gray.100);
	}
 </style>
```

It works fine, but the plugin show a red line under 'colors' along with this info:
```
"")" is expected

If you expect this syntax to work, here are some suggestions: 
If you use less/SCSS with `svelte-preprocess`, did you add `lang="scss"`/`lang="less"` to your `style` tag? If you use SCSS, it may be necessary to add the path to your NODE runtime to the setting `svelte.language-server.runtime`, or use `sass` instead of `node-sass`. 
Did you setup a `svelte.config.js`? 
See https://github.com/sveltejs/language-tools/tree/master/docs#using-with-preprocessors for more info.svelte(css-syntax-error)
```

The config is set up to use vitePrepocess():

`svelte.config.js`
```
import adapter from '@sveltejs/adapter-cloudflare'
import { vitePreprocess } from '@sveltejs/kit/vite'

/** @type {import('@sveltejs/kit').Config} */
const config = {
	kit: {
		// adapter-auto only supports some environments, see https://kit.svelte.dev/docs/adapter-auto for a list.
		// If your environment is not supported or you settled on a specific environment, switch out the adapter.
		// See https://kit.svelte.dev/docs/adapters for more information about adapters.
		adapter: adapter()
	},
	preprocess: vitePreprocess()
}

[...]

Issue #1259, opened 2021-12-03
Cannot find any of modules: sass,node-sass


hyrious:
**Describe the bug**

The same as #622, but I can not get over it.

**To Reproduce**

I didn't use `svelte-preprocess` and/or `svelte.config.js`. I'm using the [@sveltejs/vite-plugin-svelte](https://github.com/sveltejs/vite-plugin-svelte) and let vite handle the preprocessing through [useVitePreprocess](https://github.com/sveltejs/vite-plugin-svelte/blob/main/docs/config.md#usevitepreprocess) option.

In fact, this plugin works well and vite does not complain about it. But vscode does not think so, it throws error on `<style lang="scss">` tag about not founding "sass" or "node-sass" package.

I've created a reproduction repo: https://github.com/hyrious/vite-svelte-sass-bug.

I have tried these steps but no luck:

- add `svelte.language-server.runtime` option
- install `sass` both in the project and in global

**Expected behavior**

The extension should be able to find the sass module.

**Screenshots**

<img width="443" alt="Snipaste_2021-12-03_14-23-39" src="https://user-images.githubusercontent.com/8097890/144555570-c9c98a06-784f-42ac-b3ce-7c80d6d6d362.png">

**System (please complete the following information):**
 - OS: macOS 12.0.1
 - IDE: VSCode
 - Plugin/Package: Svelte for VSCode, svelte-language-server

**Additional context**

hyrious:
Update: the solution seems to be install `svelte-preprocess`.

dummdidumm:
My guess: We try to read a `svelte.config.js` and read the preprocessor from that by looking up the `preprocess` field. If that fails or there is no config (according to your description that's the case here), we fall back to using `svelte-preprocess`. But that package is inside the extension folder, so if it comes across a tag where preprocessing is needed, it'll start searching for SASS (in this example) or other packages starting from the extension folder - and SASS isn't in there, so it fails.
I'm not sure what the best way forward here would be.

jasonlyu123:
Yeah. my assumption is the same. Ideally, it probably should have a `svelte.config.js` in this case. As we can't read the preprocess config from the vite config. Maybe `vite-plugin-svelte` should expose their vite preprocessor so that it could be used in the `svelte.config.js`. But I don't know if they want to make it a public API, it does increase the maintenance burden. 

[...]

PR #1569, opened 2022-07-26
(fix) skip parser error when there's likely another preprocessor involved

dummdidumm:
#1259
#1479
#1558

This may give false negatives for people who do not use Vite/forgot to create a svelte.config.js when they use rollup or webpack, but that's likely OK since we can think about these diagnostics as being opt-in: you add a preprocessor config when you want the Svelte warnings, too. Most people who have this problem also likely use TypeScript or a CSS preprocessor. When using TS, errors are caught through the TS intellisense. When using a CSS preprocessor, the only warning they'll miss is the unused CSS warning, which is apparent anyway since those styles are not visible in the browser.

@jasonlyu123 thoughts?

### DIFF
--- a/packages/language-server/src/plugins/svelte/features/getDiagnostics.ts
+++ b/packages/language-server/src/plugins/svelte/features/getDiagnostics.ts
@@ -91,7 +91,10 @@ async function createParserErrorDiagnostic(error: any, document: Document) {
             document.scriptInfo || document.moduleScriptInfo
         );
 
-        if (!document.config?.preprocess && document.getText().includes('lang="')) {
+        if (
+            (!document.config?.preprocess || document.config.isFallbackConfig) &&
+            document.hasLanguageAttribute()
+        ) {
             Logger.error(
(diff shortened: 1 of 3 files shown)


packages/language-server/src/plugins/svelte/features/getDiagnostics.ts
```diff
@@ -91,6 +91,13 @@ async function createParserErrorDiagnostic(error: any, document: Document) {
             document.scriptInfo || document.moduleScriptInfo
         );
 
+        if (!document.config?.preprocess && document.getText().includes('lang="')) {
```
jasonlyu123, line 94: Would it be better if we use the `TagInformation.attribute.lang` here? if the `defaultLanguages` is set. It should already have the preprocess config so the attribute should accurate?

Also, should we silence the error when the fallback config is used? If so it might need more check as the preprocess config won't be falsy here.

dummdidumm: You are right! Will enhance this.

Issue #1586, opened 2022-08-13
Broke ligatures highlighting with Bracket pair colorization enabled


len0xx:
### Describe the bug

I have Svelte for VSCode extension installed, `'Fira Code', monospace` set as the Font Family and `Bracket Pair Colorization` enabled in VSCode settings. When VSCode updated the last time to 1.70.1 few days ago, the bug showed up. In all of my svelte files in the script tag with lang set to "ts" arrow function syntax highlighting is kind of broken: the first symbol of the arrow (`=`) is colorized while the second (`>`) is not

### Reproduction

1. Set Font Family to `'Fira Code', monospace`
2. Enable `Bracket Pair Colorization`
3. Enable ligatures
4. Create a svelte file
5. Add the following code to the file:
```html
<script lang="ts">
    const a = (id: number) => id + 2
</script>
```

### Expected behaviour

I expect the whole arrow to be the same color as it was before the recent VSCode update

### System Info

- OS: macOS arm64 21.4.0
- IDE: VSCode 1.70.1

### Which package is the issue about?

Svelte for VS Code extension

### Additional Information, eg. Screenshots

![photo_2022-08-13 15 42 16](https://user-images.githubusercontent.com/21990466/184480153-39a05e34-ac47-446a-904a-060e5e7d4bbf.jpeg)

PatrickG:
Seems to be a VSCode problem.
https://github.com/microsoft/vscode/issues/157322

len0xx:
> Seems to be a VSCode problem. [microsoft/vscode#157322](https://github.com/microsoft/vscode/issues/157322)

I was hestitating between reporting it to VSCode repo and this repo. Looks like its a VSCode issue actually. Thank you

PatrickG:
I think this can be closed.
*Does #1593 needs to be reverted?*

jasonlyu123:
It was suggested as a workaround. But when the upstream issue is closed it's still required. Anyway, it's unlikely anything will change so it is the solution now 😆.

PR #1593, opened 2022-08-20
(fix) workaround for unbalanced bracket

jasonlyu123:
For https://github.com/sveltejs/language-tools/issues/1586. According to the workaround from the VSCode team, see https://github.com/johnsoncodehk/volar/issues/1677#issuecomment-1207912785. These highlight scopes are all in the typescript TM language so the workaround should also apply to us. 

 I tried with a VSCode version before this config was supported and it doesn't throw an error so it is probably fine to workaround it for now.


dummdidumm (approved):
Nice! Let's keep the referenced issue open as a reminder to someday remove it once it's fixed in VS Code

PR #1596, opened 2022-08-22
(feat) preserve bind: in new transformation

dummdidumm:
On HTML attributes

@jasonlyu123 this would be needed if we want to switch to explicit `bind:X` typings in the new transformation - does that look correct to you?


jasonlyu123:
Looks good. If we understand it correctly we'll have a similar roadmap for this as the event. So we now don't have to update the definition in `svelte-jsx.d.ts`?

dummdidumm:
Yes, the event PR, this PR, the actions PR and the "remove old JSX typings" PR which is soon to come will all be merged at once after the new Svelte version with the needed changes is released.

Issue #1579, opened 2022-08-08
Arranged imports from code actions misses space in-between


CrendKing:
### Describe the bug

When I use the "Organize Imports" code action on a svelte file, the items in the braces is only delimited by `,`, not `, `.

### Reproduction

1. `import { a, b, d, c } from './Foobar.svelte'`
2. Right click -> Source Action -> Organize Imports
3. It becomes `import { a,b,c,d } from './Foobar.svelte'`

### Expected behaviour

`import { a, b, c, d } from './Foobar.svelte'`

### System Info

- OS: Windows 11
- IDE: VSCode 1.17.0

### Which package is the issue about?

Svelte for VS Code extension

ZerdoX-x:
I guess it should somehow preserve formatting, rather than placing spaces.
Otherwise after this issue being fixed some other guy would open issue about unwanted spaces after executing "Organize imports".

CrendKing:
I saw this repo ([e.g.](https://github.com/sveltejs/language-tools/blob/master/packages/typescript-plugin/src/svelte-sys.ts)) uses space in its imports all the time, so I thought it's some kind of coding style convention in Typescript. To me, it's much more difficult to distinguish each name without the space. But if that's not universally accepted standard, then I guess either won't fix this, or provide a config to satisfy everyone.

PR #1598, opened 2022-08-24
format new text in code action and completion

jasonlyu123:
For #1579. #854  Adding the default format config from typescript and overriding per file it with resolved prettier config. There're some I would like to discuss:

1. Should we read the config from vscode ts/js config? I didn't want to read that because these configs won't be used for formatting. I am afraid it would cause some confusion. Or should we just make it separate and don't use the prettier config?

2. How many of the configs should we detect from the file? The line ending makes sense since by default git on the window would format the line ending. So windows people would want to develop in `CRLF` and have git convert to `LF`. What about semi? Seems like TypeScript would detect it if the config is set to `ignore`. Maybe we should only force it to add or remove in organize-import?

https://github.com/microsoft/TypeScript/blob/394f51aeed80788dca72c6f6a90d1d27886b6972/src/services/utilities.ts#L3411

3. If we go with the prettier config route. Should we use a different default config with TypeScript? Prettier use 2 spaces by default. But VSCode uses 4 by default. 

@dummdidumm what do you think?


packages/language-server/src/plugins/typescript/features/CodeActionsProvider.ts
```diff
@@ -110,29 +109,17 @@ export class CodeActionsProviderImpl implements CodeActionsProvider {
...
             },
             {
-                newLineCharacter: documentUseLf ? '\n' : ts.sys.newLine,
-                semicolons: useSemicolons
-                    ? ts.SemicolonPreference.Insert
-                    : ts.SemicolonPreference.Remove
+                ...(await this.configManager.getFormatCodeSettingsForFile(document)),
+
+                // handle it on our own
+                baseIndentSize: undefined
```
jasonlyu123, line 122: `baseIndent` has some trouble here. Since we don't have indents for import in the svelte2tsx. So it would make it always have organize-import. Think it would make more sense to it on our own.

dummdidumm:
All these different tools wanting to have a say in code formatting, it's messy 😄 I think your approach of lettering prettier taking preference makes sense since that preserves existing behavior. I didn't fully understand what do you mean by "I don't want to use the VS Code TS/JS settings" - what would be the problem exactly? If someone has Prettier turned off, these settings would have impact since they are not obsolete after formatting (if that's what you meant).

jasonlyu123:
Did you mean we'll always use the ts/js format config? Or do we provide a config or a condition for using the ts/js config?

dummdidumm:
What I imagined was:
1. Try to load VS Code ts/js format config
2. Overwrite specific parts with what we know from the Prettier config (the semi and line breaks)
3. Fall back to defaults (the ones you listed in that PR already)

packages/language-server/src/ls-config.ts
```diff
@@ -420,6 +432,93 @@ export class LSConfigManager {
...
+                : 'javascript';
+
+        const tsFormatCodeOptions = this.tsFormatCodeOptions[configLang];
+
+        if (!filePath) {
+            return tsFormatCodeOptions;
+        }
+
+        const prettierConfig = this.getMergedPrettierConfig(
+            await importPrettier(filePath).resolveConfig(filePath, {
```
jasonlyu123, line 498: I wonder if we could disable the "using prettier 2.x.x from path" logging here. It's kind of disturbing. Als it runs on completion which triggers quite often it might hurt performance.

dummdidumm: Yeah that makes sense

dummdidumm (commented):
Looks good! I agree that we should remove the logging in this case - maybe we should remove it altogether, I can't remember when this was useful last time (it was in the early days, but not anymore). Thoughts on that?

Another thought: We query `prettier` on every autocompletion now, I wonder if this has any performance impact. I _think_ we should be fine since the dynamic import is only slow the first time and cached later on, same for resolving.

jasonlyu123:
Do you mean remove all version logging or only prettier? Maybe we could add a `logger.debug` and config to enable it. And we can remove all the version logging. 

About import prettier. I benchmark it on my machine. Logging the first time, then 1000 times with log and 1000 times without log. The cold start run is about 110ms. 1000x without logging in 150-170ms. 1000x with log is about 10% slower. So yeah, it's only slow on the first one. it's fast enough for later calls.

dummdidumm:
I like the `Logger.debug` + config idea. Name could be `svelte.language-server.debug = true/false` which is a generic enough name that we could add other things to it that are present in debug mode only.

packages/language-server/src/logger.ts
```diff
@@ -13,4 +19,10 @@ export class Logger {
     static error(...args: any) {
         console.error(...args);
     }
+
+    static debug(...args: any) {
+        if (!Logger.logErrorsOnly && this.logDebug) {
```
dummdidumm, line 24:
```suggestion
        if (!Logger.logErrorsOnly && Logger.logDebug) {
```

packages/language-server/src/plugins/typescript/service.ts
```diff
@@ -180,6 +180,7 @@ async function createLanguageService(
     let projectVersion = 0;
 
     const host: ts.LanguageServiceHost = {
+        log: (message) => Logger.debug(`[ts] ${message}`),
```
dummdidumm, line 183: Oh interesting 😄

dummdidumm (approved):

Issue #1604, opened 2022-08-26
Reactive, destructured declarations aren't recognised by TypeScript in certain combinations


Sxxov:
### Describe the bug

When writing reactive, destructured declarations (using the `$:` directive), with certain combinations of renames & default initialisers, TypeScript would be confused into thinking it's a variable assignment expression instead of a declaration statement.

### Reproduction

```svelte
<script lang="ts">
	// 1. OK  → $: ({ a = 1, c: d = 2 } = { a: 0, c: 0 });
	// 2. OK  → $: ({ a: b = 1, c: d } = { a: 0, c: 0 });
	// 3. OK  → $: ({ a: b = 1, c = 2 } = { a: 0, c: 0 });
	// 4. OK  → $: ({ a: b = 1, c: d, e: f = 2 } = { a: 0, c: 0, e: 0 });
	// 5. NOK → $: ({ a: b = 1 } = { a: 0 });
	// 6. NOK → $: ({ a: b = 1, c: d = 2 } = { a: 0, c: 0 });

	// Using no. 5 as a runnable example that fails:
	$: ({ a: b = 1 } = { a: 0 });
	//       ^ Cannot find name 'b'. ts(2304)
</script>
```

### Expected behaviour

```svelte
<script lang="ts">
	// 1. OK  → $: ({ a = 1, c: d = 2 } = { a: 0, c: 0 });
	// 2. OK  → $: ({ a: b = 1, c: d } = { a: 0, c: 0 });
	// 3. OK  → $: ({ a: b = 1, c = 2 } = { a: 0, c: 0 });
	// 4. OK  → $: ({ a: b = 1, c: d, e: f = 2 } = { a: 0, c: 0, e: 0 });
	// 5. OK  → $: ({ a: b = 1 } = { a: 0 });
	// 6. OK  → $: ({ a: b = 1, c: d = 2 } = { a: 0, c: 0 });

	// Using no. 5 as a runnable example that shouldn't fail:
	$: ({ a: b = 1 } = { a: 0 });
	//       ^ let b: number
</script>
```

### System Info

- OS: Windows 11 21H2
- IDE: VSCode 1.70.1

### Which package is the issue about?

svelte-language-server

### Additional Information, eg. Screenshots

<img width="391" alt="image" src="https://user-images.githubusercontent.com/50891910/186955657-458596c4-61cf-48ef-b6d7-e26bb18fe0ea.png">

PR #1605, opened 2022-08-27
(fix) extract name in destructuring with default

jasonlyu123:
#1604 Missed one spot when we adopt `periscopic` for typescript AST.  

There's still one left that's not adopted.

https://github.com/Rich-Harris/periscopic/blob/2ff43ecd3e60a660acce552046d5c6eac0586ae7/src/index.js#L230

 Still not sure what that's for.  Parameter like `(...args)` ? Not sure if our use case would use that.


dummdidumm (approved):
Not sure either, but doesn't seem related to our use case, so let's ignore it for now

PR #1606, opened 2022-08-28
(feat) update new typings

dummdidumm:
to conform to new on:X and bind:X transformations and to prepare the transition to official html typings.

@jasonlyu123 FYI the idea is to test out the new transformation changes the next few days, then flip the switch so everyone get the new transformation, gather feedback, and then publish the typings somewhere official. Anything I may have overlooked as part of this plan?


jasonlyu123:
Looks good. Can't think of any problem with the plan and code-wise at the moment. But I have an idea to enhance migration. What do you think about we add an information message to the extension? To notify the changes and linking to the feedback or the roadmap thread.

PR #1620, opened 2022-09-04
(feat) create sveltekit route files

david-plugge:
- [x] create a single page/layout/error/endpoint
- [x] create multiple files at once
- [x] enter url path for the file location
- [x] use `WorkspaceEdit` instead of `fs`
- [ ] ~~find out if this is really a SvelteKit app (check package.json)~~
- [ ] ask if a [matcher](https://kit.svelte.dev/docs/advanced-routing#matching) should be created if found in the path and does not exist
- [ ] only show the explorer menu in the routes folder (don´t know if this is possible, or needed)
- [ ] read svelte.config.js because people can customize the routes folder (?) (maybe later, not sure how feasible this is since we need to read a JS file)


dummdidumm:
From reading the bullet point list it sounds like we only need to do the `package.json` check, or maybe not even that because it's a dedicated command someone actively needs to invoke, and that someone likely doesn't do that in a non-Kit app. I'm ok with doing the other stuff later. Thoughts?

david-plugge:
I totally agree, it just adds unnecessary complexity.

david-plugge:
I also added some snippets for `RequestHandler` and `Load` functions.
Is it fine to include them in this PR?
Or do you think we should not add any kit related snippets?

packages/svelte-vscode/snippets/svelte.json
```diff
@@ -1,22 +1,27 @@
...
+    "each block": {
+        "prefix": "each",
+        "description": "each block",
+        "body": ["{#each ${1:array} as ${2:item}}", "\t$TM_SELECTED_TEXT$0", "{/each}"]
     },
-    "key": {
-        "body": ["{#key ${1:key}}", "\t$TM_SELECTED_TEXT$0", "{/key}"],
-        "description": "key block"
+    "key block": {
+        "prefix": "key",
```
dummdidumm, line 23: Prefixes were deliberately left out in #1373 so they don't conflict with the language server completions. It sounds like you were wondering why they didn't show up, that's why. Could you revert this change?

dummdidumm:
Apart from the snippet thing (prefix was omitted deliberatly) this seems like it's ready to review?

packages/svelte-vscode/snippets/svelte.json
david-plugge: Does this only affect the svelte snippets? The js and ts snippets don´t work for me without the prefix

dummdidumm: yes, only Svelte snippets

dummdidumm (approved):
Awesome work! Fixed one bug which made it fail on Windows, but that's all. Thank you so much for working on this 🎉

PR #1623, opened 2022-09-06
(fix) user defined prettier plugins gets overridden by svelte plugin

ni-max:
so the following config will not work and language server will use it's `prettier-plugin-svelte`
```javascript
// prettier.config.cjs
module.exports = {
    plugins: [require('prettier-plugin-tailwindcss')]
}
```


dummdidumm (approved):
Thank you!

PR #1629, opened 2022-09-12
fix `getCompletionEntryDetails` crash

zardoy:
By the type `data` can be `undefined`, unfortunately due to limited time, I can't provide exact str of to get `undefined` in `data`, most probably some another plugin causing it. Anyway, casting possibly undefined value to `any` usually a bad idea. Thank you.

#1630


zardoy:
str is extremely simple: `"svelte.enable-ts-plugin": true`

Issue #1645, opened 2022-09-18
Cannot read properties of undefined (reading 'debug') [emacs-lsp]


gustavotcabral:
### Describe the bug

The language server is failing on initialization when used with the emacs-lsp client: _"Cannot read properties of undefined (reading 'debug')"_

I fixed it using optional chaining (`?.`) when accessing the `.debug` attribute (file `server.ts`):

`... .['language-server']?.debug`

### Reproduction

Simple initialization.

### Expected behaviour

The language server successfully initializes.

### System Info

- OS: Ubuntu Bionic 18.04
- IDE: GNU Emacs 28.1 + Emacs LSP

### Which package is the issue about?

svelte-language-server

gustavotcabral:
Hi, @jasonlyu123 

I think you need to fix one more occurrence of `.debug`.

ohmree:
Since the fix for this doesn't seem to have landed in a release yet I made a `PKGBUILD` for Arch users that installs the latest version from git:

```bash
_pkgname=svelte-language-server
pkgname="nodejs-${_pkgname}-git"
pkgver=extensions.106.1.0.r8.g5ebd8992
pkgrel=1
pkgdesc='A language server (implementing the language server protocol) for Svelte.'
arch=('any')
url='https://github.com/sveltejs/language-tools'
license=('MIT')
conflicts=('nodejs-svelte-language-server')
depends=('nodejs>=12')
makedepends=('npm>=8' 'jq' 'git')
source=("${pkgname}::git+${url}.git#branch=master")
sha256sums=('SKIP')

pkgver() {
  cd "$pkgname"
  git describe --long --tags | sed 's/\([^-]*-g\)/r\1/;s/-/./g'
}

build() {
  cd "$pkgname"
  npm install --cache="${srcdir}/npm-cache"
  npm run build --workspace="${_pkgname}"
}

package() {
  cd "$pkgname"
  # NOTE: taken from here:
  # https://github.com/yochananmarqos/pkgbuilds/blob/master/pnpm/PKGBUILD#L27
  local tmppackage="$(mktemp)"
  local npmdir="packages/${_pkgname#svelte-}"
  local pkgjson="$npmdir/package.json"
  jq ".version = \"$pkgver\"" "$pkgjson" > "$tmppackage"
  mv "$tmppackage" "$pkgjson"
  chmod 644 "$pkgjson"
  npm pack --workspace="${_pkgname}"
[...]

PR #1646, opened 2022-09-18
(fix) update config null check

jasonlyu123:
#1645 

@gustavotcabral is this ok? Is there any other I missed?


gustavotcabral:
> #1645
> 
> @gustavotcabral is this ok? Is there any other I missed?

Seems right to me. I'm not aware of the internals of svelte-ls, though.

PR #1666, opened 2022-09-29
fix `prependData` esm example

ivanhofer:


jasonlyu123:
thanks! I found that cjs example also has a problem so I take the chance to fix it as well.

PR #1679, opened 2022-10-11
fix(language-server): replace deprecated package `vscode-emmet-helper` in favor of `@vscode/emmet-helper`

Some-Dood:
The `vscode-emmet-helper` package has long been deprecated in favor of the renamed `@vscode/emmet-helper`. This causes NPM to emit warnings when installing the `svelte-language-server` package. This PR fixes that issue.

### DIFF
--- a/packages/language-server/package.json
+++ b/packages/language-server/package.json
@@ -48,6 +48,7 @@
     },
     "dependencies": {
         "@jridgewell/trace-mapping": "^0.3.9",
+        "@vscode/emmet-helper": "^2.8.4",
         "chokidar": "^3.4.1",
         "estree-walker": "^2.0.1",
         "fast-glob": "^3.2.7",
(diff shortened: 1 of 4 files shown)


packages/language-server/test/plugins/css/CSSPlugin.test.ts
```diff
@@ -324,6 +324,22 @@ describe('CSS Plugin', () => {
...
+                                line: 0
+                            },
+                            start: {
+                                character: 17,
+                                line: 0
+                            }
+                        },
+                        newText: 'hwb(240 0% -25400%)'
+                    }
                 }
```
Some-Dood, line 343: This updated test case comes from the draft PR #1534. I simply adapted it here to make the tests succeed on my local installation. Please do let me know if this is an issue.

dummdidumm: That test was needed due to an update of another library in the referenced PR, you need to remove it from this PR else the tests fail.

Some-Dood: Ah, I see. I have since deleted that commit. 👍

dummdidumm (approved):

Issue #1688, opened 2022-10-18
Properties with generic type of a generic component have incorrect type


7nik:
### Describe the bug

Generic types in properties of an instance of a generic component always have the type passed into `$$Generic<>` or `unknown` if nothing was passed.

### Reproduction

Test.svelte:
```ts
<script lang="ts">
    type T = $$Generic;

    export function method(arg: T): T {
        return arg;
    }
    export const method2 = (arg: T): T => arg;

    export let prop: T | null = null;
</script>

<svelte:options accessors />

Test component {prop}
```
another component
```ts
<script lang="ts">
    import type Test from "./Test.svelte";

    let test: Test<number>;

    $: if (test) {
        const res = test.method(true); // type is unknown
        const res2 = test.method2(true); // type is unknown
        const prop = test.prop; // type is unknown
    }
</script>
```
The `method` and `method2` have type `(property) Test__SvelteComponent_<number>.method: (arg: unknown) => unknown` and `prop` have type `(property) Test__SvelteComponent_<number>.prop: unknown`.

### Expected behaviour

The `method` and `method2` have type `(arg: number) => number` and thus `res` and `res2` get type `number. `prop` have type `number | null`.

### System Info

Code - OSS
Version: 1.72.0
Commit: 64bbfbf67ada9953918d72e1df2f4d8e537d340e
Date: 2022-10-06T22:56:30.304Z
Electron: 19.0.13
Chromium: 102.0.5005.167
Node.js: 16.14.2
V8: 10.2.154.15-electron.0
OS: Linux x64 5.4.217-1-MANJARO
Sandboxed: No

Svelte for VS Code: v106.2.0

PR #1689, opened 2022-10-18
(fix) generic support for component getter and accessor

jasonlyu123:
#1688


dummdidumm:
Do you know why we don't transform getters by doing `get foo() { return this.$$prop_def.foo }`? This should work the same for components with and without generics.

dummdidumm:
Mhm, maybe (I thought $$prop_def always existed, but my memory might be tricking me 😅 ). Either way, if there's nothing speaking against it, I'd say it's probably better to transform it that way now, which would make special code for the generic case unnecessary. What do you think?

dummdidumm:
mhhm I think we can probably remove `getters` from the transpiled output then. We can't remove `this.getters` from `ExportedNames` though because that's used to distinguish between props that are writeable (`export let ..`) and those that are only readable (`export const/function ..`)

dummdidumm:
Nice one! Is this ready to merge?

jasonlyu123:
It's ready. Any other thing you want to change?

dummdidumm (approved):

Issue #1627, opened 2022-09-09
weird syntax hilight


KryptoPX:
### Describe the bug

Syntax hilight gets weird in some case when import a ts file

### Reproduction

https://github.com/sveltejs/language-tools/issues/1627#issuecomment-1257347686

### Expected behaviour

Normal syntax hilight, i gess?

### System Info

- OS: Windows
- IDE: VSCode

### Which package is the issue about?

Svelte for VS Code extension

### Additional Information, eg. Screenshots

![](https://user-images.githubusercontent.com/23123160/192175864-30fb1882-1fc1-4cf0-ae02-dba5633d239d.gif)

### Screenshots

![image](https://user-images.githubusercontent.com/79398439/189375488-1347efde-1500-4480-bf4a-e98130d4ae07.png)
![image](https://user-images.githubusercontent.com/79398439/189375663-4828007a-c11f-4139-ab3f-300d18305e20.png)
![image](https://user-images.githubusercontent.com/79398439/189375993-843f5baf-114a-417d-b4bc-7ff5da44b61c.png)
![image](https://user-images.githubusercontent.com/79398439/189377726-5c5a0edc-4922-4063-bd43-6982d4cecc21.png)

jasonlyu123:
Does this only happen in a specific file? Looks like something is wrong with source mapping to the generated code under the hood. We'll still need a reproducible to work on it. What do mean by "random moment"? Does it only happen during editing and it wouldn't persist if you restart the editor?

If it persists after you restart the editor. Can you copy the file to a new project and see if it still happens? Also, we'll need something we can copy and paste instead of just a screenshot.

@JoDo-dev @falco467 Is this what you're seeing?

KryptoPX:
Update... Maybe isn't refreshing info when i put some text or an enter

When i start VSCode:
![image](https://user-images.githubusercontent.com/79398439/189493040-46caf398-2f1a-44a3-a70f-04761c091b6b.png)

later when i add a new line:
![image](https://user-images.githubusercontent.com/79398439/189493079-cdd86ffd-4a52-4df1-aba2-9bab1b71b7da.png)

jasonlyu123:
Does the error disappear after restart? I want to know if it's related to how we update the file content stored in the memory. Also, please try removing some code to find a minimal reproduction. In text form, not a screenshot. We need more information to be able to help you.

dummdidumm:
[...]

PR #1697, opened 2022-10-21
(fix) case insenstive file system document sync

jasonlyu123:
#1627 Adopted the casing conversion helper function from typescript. And added two collections for tracking files case intensively in case-insensitive file systems. Another goal is to also address the famous "files differ only in casing" typescript error when it's VSCode using the old casing before renaming.

The strategy is that we store the lowercase filenames in the collection. And when retrieving or checking the existence of files. We convert it to lowercase before checking. And we keep the original casing in another place so we can preserve the casing. Just not sure if we need to keep track of multiple. Not sure if we should reach the filesystem to check the correct casing if there're multiple exists. Because the only way I can think of is to use the read directory of the parent directory. Not sure if it's worth the effort.  

Created as a draft to discuss if the overall strategy has any flaws. Also need to scan through the codebase to check for any filename equality check. We might also need to normalize the casing before checking.


packages/language-server/src/utils.ts
```diff
@@ -289,3 +289,28 @@ export function returnObjectIfHasKeys<T>(obj: T | undefined): T | undefined {
         return obj;
     }
 }
+
+// eslint-disable-next-line no-useless-escape
+const fileNameLowerCaseRegExp = /[^\u0130\u0131\u00DFa-z0-9\\/:\-_\. ]+/g;
+
+/**
+ * adopted from https://github.com/microsoft/TypeScript/blob/8192d550496d884263e292488e325ae96893dc78/src/compiler/core.ts#L1803
+ * see the comment there about why we can't just use String.prototype.toLowerCase() here
```
dummdidumm, line 298: there's no comment 😄

dummdidumm (commented):
This makes sense to me! About "so we can preserve the casing" -> what for exactly? If the system is case insensitive by definition there can't be multiple files which only are different in casing.

jasonlyu123:
Keeping the original is so that the hover info of a module path would show the original casing. And also that the URI converted from the typescript response is the original casing. It probably only happens in cross-document features so it probably only in typescript features. 

It's mostly the non-project files that would have the difference. Because the casing typescript uses is from the `getScriptFileNames` and I currently prioritize the casing from project files. 

About the multiple, originally I was thinking it might be more compatible with the old code. But maybe that is just hiding potential bugs. We should convert the casing before comparing. 

After writing these. I think I can remove the original casing from the `FileMap` and only keep the original casing in the `DocumentSnapshot` and change the `getScriptFileNames` to use the `DocumentSnapshot.filePath`. If we found any non-typescript features needing the filePath casing we can use `Document.getFilePath()` or `Document.uri` to track the original casing.

packages/language-server/test/plugins/typescript/features/UpdateImportsProvider.test.ts
```diff
@@ -42,21 +44,52 @@ describe('UpdateImportsProviderImpl', () => {
...
-        const { updateImportsProvider, fileUri } = await setup('updateimports.svelte');
+        const { updateImportsProvider, fileUri } = await setup(
+            'updateimports.svelte',
+            ts.sys.useCaseSensitiveFileNames
+        );
 
         const workspaceEdit = await updateImportsProvider.updateImports({
-            // imported files both old and new have to actually exist, so we just use some other test files
-            oldUri: pathToUrl(join(testFilesDir, 'diagnostics', 'diagnostics.svelte')),
+            oldUri: pathToUrl(join(testFilesDir, 'imported.svelte')),
```
jasonlyu123, line 53: This comment seems to be outdated. It now seems to be able to check even if the new file does not exist.

jasonlyu123:
I have scanned through the codebase for the `Map<string, T>`, `Set<string>`, `path` and `file` keywords. It's ready for review.

packages/language-server/src/plugins/typescript/SnapshotManager.ts
```diff
@@ -103,12 +110,19 @@ export class SnapshotManager {
...
         private fileSpec: TsFilesSpec,
-        private workspaceRoot: string
+        private workspaceRoot: string,
+        projectFiles: string[],
+        useCaseSensitiveFileNames = ts.sys.useCaseSensitiveFileNames
     ) {
         this.onSnapshotChange = this.onSnapshotChange.bind(this);
         this.globalSnapshotsManager.onChange(this.onSnapshotChange);
+        this.documents = new FileMap(useCaseSensitiveFileNames);
+        this.projectFileToOriginalCasing = new FileMap(useCaseSensitiveFileNames);
```
dummdidumm, line 121: Not quite understand the use case if this map here - it's not set with `original,case_insensitive` or similar, it's set with `originalFileName, originalFileName` in all places where it's called

packages/language-server/src/plugins/typescript/service.ts
```diff
@@ -318,6 +325,27 @@ async function createLanguageService(
...
+
+        return Array.from(
+            new Set([
+                ...projectFiles,
+                // project file is read from the file system so it's more likely to have
+                // the correct casing
+                ...snapshotManager
+                    .getFileNames()
+                    .filter(
+                        (file) => !canonicalProjectFileNames.includes(getCanonicalFileName(file))
```
dummdidumm, line 342: I believe this method is called rather often, and this would make it O(n^2) - can we use a `Set` instead?

dummdidumm (commented):
Mostly looks good, left some comments. The thing I'm most concerned about (but a that regex is likely very fast) is the cost of converting the file names that often - i.e. if it has a noteable impact on performance.

packages/language-server/src/plugins/typescript/SnapshotManager.ts
jasonlyu123: It's essentially `Map<case_insensitive, original>`. I think I can change it to directly use `Map<case_insensitive, original>` so it's more clear on its use case.

dummdidumm: What I meant is that everywhere where `set` is invoked you call it with the same variable for both parameters

jasonlyu123: Yes. Because the `FileMap` will transform the casing for the key but not the value. That is indeed confusing as it keeps two different things.

dummdidumm (approved):
Thank you so much for implementing this!

PR #1715, opened 2022-11-07
svelte-check: Add --preserveWatchOutput, which disables clearing the terminal when using --watch

odinhb:
Context:
I am trying to add svelte to our application, which has previously been using `tsc` to typecheck our codebase, (and `esbuild` to build).

`tsc` and `svelte-check` were both annoying when run from a process manager, such as [overmind](https://github.com/DarthSim/overmind), which runs multiple cli tools/watchers etc in the same terminal. This is because they by default clear the screen periodically, which blows away useful information from other tools.

`tsc` supports `--preserveWatchOutput`, ([link](https://www.typescriptlang.org/tsconfig/preserveWatchOutput.html)) so I added the same flag to svelte-check.


dummdidumm (approved):
Thank you!

dummdidumm:
Could you run `yarn format` at the root of the monorepo and push the formatting changes so CI is green?

odinhb:
@dummdidumm Is it too much to ask for a minor release for this? I'd like to stop using my fork.

Issue #1362, opened 2022-02-01
`svelte-check --version` is not correct


rgossiaux:
**Describe the bug**
If you run `svelte-check --version` (eg `npx svelte-check --version`, or in a SvelteKit project `npm run check -- --version`) you get as output: `svelte-check, 1.x`

The 1.x isn't very useful anyway, but it's also incorrect (we're now on 2.x). It comes from this hardcoded line: https://github.com/sveltejs/language-tools/blob/b96008427b5bd52e1de0156cebc55b32c127df38/packages/svelte-check/src/options.ts#L22

On a separate but related note, I noticed that the version in `package.json` is not kept up to date by your release process: https://github.com/sveltejs/language-tools/blob/b96008427b5bd52e1de0156cebc55b32c127df38/packages/svelte-check/package.json#L4 

**To Reproduce**
See above.

**Expected behavior**
See above.

**Screenshots**
N/A

**System (please complete the following information):**
N/A

**Additional context**
N/A

PR #1747, opened 2022-11-30
full version number in `svelte-check --version`

### LINKED ISSUE #1746: `svelte-check --version` should output the full version number
### Description

```
> svelte-check --version
svelte-check, 2.x
```

### Proposed solution

it should output the full version, like
```
svelte-check, 2.9.2
```

### Alternatives

_No response_

_No response_

DetachHead:
#1746, #1362

### DIFF
--- a/packages/svelte-check/src/options.ts
+++ b/packages/svelte-check/src/options.ts
@@ -20,7 +20,7 @@ export interface SvelteCheckCliOptions {
 // eslint-disable max-len
 export function parseOptions(cb: (opts: SvelteCheckCliOptions) => any) {
     const prog = sade('svelte-check', true)
-        .version('2.x')
+        .version(require('../../package.json').version)
         .option(
             '--workspace',
             'Path to your workspace. All subdirectories except node_modules and those listed in `--ignore` are checked'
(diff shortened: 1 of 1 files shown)


packages/svelte-check/src/options.ts
```diff
@@ -20,7 +20,7 @@ export interface SvelteCheckCliOptions {
 // eslint-disable max-len
 export function parseOptions(cb: (opts: SvelteCheckCliOptions) => any) {
     const prog = sade('svelte-check', true)
-        .version('2.x')
+        .version(require('../../package.json').version)
```
dummdidumm, line 23:
```suggestion
        .version(require('../../package.json').version) // ends up in dist/src, that's why we go two levels up
```

dummdidumm (approved):
Thank you!

packages/svelte-check/src/options.ts
```diff
@@ -20,7 +20,7 @@ export interface SvelteCheckCliOptions {
 // eslint-disable max-len
 export function parseOptions(cb: (opts: SvelteCheckCliOptions) => any) {
     const prog = sade('svelte-check', true)
-        .version('2.x')
+        .version(require('../../package.json').version) // ends up in dist/src, that's why we go two levels up
```
dummdidumm, line 23:
```suggestion
        // eslint-disable-next-line @typescript-eslint/no-var-requires
        .version(require('../../package.json').version) // ends up in dist/src, that's why we go two levels up
```

PR #1749, opened 2022-12-01
(fix): add "off" to `data-sveltekit-noscroll` and `data-sveltekit-reload` props definitions

oekazuma:
The `data-sveltekit-noscroll` and `data-sveltekit-reload` props should be allowed to be "off".

Definition of `data-sveltekit-noscroll`
https://github.com/sveltejs/kit/blob/e2c74b7278ed75de681dfbbe84969c6cc318660c/packages/kit/src/runtime/client/utils.js#L34

Definition of `data-sveltekit-reload`
https://github.com/sveltejs/kit/blob/e2c74b7278ed75de681dfbbe84969c6cc318660c/packages/kit/src/runtime/client/utils.js#L35


dummdidumm (approved):
Thank you!

PR #1770, opened 2022-12-09
(feat) use satisfies operator if possible

dummdidumm:
when creating SvelteKit files


aradalvand:
@dummdidumm What was the reason behind this? Why is the `satisifes` operator preferable?!

dummdidumm:
The advantage is that for the generated `./$types` we have to generate the types only once and not regenerate them if you update your load function. Also once we persuade the TypeScript team to implement this for functions, too (https://github.com/microsoft/TypeScript/issues/51556), it will look way nicer.
```ts
export function load (..) {
  ..
} satisfied PageLoad
```

aradalvand:
> The advantage is that for the generated ./$types we have to generate the types only once and not regenerate them if you update your load function

Ah okay, that makes sense, thanks.

> Also once we persuade the TypeScript team to implement this for functions, too (https://github.com/microsoft/TypeScript/issues/51556), it will look way nicer.

Yeah that would make it look far better. Too many paratheses now!

Thanks.

Issue #1664, opened 2022-09-27
SvelteKit commands should only be shown in SvelteKit projects


tony-sull:
### Describe the bug

The new `SvelteKit: Create route` menus are being shown in every VS Code project when the Svelte for VS Code plugin is enabled.  Any chance this can be scoped down to only show when in a SvelteKit project, maybe by checking for a `svelte.config.js` file?

### Reproduction

Open any project in VS Code, make sure the Svelte for VS Code extension is enabled, and right click on a directory in the project explorer

### Expected behaviour

I'd expect the SvelteKit-specific menus to only show when a SvelteKit project is open

I'd really expect this in a separate VS Code extension since it is specific to SvelteKit rather than Svelte, but I'm sure that would lead to a pile of refactoring work or code duplication 🤷 

### System Info

- OS: OS X
- IDE: VS Code

### Which package is the issue about?

Svelte for VS Code extension

### Additional Information, eg. Screenshots

<img width="600" alt="Screen Shot 2022-09-27 at 3 06 33 PM" src="https://user-images.githubusercontent.com/15836226/192627522-14a6902e-09e0-4609-b6a5-e2943d5d7f84.png">

tony-sull:
Looks like the plugin adds commands to the Command Palette and TypeScript's autocomplete as well, it'd be ideal to see this only added conditionally in a SvelteKit project

Command Palette opened in a non-SvelteKit project
<img width="735" alt="Screen Shot 2022-09-27 at 3 28 20 PM" src="https://user-images.githubusercontent.com/15836226/192631319-202b3258-8450-48cc-8a3f-7e25cbaf99fa.png">

Auto-complete in Astro frontmatter (happens in any Typescript code)
<img width="968" alt="Screen Shot 2022-09-27 at 3 29 33 PM" src="https://user-images.githubusercontent.com/15836226/192631460-dcbd082b-6839-4395-89bf-34f6f558696d.png">

dummdidumm:
I'm not sure this is possible. AFAIK VS Code only makes it possible to provide these things through the static package.json, so there's no way to dynamically turn it on or off.
I also think it's not too much of a problem. There are already hundreds of snippets/completions and commands, a few more don't change that. The only thing that _could_ get annoying is the menu entry.

tony-sull:
[...]

PR #1771, opened 2022-12-10
(feat) only show SvelteKit files context menu in kit project

jasonlyu123:
#1664 

Also adding a config to force enable and disable. One downside is that the command only shows until the extension is activated. i.e. any svelte, js, ts files are opened. That is also why there is a config to force it open. so it would be available before the extension is activated.

### DIFF
--- a/packages/svelte-vscode/package.json
+++ b/packages/svelte-vscode/package.json
@@ -115,7 +115,7 @@
                         "never",
                         "always"
                     ],
-                    "description": "Show a context menu to generate SvelteKit files. \"always\" to always show it. \"never\" to always disable it. And \"auto\" to only opened in a SvelteKit project. "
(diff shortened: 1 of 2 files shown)


packages/svelte-vscode/package.json
```diff
@@ -637,7 +647,7 @@
                     "group": "4_search"
                 },
                 {
-                    "when": "explorerResourceIsFolder",
+                    "when": "explorerResourceIsFolder && config.svelte.ui.svelteKitFilesContextMenu.enable == always || explorerResourceIsFolder && svelte.uiContext.svelteKitFilesContextMenu.enable",
```
jasonlyu123, line 650: Parentheses don't seem to work here. So the `explorerResourceIsFolder` is repeated twice.

dummdidumm (approved):
Woah, you keep finding these APIs that make these things possible - thank you!

PR #1796, opened 2022-12-21
(breaking) svelte-check v3

dummdidumm:
- minimum required Node version is now 16
- remove hint diagnostics to align with tsc
- search for nearest tsconfig/jsconfig by default


avarun42:
Was the `--fail-on-hints` arg removed on purpose? It doesn't seem to be mention in any release notes.

dummdidumm:
It's mentioned in the release notes here: https://github.com/sveltejs/language-tools/releases/tag/svelte-check-3.0.1

Issue #1725, opened 2022-11-21
vscode does not recognize `.svelte` files imported in `.ts` files


DetachHead:
### Describe the bug

i'm not quite sure it's even supposed to be allowed, but i stumbled across a project that has lots of `.ts` files importing `.svelte` files that seem to work at runtime, but not in vscode.

i suspect it's because it's using the typescript language server instead of the svelte one?

### Reproduction

```svelte
<!-- +page.svelte -->
<script lang="ts">
    export const a = 1
</script>
```
```ts
import { a } from './+page.svelte'
```

### Expected behaviour

no error

### System Info

- OS: [e.g. Windows]
- IDE: [e.g. VSCode, Atom]

### Which package is the issue about?

Svelte for VS Code extension

### Additional Information, eg. Screenshots

![image](https://user-images.githubusercontent.com/57028336/203005312-ccdf46f0-1242-4e53-bb54-aef0adfed238.png)

dummdidumm:
Did you enable the Svelte TypeScript plugin? The vs code setting is called `svelte.enable-ts-plugin`

DetachHead:
yeah that's set to `true`

ngmunyong:
I had the same issue until I realised I didn't add that vs code setting json in the project.

1) create .vscode folder at project root
2) add the line mentioned above

![Screen Shot 2022-11-22 at 10 29 54 pm](https://user-images.githubusercontent.com/61854006/203303634-c9f80ac2-0841-4034-892d-fdf9fe40cebf.png)

Hope this works for you.

hangilc:
I have encountered the same problem, and changing the TypeScript version through Cmd-Shift-P "TypeScript: Select TypeScript version..." to version 4.9.3 ('Use Workspace version' in my case) solved the problem. Before changing it has been set to 5.0.0-dev20221217 ('Use VS Code's Version').

DetachHead:
@hangilc doesn't fix it for me

jasonlyu123:
I can reproduce what @hangilc is seeing. One of the TypeScript APIs we used in the TypeScript plugin no longer works in ts 5.0. @DetachHead if you follow what @hangilc said, what's the version you're seeing in the "Use VS Code's Version" option? If it's not TypeScript 5.x. We'll need a reproduction repository. I can't reproduce it with your steps.

DetachHead:
[...]

PR #1799, opened 2022-12-22
(fix) patch resolveModuleNameLiterals for ts 5.0 nightly

jasonlyu123:
https://github.com/sveltejs/language-tools/issues/1725#issuecomment-1356781781

`resolveModuleNames` is deprecated. TypeScript would use `resolveModuleNameLiterals` when it is available. This certainly isn't related to #1725 because the change only landed 2 weeks ago.

### DIFF
--- a/packages/typescript-plugin/src/module-loader.ts
+++ b/packages/typescript-plugin/src/module-loader.ts
@@ -5,25 +5,52 @@ import { SvelteSnapshotManager } from './svelte-snapshots';
 import { createSvelteSys } from './svelte-sys';
 import { ensureRealSvelteFilePath, isVirtualSvelteFilePath } from './utils';
 
+// TODO remove when we update to typescript 5.0
+declare module 'typescript/lib/tsserverlibrary' {
+    interface LanguageServiceHost {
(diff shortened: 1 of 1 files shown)


packages/typescript-plugin/src/module-loader.ts
```diff
@@ -119,32 +151,22 @@ export function patchModuleLoader(
             return resolved;
         }
 
-        return resolved.map((moduleName, idx) => {
-            const fileName = moduleNames[idx];
-            if (moduleName || !ensureRealSvelteFilePath(fileName).endsWith('.svelte')) {
-                return moduleName;
+        return resolved.map((tsResolvedModule, idx) => {
```
dummdidumm, line 154: Is it correct to adjust the code here? I assume that you updated the code to match the new 5.0 API, but when we are on 4.9 it's the old one, so we shouldn't update this.

jasonlyu123: Just trying to reuse some parts of it but not much is reused so I can revert it if you want.

dummdidumm: Oops I'm sorry, I didn't realize you refactored something out, all good then

dummdidumm (approved):

PR #1813, opened 2022-12-27
Remove JS and TS from activationEvents

DestroyerXyz:
I noticed the Svelte extension was loaded even when using some other framework (React for example).
This is likely because of the JS/TS language trigger, as it is not loaded on other languages.
Are those triggers really required?

### DIFF
--- a/packages/svelte-vscode/package.json
+++ b/packages/svelte-vscode/package.json
@@ -41,9 +41,7 @@
     },
     "activationEvents": [
         "onLanguage:svelte",
-        "onCommand:svelte.restartLanguageServer",
-        "onLanguage:javascript",
-        "onLanguage:typescript"
+        "onCommand:svelte.restartLanguageServer"
     ],
     "capabilities": {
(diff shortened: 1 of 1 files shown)


jasonlyu123:
Yes, This is intentional. It's because we have some logic for the typescript plugin needed to run on ts/js activation. But the svelte language server only started after at least one svelte file is opened.

DestroyerXyz:
Thanks for the quick reply!

How much of a performance impact does the extension have before the language server is started?
And can the ts plugin logic also be delayed until a svelte file is opened?

jasonlyu123:
The ts plugin logic is used to enable the plugin so it can't be delayed because someone might open a ts file before they opened svelte files. The "SvelteKit files" context menu option also makes sense to trigger on ts/js files creation. 

Thanks for the contribution but I am going to close this.  However, If any other extension features can be delayed. contribution is welcomed.

DestroyerXyz:
Yeah makes sense, thanks for the info!

Issue #1792, opened 2022-12-20
Autocompletion for stores that need to be imported


dummdidumm:
### Description

Typing `$foo` in a Svelte file where `foo` is defined in some other file and not imported yet results in no auto import completion currently. It would be great if it would.

### Proposed solution

The idea would be to do a hidden global autocompletion at the start of the file and filter for the given name just without the `$` prefix. To not tank performance, those autocompletions should be cached. The idea could be to trigger them on `$` and save them for a minimum of 10 seconds or as long as no modification to another Svelte file's module context or another TS/JS file was made.

(Completions are already triggered currently when typing `$foo` but they don't contain `foo` as a suggestion due to the leading dollar sign)

### Additional Information, eg. Screenshots

Continuation of #1583

PR #1816, opened 2022-12-29
(feat) completion label detail support

jasonlyu123:
圖片

I also changed the `Auto import from ...` message to the message provided by typescript. Making it consistent with VSCode's built-in typescript feature. 

圖片

One thing that differs from VSCode+tsserver is that tsserver has some internal APIs for caching auto-import-related info. So the source of the auto-import entry would be calculated at the completion list stage. Without the caching, it would only resolve some of it and the rest would need to be resolved at the resolve completion detail stage. Causing the label detail to be pop-in afterwards.

ours:

圖片

ts
圖片

Volar has a [package](https://github.com/johnsoncodehk/volar/tree/master/packages/typescript-faster) for patching the internal API. We can consider adding that. It would also help #1792. I tested it with cache, the global completion can improve from 800ms to 100-200ms in one of my projects. But even if we use their package we'll still have to access the internal API for cache invalidation. Left the internal API testing branch here https://github.com/jasonlyu123/language-tools/tree/label-detail-internal for comparison.

### DIFF
--- a/packages/language-server/src/ls-config.ts
+++ b/packages/language-server/src/ls-config.ts
@@ -400,7 +400,8 @@ export class LSConfigManager {
                 config.suggest?.includeCompletionsForImportStatements ?? true,
             includeAutomaticOptionalChainCompletions:
                 config.suggest?.includeAutomaticOptionalChainCompletions ?? true,
-            includeCompletionsWithInsertText: true
+            includeCompletionsWithInsertText: true,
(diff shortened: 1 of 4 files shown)


dummdidumm (approved):
Thank you!

That performance enhancement using the volar library sounds promising. Just to be clear - would this improve the auto completion performance in all/most cases, or just specific ones? Definitely interested in adding this if it improves performance across the board - could you do that in a different PR?

PR #1817, opened 2022-12-30
(feat) object/class member snippet

jasonlyu123:
class member in 4.5
https://devblogs.microsoft.com/typescript/announcing-typescript-4-5/#subclass-method-snippets

object literal in 4.7
https://devblogs.microsoft.com/typescript/announcing-typescript-4-7/#object-method-snippet-completions

this should be merged after #1816 because the test will only pass with label detail.


dummdidumm (approved):
Awesome!

PR #1818, opened 2022-12-30
(feat) autoImportFileExcludePatterns config support

jasonlyu123:
added in typescript 4.8 https://devblogs.microsoft.com/typescript/announcing-typescript-4-8/#exclude-specific-files-from-auto-imports

The pattern normalization part is adopted from VSCode. But I found leading wildcard patterns don't work in Windows. It does work in Ubuntu WSL. So I changed it also to append workspace root. (Noted to self, should open an issue on either VSCode or TypeScript to ask about it)

Also, should we default some SvelteKit files with this config? Like `+error.{svelte,ts,js}`. Seen someone in the discord a few times accidentally trigger auto import when trying to throw an error. 

https://discord.com/channels/457912077277855764/1053932682955997234/1053932682955997234

### DIFF
--- a/packages/language-server/src/ls-config.ts
+++ b/packages/language-server/src/ls-config.ts
@@ -4,6 +4,8 @@ import { VSCodeEmmetConfig } from '@vscode/emmet-helper';
 import { importPrettier } from './importPackage';
 import { Document } from './lib/documents';
 import { returnObjectIfHasKeys } from './utils';
+import path from 'path';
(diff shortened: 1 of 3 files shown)


dummdidumm (approved):
Nice! I'd wait a little longer before adding any defaults. Did you see other instances of this discord question (meaning that a lot of people run into this issue)?

jasonlyu123:
At least one more. Another solution is to deprioritize route files in completion since they're rarely auto-imported.

PR #1823, opened 2023-01-09
(feat) Autocompletion for stores that need to be imported

dummdidumm:
#1792

I didn't implement the mentioned performance thing yet, I don't think it's that bad and invocations of this are probably so rare while also making use of the cache that I thought it's not worth it


janosh:
This a great to have! Definitely something I've been missing. 👍

PR #1842, opened 2023-01-23
Resolve Prettier plugins relative to document path

### LINKED ISSUE #1766: Extension is not compatible with local prettier-plugin-svelte plugin
### Describe the bug

When the project `.prittierrc` contains

```json
{
   plugins: [prettier-plugin-svelte],
   (...)
}
```

the extension cant format the `.svelte` files
I know that the extension has "prettier-plugin-svelte" itself but I need the plugin to make pretty-quick work

### Reproduction

Add `prettier-plugin-svelte` to `.prittierrc` plugins

Try to format .svelte file

### Expected behaviour

extension should work with local  prettier-plugin-svelte

### System Info

- OS: Windows
- IDE:  VSCode

Svelte for VS Code extension

_No response_

joas8211:
Fixes #1766

Prettier resolves plugins relative to process's current working directory. For the language server it is VSCode's open workspace directory. The CWD probably would be in the node_modules folder if Prettier were to be ran in a separate process (eg. in command line). So it doesn't find node_modules from a sub-directory. That's why it should be resolved relative to formatted file.

### DIFF
--- a/packages/language-server/src/plugins/svelte/SveltePlugin.ts
+++ b/packages/language-server/src/plugins/svelte/SveltePlugin.ts
@@ -106,7 +106,12 @@ export class SveltePlugin
 
         const formattedCode = prettier.format(document.getText(), {
             ...config,
-            plugins: [...(config.plugins ?? []), ...getSveltePlugin()],
+            plugins: [
+                ...(config.plugins ?? []).map((id: string) =>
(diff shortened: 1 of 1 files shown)


dummdidumm:
Thank you! #1829 contains a more robust fix so I'll close in favor of that.

PR #1853, opened 2023-02-02
Use satisfies in typescript snippets

david-plugge:


dummdidumm:
I like this change, but I'll wait a few more weeks so chance is higher that people had time to upgrade TypeScript and VSCode to a version where `satisfies` exists.

benmccann:
@dummdidumm perhaps this one is worth another look now?

Issue #1336, opened 2022-01-14
TypeScript inlay hints


blaumeise20:
**Is your feature request related to a problem? Please describe.**
VSCode added inlay hints a while ago. The problem is that svelte doesn't show them in script tags.

**Describe the solution you'd like**
Show inlay hints like in normal JS/TS files.

jasonlyu123:
I think we'll wait until the API is finalized and ships in the Language Server Protocol. Currently, it seems like the API is still in the proposed stage in vscode.

dummdidumm:
LSP 3.17 has support for inlay hints https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_inlayHint

LeoDog896:
TypeScript now has [inlay hints](https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/)

insilications:
Has the API been finalized in LSP?

jasonlyu123:
Yes. And the feature we want to prioritize is also done. We'll do it soon.

PR #1855, opened 2023-02-04
(feat) TypeScript Inlay Hint

jasonlyu123:
#1336

This is not enabled by default. You'll have to enable it with the TypeScript inlay hints configs of VSCode. There are multiple configs under `javascript.inlayHints.*` and `typescript.inlayHints.*`. Enable them based on your preference. 

## implementation detail
The test is set up similarly to the diagnostics snapshot test. You can also run `yarn test --debug` to output `debug.svelte` with inlay hints added to the original file as comments. for example:

```svelte
<script lang="ts">
    import { SvelteComponentTyped } from 'svelte';

    let Component: typeof SvelteComponentTyped;
    function hi(name: string)/* : void*/ {}
</script>

<Component text={hi(/*name: */'')}></Component>
``` 

## Screenshots

圖片
圖片

(enable all inlay hints options is really just too much for my taste lol)


dummdidumm (approved):
Damn, this is impressive work - thank you so much!

polyzen:
Is this also available via `svelte.plugin.typescript.inlayHints.enable`? Not listed in the readme. Would like to use this in Neovim.

dummdidumm:
This uses the VSCode TypeScript options to enable it. How does Neovim enable it for TS (through which option(s))?

polyzen:
> This uses the VSCode TypeScript options to enable it. How does Neovim enable it for TS (through which option(s))?

https://github.com/lvimuser/lsp-inlayhints.nvim#typescript

aidenlangley:
https://github.com/sveltejs/language-tools/issues/1895

Neovim doesn't handle any of the workspace/something/refresh requests right now. Need this to be configurable please.

dummdidumm:
@polyzen you can read about the settings here: https://github.com/sveltejs/language-tools/tree/master/packages/language-server#settings , settings references are linked

GustavoJCL:
How can I enable it on neovim? Neovim 0.10 now have built-in inlay hints

PR #1860, opened 2023-02-09
chore: move `typescript` to `svelte-check`'s peer dependency

Andarist:
It is kinda debatable but I feel that it's better to use the installed version of `typescript`. I just had to go through a surprising debugging session since my IDE was using TypeScript 4.8 but `svelte-check` was using TypeScript 4.5 (I had an old version that had the `*` range declared on typescript and TS 4.5 was left in the lockfile from before the rest of the monorepo bumped TS to TS 4.8.

I understand that this might create a different problem though - now the `svelte-check` might run with a different version of TS than the VS Code extension. However, this isn't a new problem - it already exists.

### DIFF
--- a/packages/svelte-check/package.json
+++ b/packages/svelte-check/package.json
@@ -29,11 +29,11 @@
         "fast-glob": "^3.2.7",
         "import-fresh": "^3.2.1",
         "sade": "^1.7.4",
-        "svelte-preprocess": "^5.0.0",
-        "typescript": "^4.9.4"
+        "svelte-preprocess": "^5.0.0"
     },
     "peerDependencies": {
-        "svelte": "^3.55.0"
+        "svelte": "^3.55.0",
+        "typescript": "^4.9.0"
     },
     "scripts": {
         "build": "rollup -c && node ./dist/src/index.js --workspace ./test --tsconfig ./tsconfig.json",
@@ -52,6 +52,7 @@
         "rollup-plugin-cleanup": "^3.2.0",
         "rollup-plugin-copy": "^3.4.0",
         "svelte-language-server": "*",
+        "typescript": "^4.9.4",
         "vscode-languageserver": "8.0.2",
         "vscode-languageserver-protocol": "3.17.2",
         "vscode-languageserver-types": "3.17.2",


dummdidumm:
I'm very hesitant to merge this. Having it as a peer dependency allows us to make use of new TypeScript constructs without having to release a major version `svelte-check`. It doesn't happen often but it did before.

Andarist:
From my PoV it's important to consider user expectations here.

Scenario 1:
1. I upgrade my project to TS 5.0 (the version itself doesn't really matter, it's just an example) and I start using `const` type parameters
2. Can I expect `svelte-check` to handle this new syntax? And this isn't even about new syntax - any TS version can change the type checking behavior. Is it expected that I don't get any errors when running `tsc` or in the IDE (in my regular TS files) or is it expected that I can run into problems here? Is this caveat communicated in the docs anyhow?

Scenario 2:
1. I upgrade `svelte-check` and it starts using TS 5.0 with its [more strict relational operators](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0-beta/#forbidden-implicit-coercions-in-relational-operators)
2. Is it expected that I start getting new errors for the existing code only when running through `svelte-check`?

> without having to release a major version svelte-check

What's the problem with new major versions though? They are honest and convey the intent nicely. It's just like age - it's just a number 😉

DavidArchibald:
I can second that I ran into this problem and it was an annoying surprise. I was going to file this as an issue myself!

packages/svelte-check/package.json
```diff
@@ -29,11 +29,11 @@
...
         "import-fresh": "^3.2.1",
         "sade": "^1.7.4",
-        "svelte-preprocess": "^5.0.0",
-        "typescript": "^4.9.4"
+        "svelte-preprocess": "^5.0.0"
     },
     "peerDependencies": {
-        "svelte": "^3.55.0"
+        "svelte": "^3.55.0",
+        "typescript": "^4.9.0"
```
benmccann, line 36: This should be updated to allow Typescript 5 to be used as well
```suggestion
        "typescript": ">=4.9.0"
```

dummdidumm:
Closing in favor of #2453

PR #1865, opened 2023-02-13
Add -replacestate/-keepfocus to SvelteKit anchor tag props

axwalker:
Corresponding changes for `data-sveltekit-replacestate` addition within SvelteKit: https://github.com/sveltejs/kit/pull/9019


dummdidumm (approved):
Thank you! There's one more place where this needs to be added, and that's in https://github.com/sveltejs/svelte/blob/master/elements/index.d.ts

axwalker:
Ah, that's why this one is marked as deprecated. Thanks for the direction! Here's the PR: https://github.com/sveltejs/svelte/pull/8281

PR #1872, opened 2023-02-15
(fix) use closest config file

dummdidumm:
Not any tsconfig over a closer jsconfig. This aligns with TS VS Code behavior.

@jasonlyu123 could you double-check if that's the correct behavior?

### DIFF
--- a/packages/language-server/src/plugins/typescript/service.ts
+++ b/packages/language-server/src/plugins/typescript/service.ts
@@ -91,13 +91,23 @@ export async function getService(
         return getServiceForTsconfig(tsconfigPath, dirname(tsconfigPath), docContext);
     }
 
+    // Find closer boundary: workspace uri or node_modules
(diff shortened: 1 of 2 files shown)


jasonlyu123:
Yeah, it should be the same behaviour as the VSCode/tsserver. 

They didn't use `findConfigFile`, though. But a specialized method for the tsserver with some optimization. The loading logic is here. You can check if there are any differences. 
https://github.com/microsoft/TypeScript/blob/2a8436c529640327a3460463fc945694584ff1c3/src/server/editorServices.ts#L2090

It starts from the directory where the file is and walks up the directory tree. If there is any tsconfig or jsconfig, return it. 

On a side note, We could also consider caching the result like tsserver in the future. But since the file-exists check is usually fast, it probably won't make a huge difference.

dummdidumm:
Thanks for the check! I kept the more simple version we have now but added an additional `node_modules` check (like they do) and added the caching - I think it's a good thing to make, much less IO operations then.

Issue #1863, opened 2023-02-10
Rename of prop should automatically expand abbreviated assignments


wiwa:
### Description

(New to Svelte, sorry if already answered)

In VSCode (v 1.75.0), using "Rename Symbol":
When a parent passes a prop, say `{sameNameProp}` to a child component (e.g. `Child`) with `export let sameNameProp`, and that child renames their symbol, to say `otherNameProp`, then VSCode will rename the prop reference in the parent as well.

Unfortunately, this currently results in:
`<Child {otherNameProp}>`, but the parent still only defines `sameNameProp`, causing a breakage.

### Proposed solution

When renaming a prop whose references include abbreviations, automatically expand the abbreviations. In the example case, the parent's code should have become: 

`<Child otherNameProp={sameNameProp} />`

### Alternatives

On the other hand, this could be intended behavior which forces the developer to think about renaming props. Possible alternatives:

1) Instead of changing "Rename Symbol" behavior, add an action to the Svelte extension so developers can choose to compatibly rename props or not.  
2) Add the action, but allow an option somewhere to set the default behavior of "Rename Symbol".
3) (Assuming the current behavior is intended/desirable): Add documentation about it.

### Additional Information, eg. Screenshots

![svelte-rename-prop-vscode](https://user-images.githubusercontent.com/1862202/218199970-8e2068e1-ef5c-4453-b502-03334f4633db.png)

jasonlyu123:
We already have this mechanism. This might be a bug in some edge cases. What is the version of the Svelte for VSCode you're using? Can you provide a reproduction and specify where did you trigger rename?

jasonlyu123:
Oh. Never mind. I misunderstood.

PR #1877, opened 2023-02-17
(fix) don't rename props name when triggered in shorthand

jasonlyu123:
Found this while figuring out #1863 

If you trigger the rename in the shorthand of ComponentB in ComponentA, we'll check the props rename for ComponentA. And if they have the same name, for example

```svelte
<script>
import ComponentB from './ComponentB.svelte'
export let name;
</script>

<ComponentB {name} /> <!--trigger here in {name} -->
```

Because there might be two edits to the exact shorthand location. It sometimes became `name={name}` and sometimes became `na={nameme}`.

I also fixed a problem when the rename is triggered in `export let` without type annotation or default. Rename won't update props usage. This is done in the same PR because this problem prevents the main issue from happening. Thus the added test won't fail even without the fix. 

```svelte
<script>
import ComponentB from './ComponentB.svelte'
export let name; // trigger here
</script>
```

### DIFF
--- a/packages/language-server/src/plugins/typescript/features/RenameProvider.ts
+++ b/packages/language-server/src/plugins/typescript/features/RenameProvider.ts
@@ -7,7 +7,13 @@ import {
     isInHTMLTagRange,
     getNodeIfIsInHTMLStartTag
 } from '../../../lib/documents';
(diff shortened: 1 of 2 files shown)


dummdidumm (approved):
nice!

Issue #1835, opened 2023-01-18
emitDts should print diagnostic messages


dangreen:
### Describe the bug

When there is a type error in a svelte component, `emitDts` just silently doesn't emit `.d.ts` file

### Reproduction

Any svelte component with type error

### Expected behaviour

https://github.com/sveltejs/language-tools/blob/6ef2442d48b1e90a38e4e8c4e518b336aeaa191d/packages/svelte2tsx/src/emitDts.ts#L16

here need to handle the result of emit, and print diagnostic messages if any exists

### System Info

- OS: any
- IDE: any

### Which package is the issue about?

svelte2tsx

### Additional Information, eg. Screenshots

Related issue: https://github.com/sveltejs/kit/issues/3783

dummdidumm:
`emitDts` is a primitive who's job is not type checking. The error messages that come from it would probably contain a lot of weird things and false positives; also the line/columns of the errors would be all wrong because they use the generated code, which is why we wrap its usage inside the `svelte-language-server` to augment/adjust the output.

I think the better way would be to run `svelte-check` before running `svelte-package`.

PR #1886, opened 2023-02-21
fix(emitDts): set rootDir for consistent output structure

gtm-nayan:
Ref https://github.com/sveltejs/kit/issues/9118#issuecomment-1437141020 and https://github.com/microsoft/TypeScript/issues/52889

Marking as draft cause, should it be a hard error or a warning and who is responsible for logging it: emitDts or the caller?

### DIFF
--- a/packages/svelte2tsx/src/emitDts.ts
+++ b/packages/svelte2tsx/src/emitDts.ts
@@ -52,7 +52,7 @@ function loadTsconfig(config: EmitDtsConfig, svelteMap: SvelteMap) {
         tsConfig,
         ts.sys,
         basepath,
-        { sourceMap: false },
+        { sourceMap: false, rootDir: config.libRoot },
         tsconfigFile,
         undefined,
         [{ extension: 'svelte', isMixedContent: true, scriptKind: ts.ScriptKind.Deferred }]
@@ -81,7 +81,7 @@ function loadTsconfig(config: EmitDtsConfig, svelteMap: SvelteMap) {
             declaration: true, // Needed for d.ts file generation
             emitDeclarationOnly: true, // We only want d.ts file generation
             declarationDir: config.declarationDir, // Where to put the declarations
-            allowNonTsExtensions: true
+            allowNonTsExtensions: true,
         },
         filenames
     };
--- a/packages/svelte2tsx/src/emitDts.ts
+++ b/packages/svelte2tsx/src/emitDts.ts
@@ -81,7 +81,7 @@ function loadTsconfig(config: EmitDtsConfig, svelteMap: SvelteMap) {
             declaration: true, // Needed for d.ts file generation
             emitDeclarationOnly: true, // We only want d.ts file generation
             declarationDir: config.declarationDir, // Where to put the declarations
-            allowNonTsExtensions: true,
+            allowNonTsExtensions: true
         },
         filenames
     };


packages/svelte2tsx/src/emitDts.ts
```diff
@@ -13,7 +13,13 @@ export async function emitDts(config: EmitDtsConfig) {
     const { options, filenames } = loadTsconfig(config, svelteMap);
     const host = await createTsCompilerHost(options, svelteMap);
     const program = ts.createProgram(filenames, options, host);
-    program.emit();
+
+    return program.emit(undefined, (...args) => {
+        const [filename] = args;
+        if (isSubpath(config.declarationDir, path.resolve(filename))) {
+            host.writeFile(...args);
```
dummdidumm, line 20: Can you add a little comment explaining why we're doing this?

dummdidumm (commented):
Thanks for figuring that out! This makes sense to me

packages/svelte2tsx/src/emitDts.ts
gtm-nayan: Done

gtm-nayan:
I made it return an object containing the diagnostics from ts itself, `@sveltejs/package` should probably filter based on the specific "files outside root" error's code cause otherwise it also contains errors on imports like `$app/..` because include is different from `.svelte-kit/tsconfig.json`.

The test timeout for the plugin seems unrelated (hopefully 😬).

packages/svelte2tsx/src/emitDts.ts
```diff
@@ -13,7 +13,20 @@ export async function emitDts(config: EmitDtsConfig) {
     const { options, filenames } = loadTsconfig(config, svelteMap);
     const host = await createTsCompilerHost(options, svelteMap);
     const program = ts.createProgram(filenames, options, host);
-    program.emit();
+
+    // EmitResult below doesn't contain them for some reason, so get them separately
+    const diagnostics = ts.getPreEmitDiagnostics(program);
```
dummdidumm, line 18: I'm wondering how much slower the emit is becoming through this, AFAICT TS will do some thorough code analyses there (which might also close #1835). Should we have this behind an option?

gtm-nayan: The problem with rolling our custom warnings is that it doesn't contain information on where the bad import happened. I suppose we could set a flag if such files are found and only run the diagnostics if it's true?

gtm-nayan: As for #1835, I don't think emitDts's diagnostics can be fully relied upon. It already gets a few false positives because the tsconfig passed to it doesn't line up with the generated by SvelteKit.

dummdidumm:
My idea for generating d.ts files is now that we don't pass the `libRoot` option anymore, instead only keeping the the generated `d.ts` files inside of `src/lib` (or whatever is configured) later on. This would make this PR obsolete I think?
If people want to have a check that their generated output is correct they could run `tsc` on the `dist` folder's d.ts files afterwards.

gtm-nayan:
> My idea for generating d.ts files is now that we don't pass the libRoot option anymore

Do you mean you want to change it to be that way at some point? Right now it's still there. I wrote my thoughts on not wanting to do that in that [other thread](https://github.com/sveltejs/kit/issues/9118#issuecomment-1437293913), would there be any benefit to it?

> If people want to have a check that their generated output is correct they could run tsc on the dist folder's d.ts files afterwards.

Do we trust people to do that or should svelte-package include the check in its scripts?

dummdidumm:
> > My idea for generating d.ts files is now that we don't pass the libRoot option anymore
> 
> Do you mean you want to change it to be that way at some point? Right now it's still there. I wrote my thoughts on not wanting to do that in that [other thread](https://github.com/sveltejs/kit/issues/9118#issuecomment-1437293913), would there be any benefit to it?

The benefit is that we don't mess around with the tsconfig include/files settings. Input/output settings are no problem because `<output>/__package_types_tmp__` is the declaration dir and we flatten the directory structure afterwards (`relative(tsconfigdir, input)`). It's possible slower though, that is right. Maybe we can do this later if more problems arise.

New proposal: Don't return any diagnostics. Just emit a one-time warning when we detect that some file would be written out that is outside the root dir.

> > If people want to have a check that their generated output is correct they could run tsc on the dist folder's d.ts files afterwards.
> 
> Do we trust people to do that or should svelte-package include the check in its scripts?

We trust people to do that if they want to. It's finicky to build it into `svelte-package`, I wouldn't want to do that right now - the related issue is somewhat of an edge case to me.

dummdidumm:
I'm switching my mind constantly 😅 I now think we should just set `rootDir` to `config.libRoot` (if it's undefined then `rootDir` is, too) and on the SvelteKit side we can decide to set it to `"."` to say "relative to the tsconfig"

gtm-nayan:
Yeah that's what I meant, should this stay in place for now? https://github.com/sveltejs/language-tools/pull/1886/files#diff-783d276c3f0448d53760c169c6843c026704cd3329bcabbbd46c6234cdcef69bR47

dummdidumm (approved):
Yeah that makes sense!

PR #1889, opened 2023-02-22
(feat) call hierarchy support

jasonlyu123:
Call hierarchy lets you see where a function or a class is called and follow the call stack up. Or see what other functions a process called and follow the call stack down. It's powerful for a large project. This PR added the support and some custom logic to make it work with svelte components. 

## How to
Right-click on a function or class name and choose `Show Call Hierarchy`. 
圖片

It also works with a component tag and the `script` tag.

## Screenshot
圖片

## implementation

The `IncomingCalls` is triggered in the callee to find the caller, and the `OutgoingCalls` is started in the caller to see what function it calls. 

Besides what TypeScript finds, In `IncomingCalls`, we use the component's references to find those passed as an argument to `__sveltets_2_ensureComponent`. And redirect the position to the generated class. In `OutgoingCalls`, we also check all the start tags and redirect the search position to the generated render function.


dummdidumm (approved):
Spectacular!

Issue #1895, opened 2023-02-24
SemanticTokensRefresh & InlayHintRefresh are unsupported by Neovim


aidenlangley:
### Describe the bug

I believe `workspace/SemanticTokens/refresh` & `workspace/inlayHints/refresh` requests made to Neovim are causing the server to crash, since the client doesn't support these operations.

My setup is to have `svelte-language-server` running in `.svelte` buffers, but `tsserver` in `.ts` buffers. Saving, or doing just about anything, in a `typescript` file is causing the `svelte-language-server` to send the requests I mentioned, so it can provide an update on the diagnostics state to the client.

It should perhaps be configurable - none of the options provided disable this functionality.

### Reproduction

A Neovim setup using with the default svelte and typescript config will suffice.

1. Edit a `svelte` file to start up the `svelte-language-server`.
2. Edit a `typescript` file to start up `tsserver`.
3. `svelte-language-server` will crash with the following output to `.local/state/nvim/lsp.log` or equivalent on another OS:

```config
[WARN][2023-02-24 01:06:28] .../lua/vim/lsp.lua:1065	"server_request: no handler found for"	"workspace/inlayHint/refresh"
[WARN][2023-02-24 01:06:28] .../lua/vim/lsp.lua:1065	"server_request: no handler found for"	"workspace/semanticTokens/refresh"
[...]

PR #1896, opened 2023-02-24
(fix) don't refresh for unsupported client

jasonlyu123:
#1895

### DIFF
--- a/packages/language-server/src/ls-config.ts
+++ b/packages/language-server/src/ls-config.ts
@@ -305,14 +305,14 @@ export class LSConfigManager {
     /**
      * Updates config.
      */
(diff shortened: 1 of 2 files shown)


packages/language-server/src/ls-config.ts
```diff
@@ -615,4 +617,13 @@ export class LSConfigManager {
             this.listeners.forEach((listener) => listener(this));
         });
     }
+
+    updateClientCapabilities(clientCapabilities: ClientCapabilities) {
```
jasonlyu123, line 621: Put it here so we can access it in the plugins later if needed.

dummdidumm (approved):

PR #1914, opened 2023-03-03
Make bug template package selection `multiple`

brunnerh:
Type issues commonly affect the editor experience via the extension and `svelte-check`, so it is helpful if both can be selected directly.


dummdidumm (approved):
Makes sense, thanks!

PR #1919, opened 2023-03-07
(chore) TypeScript 5.0

dummdidumm:
TODOs:

- [x] `ModuleResolutionKind` enum has changed (also see https://github.com/microsoft/TypeScript/issues/53131)
- [x] sinon.replace doesnt work anymore because ts is a module now, we need to find a different way for some of our tests
- [x] bunch of tests need updates

Note to self: We don't need to handle the `ignoreDeprecations` situation because we parse the tsconfig but don't use its diagnostics, which is the only time where errors could be thrown because of an invalid tsconfig. It doesn't matter down the line for the other parts of language tools.


dummdidumm:
Let's wait for 5.0.3 before merging/releasing this so the TS team had the opportunity to iron out the worst bugs - this being such a big release there are likely more of those than usual.

janosh:
Just tried it out. Resolution works great on packages using `"moduleResolution": "bundler"`. Thanks!

Does this extension have control over whether command-clicking an import opens the `.d.ts` file or the actual component? Currently it's the former but would prefer latter.

PR #1931, opened 2023-03-10
(fix) Kit fixes

dummdidumm:
- hopefully fix issue with linux
- check return type of api methods


dummdidumm:
@jasonlyu123 FYI I had to revert the document registry improvement suggestion of yours, seems it fails on Linux for some reason. When I have WSL set up etc I may look into why exactly this is the case.

PR #1939, opened 2023-03-15
(feat) fix-all quick fix

### LINKED ISSUE #1885: Quick fix "Add all missing imports" for components
### Description

When copying a markup snippet from another Svelte file, multiple component imports might be missing.
It would be helpful to automatically add all such imports at the same time.

### Proposed solution

Add a quick fix action that adds the best matching import suggestion for all currently unknown components in the file.

### Alternatives

Import each individually.

This would be analogous to this TS/JS quick fix:
image

jasonlyu123:
#1885. I might still do some refactoring later. But I made many changes, so opening it as a draft for now as a heads-up for potential conflict. 

There are two parts to this PR. First is supporting the combined quick for TypeScript. And Second part is to add our "import component" and "import store" fix on top of it. Handler is skipped since TypeScript 5.0 should have full support, and our custom code would only be backup for non-VSCode users still using < 5.0.

~~About the custom quick fix. I reuse the logic in the usual quick fix. Except we ask TypeScript for the full diagnostics first as the source. We take the first one from every name. Then remove the empty line in between, and wrap the script tag if needed. The caveat is that the "which is first" logic Typescript used is way more complex, our custom imports need to be sorted, and those from the same module need to be combined into one import. This logic is very complex, so I am leaning toward not doing anything if we can't find a workaround to let TypeScript do its magic.~~

Edit:
The implementation of the customs import fix has changed. Now I create a copy of the file with the declared name of the store and component variables at the end of the script tags. This way, we don't have to deal with the "which is first" and import organization problems. Performance-wise, It's sometimes slower than the custom solution because TypeScript will rebuild and invalidate the cache. But it's more consistent. The original method might be very slow if there are a lot of missing imports.


packages/language-server/src/plugins/typescript/features/CodeActionsProvider.ts
```diff
@@ -264,324 +542,504 @@ export class CodeActionsProviderImpl implements CodeActionsProvider {
...
+                        if (fix.fixName === 'unusedIdentifier') {
+                            originalRange = this.checkRemoveImportCodeActionRange(
+                                edit,
+                                snapshot,
+                                originalRange
+                            );
+                        }
+
+                        if (fix.fixName === 'fixMissingFunctionDeclaration') {
+                            const position = 'position' in fix ? fix.position : undefined;
```
jasonlyu123, line 697: Another difference other than indent in `convertAndFixCodeFixAction`

packages/language-server/src/plugins/typescript/features/CodeActionsProvider.ts
```diff
@@ -264,324 +542,504 @@ export class CodeActionsProviderImpl implements CodeActionsProvider {
...
-                ts.isIdentifier(node.parent.expression) &&
-                node.parent.expression.text === '__sveltets_2_ensureComponent' &&
-                ts.isIdentifier(node)
-        );
+        const nameToPosition = new Map<string, number>();
 
-        if (!node) {
-            return;
-        }
+        for (const diagnostic of diagnostics) {
```
jasonlyu123, line 882: I didn't revert this part when switching to the virtual file method because you can select a range to start code action. So there are places where it will have multiple positions.

jasonlyu123:
Guess it might not be a huge difference. Because we filtered out unmappable code actions. And the completion-powered one all need the code action returned by completion resolve.

PR #1940, opened 2023-03-15
(chore) mark svelte2tsx as compatible with TS 5

dummdidumm:


packages/svelte2tsx/src/emitDts.ts
```diff
@@ -176,6 +176,22 @@ async function createTsCompilerHost(options: any, svelteMap: SvelteMap) {
             return resolveModuleName(moduleName, containingFile, compilerOptions);
         });
     };
+    host.resolveModuleNameLiterals = (
```
dummdidumm, line 179:
```suggestion
    // @ts-expect-error remove once we bump dev dep to TS 5
    host.resolveModuleNameLiterals = (
```

Issue #1938, opened 2023-03-14
TS errors when destucturing in a reactive statement without declaring variables first


brandonmcconnell:
### Describe the bug

When using Svelte Language Tools, TypeScript errors are encountered when destructuring an object in a reactive statement without explicitly declaring the variables first. This occurs even when TypeScript should be able to infer the types and assignments correctly. The expected behavior is that the destructuring should work without any TypeScript errors if the types can be inferred.

** This is a NEW bug, and this was working correctly up until recently. I am not sure of when this bug first appeared.

### Reproduction

```svelte
<script lang="ts">
  const person: {
    name: string;
    age: number;
  } = {
    name: "John Doe",
    age: 30,
  };

  $: ({ name, age } = person);
</script>
```

In this example, I don't immediately see the same issue, but I do see that the type for `age` is inferred as an implicit `any`.

To contrast, note that declaring the variables before destructuring correctly types `age` as `number`:

```svelte
<script lang="ts">
  const person: {
    name: string;
    age: number;
  } = {
    name: "John Doe",
    age: 30,
  };

  let name: string;
  let age: number;

  $: ({ name, age } = person);
</script>
```

### Expected behaviour

The types should be properly inferred when creating variables via a reactive statement without needing to declare them beforehand.

### System Info

- OS: macOS Ventura v13.1 (22C65)
- IDE: VSCode v1.76.1 (Universal)

### Which package is the issue about?

Svelte for VS Code extension

### Additional Information, eg. Screenshots

Screen recording:

https://user-images.githubusercontent.com/5913254/225159116-8fa20a0c-7305-4a7a-bd40-40430d45ca71.mov

<br />

It's also noteworthy that if I first initialize the variables this way, I can then destructure them the same way I did before and not get the same errors:

<img width="603" alt="Screenshot 2023-03-14 at 3 44 39 PM" src="https://user-images.githubusercontent.com/5913254/225159139-4da4337c-04cc-409c-9c8f-9a5d77934b63.png">

jasonlyu123:
I cannot reproduce with your reproduction. Can you provide another one? This might be an edge case that we didn't consider.

brandonmcconnell:
@jasonlyu123 Is there any online Svelte REPL that supports TS and intellisense so I can reproduce it reliably online?

brandonmcconnell:
Also, @jasonlyu123 — are you running on the latest release of the Svelte VSCode extension that was released yesterday, and also the latest release of VSCode?

[...]

PR #1941, opened 2023-03-16
(fix) more function scope check

jasonlyu123:
#1938

Found there is an `isFunctionLike` function.


dummdidumm:
Nice find! Does this also fix https://github.com/sveltejs/language-tools/issues/1938#issuecomment-1471230569 ?

jasonlyu123:
Yes. the function type part has the same syntax kind as a normal type annotation

brandonmcconnell:
@jasonlyu123 Thanks for the quick res!

PR #1943, opened 2023-03-16
(feat) add svelte files to path completion in the ts plugin

jasonlyu123:
TypeScript 5.0 will have path completion for *.ext ambient modules. We'll lose these completions when the ts plugin is enabled. Adding the `.svelte` extension to readDirectory as we do in the language server in this PR. 

I also change the null check because if this parameter is falsy in the TypeScript source code, TypeScript won't filter files. So this null check is wrong, but I don't know if there are any places where it would be null. 

The last change is also patching `getDirectories` in sveltekit language service host proxy. This is used for the directories in the path completion.


dummdidumm (approved):
👍

Issue #1956, opened 2023-03-23
JS/TS language service crashes


Ballig:
### Describe the bug

When more than one Typescript file is opened then the JS/TS language service immediately crashes 5 times until VS Code disables it:

`The JS/TS language service immediately crashed 5 times. The service will not be restarted. This may be caused by a plugin contributed by one of these extensions: svelte.svelte-vscode. Please try disabling these extensions before filing an issue against VS Code.`

### Reproduction

Having more than one Typescript file open.

### Expected behaviour

The TS/JS language service to continue working.

### System Info

- OS: Windows_NT x64 10.0.19044
- IDE: VS Code version: Code 1.76.2 (ee2b180d582a7f601fa6ecfdad8d9fd269ab1884, 2023-03-14T17:55:54.936Z)
Extension version: 107.2.5

### Which package is the issue about?

Svelte for VS Code extension

### Additional Information, eg. Screenshots

![svelte-ss1](https://user-images.githubusercontent.com/7276642/227149307-8bddbecf-9fe3-4597-b910-b59edbae39c6.PNG)
![svelte-ss2](https://user-images.githubusercontent.com/7276642/227149339-c0c6b255-1069-4e66-8eaa-a6bcbc1e6308.PNG)

jasonlyu123:
Can you provide the log file listed in the TypeScript output?

dummdidumm:
Strange, that log stop right after it says it's about to load the TypeScript plugin. What happens if you set `"svelte.enable-ts-plugin": false` in your VSCode settings? This should turn off the TypeScript plugin which should help in diagnosing whether or not it's related to something during its setup, or if the whole plugin is somehow corrupted on disk. If it keeps happening with the TS plugin turned off, does this error go away when you disable the Svelte extension completely?

Ballig:
I still get the JS/TS language service immediately crashing 5 times (when opening a second TS file) after turning the Enable TS Plugin option off (and restarting VS Code). The error does go away when disabling the Svelte extension completely.

dummdidumm:
Interesting - this sounds to me like the installation is corrupted in some way. Does manually deleting the extension from disk and then reinstalling it help? Alternatively, does manually deleting the extension and then manually downloading and installing it from https://marketplace.visualstudio.com/items?itemName=svelte.svelte-vscode help?

Ballig:
I manually deleted the extension from disk and then reinstalled it from within VS Code - JS/TS language service still crashed.

[...]

PR #1962, opened 2023-03-24
(fix) prevent ts assertion error

dummdidumm:
#1956

Only add the file to the project if it has a projectRootPath, because else a ts.Assert error will be thrown when multiple inferred projects are tried to be merged.

@jasonlyu123 does this look good to you?


jasonlyu123 (approved):
I tested it. It does work in both cases. 👍

Issue #1952, opened 2023-03-22
`svelte-check` should fail when no files are checked


DetachHead:
### Description

`tsc` fails if no files were checked:
```
> tsc
error TS18003: No inputs were found in config file 'C:/project/tsconfig.json'. Specified 'include' paths were '[]' and 'exclude' paths were '[]'.


Found 1 error.
```
however `svelte-check` does not:
```
> svelte-check --tsconfig .\src\tsconfig.json
Debugger attached.

====================================
Loading svelte-check in workspace: c:\project
Getting Svelte diagnostics...

====================================
svelte-check found 0 errors and 0 warnings in 0 files
```

### Proposed solution

fail if there were no files checked, as this almost certainly means there's an issue with your config

gustavojms:
Hey, i'm willing to take this implementation, can it be assigned to me? Also it will be helpful if possible to specify all files that need to be changed to make this enhancement.

dummdidumm:
Sure, go ahead! Change is likely scoped to inside `getDiagnostics` within `packages/svelte-check/src/index.ts`

PR #1964, opened 2023-03-25
Svelte-check fail when no files were found in the path provided

gustavojms:
## Description

Implemented a verifier to check if the path provided has files to check, if there's no file then will throw a error.

Resolve issue #1952

### DIFF
--- a/packages/svelte-check/src/index.ts
+++ b/packages/svelte-check/src/index.ts
@@ -66,6 +66,15 @@ async function getDiagnostics(
             fileCountWithProblems: 0
         };
 
+        const hasFilesToCheck = fs
+            .readdirSync(workspaceUri.fsPath)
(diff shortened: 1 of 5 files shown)


packages/svelte-check/src/index.ts
```diff
@@ -75,6 +76,12 @@ async function getDiagnostics(
 
             let fileHasProblems = false;
 
+            if(!diagnostic.diagnostics.length) {
```
jasonlyu123, line 79: `diagnostic.diagnostics.length` is the diagnostics count of each file, not the files count. Also, you need to run `yarn format`

jasonlyu123 (changes requested):

dummdidumm:
This turned out to be a little trickier than I initially thought, closing in favor of #1974 - thank you

PR #1968, opened 2023-03-31
Align types with Svelte v3.58.0

brunnerh:
Makes `target` property match type change from:
- https://github.com/sveltejs/svelte/pull/7554

### DIFF
--- a/packages/svelte2tsx/svelte-shims.d.ts
+++ b/packages/svelte2tsx/svelte-shims.d.ts
@@ -65,7 +65,7 @@ interface Svelte2TsxComponentConstructorParameters<Props extends {}> {
     /**
      * An HTMLElement to render to. This option is required.
      */
-    target: Element | ShadowRoot;
+    target: Element | Document | ShadowRoot;
     /**
      * A child of `target` to render the component immediately before.
      */


brunnerh:
Releases of the relevant packages/extensions would be appreciated, as right now the mismatch leads to errors:

```svelte
import type { ComponentType } from 'svelte';
import Messages from './messages.svelte';

let component: ComponentType;
component = Messages;
```

```
Type 'typeof Messages__SvelteComponent_' is not assignable to type 'ComponentType<SvelteComponentTyped<any, any, any>>'.
  Types of parameters 'options' and 'options' are incompatible.
    Type 'ComponentConstructorOptions<any>' is not assignable to type 'Svelte2TsxComponentConstructorParameters<{ messages?: string[] | undefined; clear?: (() => void) | undefined; add?: ((message: string) => void) | undefined; }>'.
      Types of property 'target' are incompatible.
        Type 'Element | ShadowRoot | Document' is not assignable to type 'Element | ShadowRoot'.
          Type 'Document' is not assignable to type 'Element | ShadowRoot'.
            Type 'Document' is missing the following properties from type 'ShadowRoot': delegatesFocus, host, mode, slotAssignment, innerHTMLts(2322)
```

Issue #1963, opened 2023-03-24
Typescript server repetitively "initializing language features" causing high CPU usage in extension v107.2.5


mpearce-bain:
### Describe the bug

When working in a Svelte repository with a TS and Svelte file open, VSCode begins rapidly flashing the `initializing JS/TS language features` notification in the bottom LH bar. I also notice the notification bell in the bottom RH bar flickering on and off with a new notification. This drives high CPU usage and noticeably slows down intellisense in my repo.

This behavior doesn't appear to be linked to any user input other than opening both `*.svelte` and `*.ts file`s concurrently, and it will continue indefinitely until the VSCode instance is restarted. Running the `Typescript: Restart Language Server` command will temporarily halt the behavior, but it resumes within 30s afterward. If TS logging is turned on, the logs will grow indefinitely in size (in one instance I had a 2GB log file in <5 minutes).

I'm noticing the following lines appearing over and over in the `Svelte` output tab of my terminal. I am not modifying the `tsconfig.json` file at all, so it's odd to me it's detecting an updated config.

> Reloading ts service at  c:/Users/{my-project-directory} due to config updated
> Trying to load configs for c:/Users/{my-project-directory}

Would anyone be able to suggest a potential next troubleshooting step?

**Troubleshooting Attempts**

- Restart TS Language Server
- Restart VSCode
- Clean Install of VSCode
- Update to latest nightly version of TS

### Reproduction

* Open my Svelte directory with both TS & Svelte  files in VS Code 

    * Unfortunately I haven't been able to reproduce with a `hello-world` project , so I'm aware that this could mean there is an issue with my configs. I'm hoping I might be able to get any guidance or suggestions on next troubleshooting steps)

* Disable all extensions apart from 'Svelte for VS Code'
* Open both a `*.svelte` and `*.ts` simultaneously in your window

Frustratingly, I can't share my full project directory because it's proprietary, but I have tried to narrow down the issue to a subset of causes. What I have noticed is that the error is reproducible using the latest Svelte for VSCode extention `v107.2.5`, but disapears when using any version prior to `v106.0.0` -- even when using the same project directory with no code changes.

### Expected behaviour

* `Initializing JS/TS language features` appears once in the status bar and disappears after several seconds

### System Info

- OS: Windows 10
- IDE: VS Code [details below]
- Svelte VS Code Extension: v107.2.5
 ```
    Version: 1.76.2 (user setup)
[...]

PR #1973, opened 2023-04-04
(fix) check modified time for tsconfig watching

jasonlyu123:
#1963 didn't go with the chokidar approach because we have a virtual file system test for config watching. 

@mpearce-bain It would be great if you could follow the [setup guide](https://github.com/sveltejs/language-tools/blob/master/README.md#setup) to test if this work for you. You can switch to this PR branch after cloning using the [GitHub Pull Requests and Issues](https://marketplace.visualstudio.com/items?itemName=GitHub.vscode-pull-request-github) extension.


dummdidumm (approved):
This change makes sense either way, so I'd lay let's merge and release, then testing is easier.

PR #1983, opened 2023-04-11
feat: Auto-typing support for `entries`

tcc-sejohnson:
See https://github.com/sveltejs/kit/pull/9571. This adds auto-typing support for that. I think?

The behavior should be:
- `entries` is a function returning `Array<RouteParam>`
- `entries` can be returned from any Kit file except `Layout` files

But I can't seem to get the host to work locally, so I'm not sure if this works or not... 👀

### DIFF
--- a/packages/svelte2tsx/src/helpers/sveltekit.ts
+++ b/packages/svelte2tsx/src/helpers/sveltekit.ts
@@ -171,6 +171,23 @@ function upserKitRouteFile(
         insert(pos, inserted);
     }
 
+    // add type to entries function if not explicitly typed
+    const entries = exports.get('entries');
+    if (
+        entries?.type === 'function' &&
+        entries.node.parameters.length === 0 &&
+        !entries.hasTypeDefinition &&
+        !basename.includes('layout')
+    ) {
+        if (!entries.node.type && entries.node.body) {
+            const returnPos = ts.isArrowFunction(entries.node)
+                ? entries.node.equalsGreaterThanToken.getStart()
+                : entries.node.body.getStart();
+            const returnInsertion = surround(`: Array<import('./$types.js').RouteParams> `);
+            insert(returnPos, returnInsertion);
+        }
+    }
+
     // add type to actions variable if not explicitly typed
     const actions = exports.get('actions');
     if (actions?.type === 'var' && !actions.hasTypeDefinition && actions.node.initializer) {
--- a/packages/svelte2tsx/src/helpers/sveltekit.ts
+++ b/packages/svelte2tsx/src/helpers/sveltekit.ts
@@ -183,9 +183,7 @@ function upserKitRouteFile(
             const returnPos = ts.isArrowFunction(entries.node)
                 ? entries.node.equalsGreaterThanToken.getStart()
                 : entries.node.body.getStart();
-            const returnInsertion = surround(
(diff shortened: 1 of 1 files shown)


packages/svelte2tsx/src/helpers/sveltekit.ts
```diff
@@ -171,6 +171,25 @@ function upserKitRouteFile(
...
+        entries.node.parameters.length === 0 &&
+        !entries.hasTypeDefinition &&
+        !basename.includes('layout')
+    ) {
+        if (!entries.node.type && entries.node.body) {
+            const returnPos = ts.isArrowFunction(entries.node)
+                ? entries.node.equalsGreaterThanToken.getStart()
+                : entries.node.body.getStart();
+            const returnInsertion = surround(
+                `: Promise<Array<import('./$types.js').RouteParams>> | Array<import('./$types.js').RouteParams> `
```
dummdidumm, line 187: In the related PR I see `EntryGenerator` being generated, shouldn't we use that one instead?

Also, is `RouteParams` the right one? And in the PR it says it's also valid for `+server.js` - does that have any implications on this?

Rich-Harris: yeah, `RouteParams` seems right to me — this can't be used with `LayoutParams`.

Using `EntryGenerator` is probably better but does that mean we can't release this until https://github.com/sveltejs/kit/pull/9571 is released? because i was thinking we couldn't release https://github.com/sveltejs/kit/pull/9571 until _this_ is released 😅

tcc-sejohnson: ok, think we should be good to go, but please help me make sure those angle brackets are matching correctly. my eyes are crossing. 🤣

dummdidumm: In the related PR `EntryGenerator` is typed as `export type EntryGenerator = () => Promise<Array<RouteParams>> | Array<RouteParams>;` - that means we should be good with just `: ReturnType<import('./$types').EntryGenerator>`

packages/svelte2tsx/src/helpers/sveltekit.ts
```diff
@@ -171,6 +171,25 @@ function upserKitRouteFile(
...
+        entries.node.parameters.length === 0 &&
+        !entries.hasTypeDefinition &&
+        !basename.includes('layout')
+    ) {
+        if (!entries.node.type && entries.node.body) {
+            const returnPos = ts.isArrowFunction(entries.node)
+                ? entries.node.equalsGreaterThanToken.getStart()
+                : entries.node.body.getStart();
+            const returnInsertion = surround(
+                `: Promise<Array<ReturnType<import('./$types.js').EntryGenerator>>> | Array<ReturnType<import('./$types.js').EntryGenerator>> `
```
tcc-sejohnson, line 187:
```suggestion
                `: ReturnType<import('./$types').EntryGenerator> `
```

packages/svelte2tsx/src/helpers/sveltekit.ts
tcc-sejohnson: I do astound myself with my own stupidity from time to time 🤣 Implemented.

dummdidumm (approved):
🎉

Issue #2008, opened 2023-05-09
svelte-language-server watcher doesn't work in neovim lspconfig


ryoppippi:
I'm in the same problem with my sveltekit project.
I'm using `typescript-svelte-plugin`.
Although typescript files (like +page.server.ts) works realtime, diagnostics of `.svelte` file does not updated.	

I'm configuring `root_pattern(  {"package.json", "tsconfig.json", "tsconfig.jsonc", "node_modules"})` for svelte language server

https://user-images.githubusercontent.com/1560508/237040552-0a1be152-0c2b-45ab-b98c-3e927ffeae34.mov

_Originally posted by @ryoppippi in https://github.com/sveltejs/language-tools/issues/1415#issuecomment-1539681042_

ryoppippi:
Okay, this is a truly hacky things, but I solved this problem with this setting

``` lua
function M.on_attach(on_attach)
  vim.api.nvim_create_autocmd("LspAttach", {
    callback = function(args)
      local buffer = args.buf
      local client = vim.lsp.get_client_by_id(args.data.client_id)
      on_attach(client, buffer)
    end,
  })
end

on_attach(function(client, bufnr)
      vim.api.nvim_create_autocmd("BufWritePost", {
        pattern = "*",
        callback = function(ctx)
          if client.name == "svelte" then
            client.notify("$/onDidChangeTsOrJsFile", { uri = ctx.file })
          end
        end,
      })
    end)
end)
```

kamalkech:
@ryoppippi that s OK for me too, are u found another solution more better ??

jasonlyu123:
This is fixed in language-server 0.15.12. You don't need the workaround.

jasonlyu123:
The workaround only solves part of the problem. The API used in the workaround was meant to assist file watching instead of replacing it. What is your neovim version? This issue is about a problem introduced in neovim 0.9.0.

kamalkech:
@jasonlyu123 
NVIM v0.9.0
Build type: Release
LuaJIT 2.1.0-beta3

ryoppippi:
Not working in my env either

jasonlyu123:
Do keep in mind that "file watchers" require files to be saved. And the  `$/onDidChangeTsOrJsFile` is for syncing changes without file saving. You still need the custom code if you want it to sync without file save. I am not sure which problem are seeing. But I just tested it again and file watching does trigger on neovim with the latest version.

joriskoris:
[...]

PR #2009, opened 2023-05-10
(fix) move fs watcher config to language-server

### LINKED ISSUE #1613: Language server does not refresh JSON imported with resolveJsonModule
### Describe the bug

After importing some data from JSON with typescript's resolveJsonModule, when you update the JSON file language server does not refresh it - running `Restart language server` helps.

### Reproduction

1. Create a JSON file with some values in it.
2. Import the file in Svelte component
3. Introduce a new key to JSON file
4. Try to use it in code - newly introduced key is not found.

### Expected behaviour

Changes in JSON file should be watched in real time.

### System Info

- OS: Windows
- IDE: VSCode 1.70.2
- Svelte for VS Code v105.22.1

svelte-language-server

JSON file:
```json
{
    "test": {
        "val": "test"
    }
}
```

image

After Command Pallete > Svelte: Restart language server:
image

jasonlyu123:
#2008 

The problem seems to be becuase neovim recently added the `didChangeWatchedFiles` support. But they didn't have a watcher config on their side. This PR move the watcher config to the language server with dynamic registration. It was on the client side because it can't be set in the server as a capability. 

Existing clients with the config I know of are VSCode, coc-svelte, visual studio and emacs-lsp. I tested with them to see if moving to the server side would cause any problems. VSCode and coc-svelte can handle both and will not trigger twice. Visual Studio doesn't support dynamic registration this won't affect it. I didn't manage to install emacs-lsp successfully. But the Volar client in emacs does have this double registration problem, and it doesn't seem to have an existing issue, so it's probably fine. 

I also added json to the watcher config. So this should fix #1613 as well. In resolveJsonModule, json file is just like ts, js files and in normal mode, We'll still cache some JSON files because TypeScript reads them through `languageServiceHost.readFile`. We can refresh those in the file watch as well.

### DIFF
--- a/packages/language-server/src/server.ts
+++ b/packages/language-server/src/server.ts
@@ -21,7 +21,8 @@ import {
     CallHierarchyOutgoingCallsRequest,
     InlayHintRequest,
     SemanticTokensRefreshRequest,
-    InlayHintRefreshRequest
+    InlayHintRefreshRequest,
+    DidChangeWatchedFilesNotification
 } from 'vscode-languageserver';
 import { IPCMessageReader, IPCMessageWriter, createConnection } from 'vscode-languageserver/node';
 import { DiagnosticsManager } from './lib/DiagnosticsManager';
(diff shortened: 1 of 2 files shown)


dummdidumm:
> will not trigger twice

When and where would it trigger twice, or rather, where did you fear it would and why?

jasonlyu123:
I am just worried that some clients have the watch config as we do. And now we have added an extra one to the server. It will trigger twice. But triggering twice is probably fine, just a slight performance penalty. This won't affect VSCode since I removed it, testing in VSCode is just trying to see how clients might react to it.

dummdidumm:
Aah I understand, so the worry is that language clients which have a watcher config on the client side need to adjust/remove that config if they want to use the latest language server version and not have file updated triggered twice. I'm ok with this, especially since it doesn't seem to make a difference (only in performance a bit) in your testing.

dummdidumm (approved):

PR #2012, opened 2023-05-11
feat: add `machine-verbose` to the output options

### LINKED ISSUE #2011: Add end errors in machine readable svelte-check logs
### Description

Currently the errors and warnings reported by svelte check just mention the start of the error/warning.

It would be useful to also have the end of the error warning (and maybe the line:column representation too). We could specifically use this in SvelteLab.dev to integrate with codemirror diagnostic and show errors and warnings from svelte-check.

### Proposed solution

Add all of that to the machine readable output. To avoid make this a breaking change this could be behind a flag.

### Alternatives

_No response_

I will work on this and prepare a PR, I'm opening this issue so that we can reference it and for someone to stop me before I do it if that's not wanted

paoloricciuti:
Closes #2011 

It adds an option `machine-verbose` to the output options that provide a more complete info on the diagnostics


packages/svelte-check/src/writers.ts
```diff
@@ -159,10 +163,23 @@ export class MachineFriendlyWriter implements Writer {
...
-                this.log(`${type} ${fn} ${line + 1}:${character + 1} ${msg}`);
+                if (this.isVerbose) {
+                    const stringifiedCodeDescription = JSON.stringify(codeDescription);
+                    const stringifiedSource = JSON.stringify(source);
+                    this.log(
+                        `${type} ${fn} ${startLine + 1}:${startCharacter + 1} ${endLine + 1}:${
+                            endCharacter + 1
+                        } ${msg} "${code ?? ''}" ${stringifiedCodeDescription ?? '""'} ${
+                            stringifiedSource ?? '""'
+                        }`
```
gtm-nayan, line 178: Something like the ndjson format would be easier for a machine to parse reliably compared to an arbitrary space separated line.

paoloricciuti: I'm totally ok with it, I was sticking to that to keep it similar to the `machine` version...I can change that

paoloricciuti: Obviously let me know if that's what we want

gtm-nayan: Wait for a maintainer to confirm, but a standardized format would be my recommendation. I don't think you'll need a separate dependency for it, just `JSON.stringify(thing)` without any indentation should output it on a single line that complies with ndjson.

paoloricciuti:
> Honestly I'm not sure that I want yet another format. What's the use case for this?

As i've specified in the linked issue we could use this in sveltelab to show errors inside the codemirror editor. Other than that i think having a more complete rundown of the errors/warnings could be beneficial for every application that wants to interact with `svelte-check` on a more granular level. Also is non breaking and quite small...i know i'm biased because i need this but honestly i'd go with it :D

paoloricciuti:
> Honestly I'm not sure that I want yet another format. What's the use case for this?

And if with "another format" you are referring to ndjson is the reason i've implemented it similar to the current `machine` readable format. I'm totally ok with both formats

PuruVJ:
Huh I think this can benefit svelte REPL as well. Although TBH, until now I didn't even know such an option existed with those kind of values, so I'm not fully knowledgable on it 😅

dummdidumm:
"Yet another format" == "a fourth format". Machine readable was already a concession. Can you extract the info from the human output? It's also a structured in a way.

paoloricciuti:
> Although not ideal, you can extract the diagnostic ends from human verbose because the error range is coloured in ANSI escape codes. You can also get the source of the error/warning. The code is not, though, but I think it makes sense to add it.

Mmm I see...well that would be kinda of a hell to parse. I'm willing to do it but I would strongly prefer not to, especially because this seems like a simple non breaking addition

SarcevicAntonio:
Maybe we are doing something wrong, but if we work together should be able to figure out what is missing for making the svelte language tools easier to adapt in projects like the new REPL and SvelteLab, right?

SarcevicAntonio:
> Machine readable was already a concession.

What was it added for? @dummdidumm do you have the issue/pr, so we can read up? 😅

paoloricciuti:
> > Machine readable was already a concession.
> 
> What was it added for? @dummdidumm do you have the issue/pr, so we can read up? 😅

This should be the issue #124 and this the pr #126

paoloricciuti:
Small update on this 

> Although not ideal, you can extract the diagnostic ends from human verbose because the error range is coloured in ANSI escape codes. You can also get the source of the error/warning. The code is not, though, but I think it makes sense to add it.

I kinda got it working but right now I've discovered that errors and warnings use slightly different coloring methods which make it very difficult to reason coherently...I'll give it another shot in the meantime but this is honestly a pain to write and I can only imagine the pain it would be to maintain. Again, I would greatly appreciate this pr even so because SvelteLab will not be the only one to benefit from this since with this every tools that would like to integrate with svelte-check could do it in a way easier way.

@dummdidumm if you really really really don't want to add a 4th option could you be ok with just add the information in the machine one? I was avoiding that because maybe someone relies on that specific pattern so that would be a breaking change.

dummdidumm:
Let's go with @gtm-nayan's proposal and make this JSON strings, that way don't have the possibility of breaking changes in the future

paoloricciuti:
> Let's go with @gtm-nayan's proposal and make this JSON strings, that way don't have the possibility of breaking changes in the future

Aight thank you very much...i'll make the change rn

paoloricciuti:
I've swapped the output...i've decided to not stringify the whole Diagnostic because there were some arrays that maybe could really bloat the output. @dummdidumm let me know if you prefer to include everything and just stringify the whole Diagnostic at this point.

packages/svelte-check/src/writers.ts
```diff
@@ -159,10 +163,25 @@ export class MachineFriendlyWriter implements Writer {
...
+                            message,
+                            code,
+                            codeDescription,
+                            source
+                        })
+                    );
+                } else {
+                    this.log(`${type} ${fn} ${start.line + 1}:${start.character + 1} ${msg}`);
+                }
             }
```
gtm-nayan, line 185:
```suggestion
                const { start, end } = range;
                if (this.isVerbose) {
                    this.log(
                        JSON.stringify({
                            type,
                            filename,
                            start,
                            end,
                            message,
                            code,
                            codeDescription,
                            source
                        })
                    );
                } else {
                    const fn = JSON.stringify(filename);
                    const msg = JSON.stringify(message);
                    this.log(`${type} ${fn} ${start.line + 1}:${start.character + 1} ${msg}`);
                }
            }
```

gtm-nayan: Holy hell, that's a lot taller than I thought it'd be.

TL;DR: Move the `JSON.stringify`s inside the else branch

paoloricciuti: Good suggestion...did it

packages/svelte-check/README.md
```diff
@@ -103,6 +103,16 @@ to the workspace directory. The filename and the message are both wrapped in quo
 1590680326778 WARNING "imported-file.svelte" 0:37 "Component has unused export property 'prop'. If it is for external reference only, please consider using `export const prop`"
 ```
 
+If the argument is `machine-verbose` it will tell us the filename, the starting line and column numbers, the ending line and column numbers, the error message, the code of diagnostic, the human-friendly description of the code and the human-friendly source of the diagnostic (eg. svelte/typescript). The filename is relative to the workspace directory. All the information's are represented as [ndjson](https://en.wikipedia.org/wiki/JSON_streaming#Newline-Delimited_JSON)
```
gtm-nayan, line 106:
```suggestion
If the argument is `machine-verbose` it will tell us the filename, the starting line and column numbers, the ending line and column numbers, the error message, the code of diagnostic, the human-friendly description of the code and the human-friendly source of the diagnostic (eg. svelte/typescript). The filename is relative to the workspace directory. Each diagnostic is represented as an [ndjson](https://en.wikipedia.org/wiki/JSON_streaming#Newline-Delimited_JSON) line prefixed by _____
```

I'm assuming the numbers at the beginning are a timestamp of the log?

paoloricciuti:
@dummdidumm are those modifications ok?

dummdidumm (approved):
Thank you!

dummdidumm:
Could you fix the lint error (formatting)? Then we're good to merge.

paoloricciuti:
> Could you fix the lint error (formatting)? Then we're good to merge.

Sure...right on it

paoloricciuti:
@dummdidumm should be fixed now

Issue #2018, opened 2023-05-16
Cannot find namespace 'svelteHTML' error toggled on and off by svelte-kit sync


connerdassen:
### Describe the bug

When trying to use the svelteHTML namespace I noticed I occasionally got this error:
 `Cannot find namespace 'svelteHTML' .ts(2503)`

However, the project built normally and `svelte-check` didn't report any errors.
After investigating, I noticed that the error went away and was reintroduced with every run of `svelte-check`

### Reproduction

1. `npm create svelte@latest my-app`
    - Select `Skeleton project`, `Yes, using TypeScript syntax`,  no additional options
2. `cd my-app`
3. `pnpm install`
4. Create `src/test.ts` 
    - Write `type Test = svelteHTML.HTMLAttributes;`

At this point, you may or may not get the error  `Cannot find namespace 'svelteHTML'  .ts(2503)`.
Running `npx svelte-kit sync` toggles the error on / off every time I run it: 

https://github.com/sveltejs/language-tools/assets/33068114/f1612cd8-c8b1-4f5b-a019-a0a722ce5d75

### Expected behaviour

svelteHTML is defined

### System Info

- OS: Windows
- IDE: VSCode 
- Svelte for VSCode: v107.4.3

### Which package is the issue about?

Svelte for VS Code extension

connerdassen:
`npm run dev` which runs `vite dev --port 3000 --open` also toggles the error just like `npx svelte-kit sync`...
`npm run build` toggles the error twice, first when `vite build` runs and then again when `npm run package` runs

jasonlyu123:
We found a compatibility issue with the [IntelliCode](https://marketplace.visualstudio.com/items?itemName=VisualStudioExptTeam.vscodeintellicode) extension. Do you have this installed? I want to make sure that the problem we found is the same as yours. If not, can you check what other extension might have caused the problem?  You can clone a new settings profile from the existing one, so it's easier to revert after testing.

cmjoseph07:
@jasonlyu123 
This seems to have resolved my issue. If I uninstall the extension and run vitest I am not running into this error anymore. Thank you so much for reaching out to me on this!!

connerdassen:
@jasonlyu123 sorry for the late response, disabling IntelliCode did indeed fix it

rutinerad:
For anyone else who has this issue, simply restarting VSCode can help.

PR #2019, opened 2023-05-17
(fix) prevent ts plugin crash when config updated

jasonlyu123:
#2018 https://github.com/sveltejs/kit/issues/9932

This happens when tsconfig or extending tsconfig updates and another ts-plugin returns another language service object. For example, https://marketplace.visualstudio.com/items?itemName=VisualStudioExptTeam.vscodeintellicode. Because the patched symbol is no longer there, patching will run again. And when we call `project.addRoot`, a Debug assertion error will be thrown as the file is already a root file. TypeScript will then removes the external files because the plugin failed to initialise.

### DIFF
--- a/packages/typescript-plugin/src/language-service/index.ts
+++ b/packages/typescript-plugin/src/language-service/index.ts
@@ -15,10 +15,10 @@ import { decorateRename } from './rename';
 import { decorateUpdateImports } from './update-imports';
 import { decorateLanguageServiceHost } from './host';
 
-const sveltePluginPatchSymbol = Symbol('sveltePluginPatchSymbol');
+const patchedProject = new Set<string>();
 
-export function isPatched(ls: ts.LanguageService) {
-    return (ls as any)[sveltePluginPatchSymbol] === true;
+export function isPatched(project: ts.server.Project) {
+    return patchedProject.has(project.getProjectName());
 }
 
 export function decorateLanguageService(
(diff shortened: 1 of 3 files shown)


packages/typescript-plugin/src/language-service/index.ts
```diff
@@ -64,11 +69,6 @@ function createProxyHandler(configManager: ConfigManager): ProxyHandler<ts.Langu
 
     return {
         get(target, p) {
-            // always return patch symbol whether the plugin is enabled or not
-            if (p === sveltePluginPatchSymbol) {
-                return true;
-            }
-
```
dummdidumm, line 71: Is the `if (!configManager.getConfig().enable || p === 'dispose') {` below still correct?

dummdidumm:
> This happens when tsconfig or extending tsconfig updates and another ts-plugin returns another language service object

another language service object which still has our patches applied? Because if not, what would happen how?

packages/typescript-plugin/src/language-service/index.ts
jasonlyu123: Hmm. I don't remember why `dispose` has special treatment here. But it seems wrong.

jasonlyu123:
>Another language service object which still has our patches applied. 

Yes. But not the same object. Debugging show that they use `Object.keys()` to copy over unpatched method/property to the new object. So the symbol is not copied over.

dummdidumm:
Do they just copy over everything? In other words, if we didn't use a symbol but a `__svelte-language-tools__` property it would also work? I guess we're somewhat stuck between false positives or false negatives, and the question is which is worse.

jasonlyu123:
> if we didn't use a symbol but a `__svelte-language-tools__` property it would also work?

Yes. But TypeScript will copy over that property no matter if the plugin has extended our patch (proxy Inception lol).
https://github.com/microsoft/TypeScript/blob/24ac9e75c2f04f9084fcd21df7d3fabc7dee9fe0/src/server/project.ts#L2005

Once we add that property, it'll be there until the language service is destroyed. So it's not much better than checking patched project names. But it's a more direct check. I can change to that if you prefer that.

packages/typescript-plugin/src/language-service/index.ts
jasonlyu123: I remember why `dispose` has special treatment here now. It should run the decorated version even if the typescript-plugin is disabled, but the current check is wrong.

jasonlyu123:
@Princesseuh FYI, Astro might also have this problem.

PR #2020, opened 2023-05-17
(feat) implement generics attr on script tags

dummdidumm:
https://github.com/sveltejs/rfcs/pull/38#issuecomment-1486353966

### DIFF
--- a/packages/language-server/src/plugins/html/dataProvider.ts
+++ b/packages/language-server/src/plugins/html/dataProvider.ts
@@ -363,6 +363,13 @@ const addAttributes: Record<string, IAttributeData[]> = {
         {
             name: 'bind:open'
         }
+    ],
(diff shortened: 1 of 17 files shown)


jasonlyu123:
Overall looks good. There are two minor issues I found. Do we need an equivalent `throwIfIsGeneric` for the generic attribute? You can define generic on the module script now, but it doesn't do anything. I also notice that semantic tokens sometimes highlight `script lang="ts" generic` as a function. The current check needs to be more. I think we can skip the highlight if the highlighted identifier is `render` and if the mapped source includes a script tag.

dummdidumm:
> I also notice that semantic tokens sometimes highlight script lang="ts" generic as a function.

Do you have a code snippet that reproduces this?

PR #2026, opened 2023-05-25
chore: add packageManager to root package.json to support corepack

dominikg:
needed for svelte-ecosytem-ci  (and helps users that use corepack on their local machines)


dummdidumm:
Closing in favor of #2027

PR #2027, opened 2023-05-25
chore: switch to pnpm 

dummdidumm:
Svelte repos use it, we should, too


.github/workflows/DeployExtensionsProd.yml
```diff
@@ -11,37 +11,41 @@ jobs:
...
 
       # Lets us use one-liner JSON manipulations the package.json files
       - run: "npm install -g json"
 
       # Setup the environment
-      - run: 'json -I -f packages/svelte-vscode/package.json -e "this.version=\`${{ github.ref }}\`.split(\`-\`).pop()"'
+      - run: json -I -f packages/svelte-vscode/package.json -e "this.version=\`${{ github.ref }}\`.split(\`-\`).pop()"
 
-      # To deploy we need isolated node_modules folders which yarn won't do because it is a workspace
+      # To deploy we need isolated node_modules folders which pnpm won't do because it is a workspace
```
dominikg, line 32: what exactly is this about?  sounds like `pnpm deploy` could help you. running `npm install` (see below) would get very different versions of the dependencies...

dummdidumm: Yeah that's intended. We want to get the latest minors of everything we specified

gtm-nayan: When I tried it last time, pnpm deploy still left symlinks in the node_modules folder which the VSIX format can't handle.

dominikg: if we can't use symlinks, then we should use `node-linker=hoisted` instead. or stick with npm .  the way it's done now basically means that whatever we tested is not what gets released. :thinking:

dominikg: i tried to use `pnpm --config.node-linker=hoisted --filter svelte-vscode deploy --prod output-dir` and also tried symlink=false and package import copy, didn't work as intendent. the only other option i can think of with pnpm deploy would be to resolve the symlinks on the fs level after it is done.

gtm-nayan: `node-linker=hoisted` didn't work either for me because. IIRC it was because it's in a workspace, the npmrc doesn't apply if it's not at the workspace root, but putting it in the workspace root had some weird changes that I don't fully remember.

FWIW I looked at how Vue does it and they just bundle it with esbuild instead.

https://github.com/vuejs/language-tools/commit/ab1fafcfc0033f62cdf5104f027d37b9a11d4a14

dummdidumm: ...ooooor we just use npm, as done here, because it works as I want it to

gtm-nayan: You've changed some of the specifiers to use the `workspace:` protocol, does `npm install` resolve them correctly?

dummdidumm: In the yaml script I modify the package.json to change it back to `*` so npm knows what to do

PR #2034, opened 2023-05-26
Align templates with "zero effort typing"

stephane-vanraes:
With the introduction of "zero effort typing" the generated templates from the Svelt Extension no longer need the types, this change reflects that by removing these types from them.


jasonlyu123:
Can we make this configurable? Currently typescript eslint doesn't support zero config type (#2073) and there is also #2053 that we haven't figure out the exact reason. Also, these types are not redundant some people might prefer the explicit of always type it.

dummdidumm:
Agree that this should be configurable - there should be a config option to omit the types (disabled by default).

PR #2041, opened 2023-06-02
fix build and test with svelte 4

jasonlyu123:
https://github.com/Rich-Harris/estree-walker/pull/34 changes the type of the parameter of walk from BaseNode to Node. It should be fine in the runtime, but this is technically not what estree-walker is designed for. Not sure if it makes sense to revert it back in estree-walker.

Bumping to svelte 4 only to test it out in CI. I'll revert it before merging.


jasonlyu123:
I made some of the test runs in both versions. There are still some failures in v4. Reason being:

1. SvelteComponent type changes to be more strict with required props
2. each now allows Iterable
3. svelte.JSX namespace support is dropped when using Svelte 4.

I'll merge this for now.  We can think about testing the difference between both versions later.

PR #2044, opened 2023-06-05
add support for resize observer bindings

PatrickG:
Related https://github.com/sveltejs/svelte/pull/8022

I am not sure if that is all that needs to be done.

### DIFF
--- a/packages/language-server/src/plugins/html/dataProvider.ts
+++ b/packages/language-server/src/plugins/html/dataProvider.ts
@@ -65,6 +65,22 @@ const svelteAttributes: IAttributeData[] = [
         description: 'Available for block level elements. (read-only)'
     },
     {
+        name: 'contentRect',
+        description: 'Available for all elements. (read-only)'
+    },
+    {
+        name: 'contentBoxSize',
+        description: 'Available for all elements. (read-only)'
+    },
+    {
+        name: 'borderBoxSize',
+        description: 'Available for all elements. (read-only)'
+    },
+    {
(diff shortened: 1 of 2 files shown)


packages/language-server/src/plugins/html/dataProvider.ts
```diff
@@ -64,6 +64,22 @@ const svelteAttributes: IAttributeData[] = [
...
+        description: 'Available for all elements. (read-only)'
+    },
+    {
+        name: 'borderBoxSize',
+        description: 'Available for all elements. (read-only)'
+    },
+    {
+        name: 'devicePixelContentBoxSize',
+        description: 'Available for all elements. (read-only)'
+    },
```
ghostdevv, line 82:
```suggestion
    {
        name: 'bind:contentRect',
        description: 'Available for all elements. (read-only)'
    },
    {
        name: 'bind:contentBoxSize',
        description: 'Available for all elements. (read-only)'
    },
    {
        name: 'bind:borderBoxSize',
        description: 'Available for all elements. (read-only)'
    },
    {
        name: 'bind:devicePixelContentBoxSize',
        description: 'Available for all elements. (read-only)'
    },
```

PatrickG: Of course 🤦‍♂️

jasonlyu123 (approved):
That should be enough. Thanks! The type definition now lives in the Svelte repo, and it's already updated in the original PR.

PR #2047, opened 2023-06-07
Add auto-complete for transition and animate functions

navorite:
This adds auto-completion for all built-in transition and animate functions.

 Could be probably improved by auto-importing the functions on completion? Not sure if it's possible.

### DIFF
--- a/packages/language-server/src/plugins/html/dataProvider.ts
+++ b/packages/language-server/src/plugins/html/dataProvider.ts
@@ -68,6 +68,59 @@ const svelteAttributes: IAttributeData[] = [
         name: 'bind:this',
         description:
             'To get a reference to a DOM node, use bind:this. If used on a component, gets a reference to that component instance.'
+    },
+    {
+        name: 'transition:blur',
+        description: "Animates a blur filter alongside an element's opacity."
+    },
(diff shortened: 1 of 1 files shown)


jasonlyu123:
I am leaning toward only having `in:`, `out:`, `transition:` and even `use:` here. That would also benefit people with custom transition/animation/action. And if we want to have a complete list. We add that in the `CompletionProvider`. And add auto-import, possibly with typescript API. Without auto-import, it doesn't save many key-stroke, and I'll still have to go back a few characters to trigger completion/code-action. At least for me, it just feels worse. But that would be way harder to implement thought. @dummdidumm What do you think?

dummdidumm:
I agree, we should only have the `in:` etc prefixes and no specific fully specified names, will probably only get in the way most of the time.

jasonlyu123:
Ah. Right. If there is only the directive keyword. We can't rely on `dataProvider`. vscode-html-language-service would insert `=""` after that. We need to do it on our own instead. Like something here https://github.com/sveltejs/language-tools/blob/master/packages/language-server/src/plugins/svelte/features/getCompletions.ts#L59

jasonlyu123:
Closing because we went for the "only keyword" route

Issue #2050, opened 2023-06-10
Diagnostics only show after edit


WoodyWoodsta:
### Describe the bug

Using `svelte-check`, I can see TS errors for the rule `noUncheckedIndexedAccess` in a svelte file. When I open the file, the errors do not show until I edit the code.

### Reproduction

Any code which fails `noUncheckedIndexedAccess`.

### Expected behaviour

All errors should show when I open a file I know has errors.

### System Info

- OS: macOS Ventura
- IDE: VScode 1.79.0
- Svelte plugin: 107.5.1

### Which package is the issue about?

Svelte for VS Code extension

### Additional Information, eg. Screenshots

Once the errors are showing, if I close and open the file again, the errors are still present. If I restart VSCode, the issue presents.

https://github.com/sveltejs/language-tools/assets/8268040/bf689d83-ed2b-4d11-b4de-ad5ad9c2315d

PR #2055, opened 2023-06-14
(fix) diagnostics throttle shouldn't cancel other files

jasonlyu123:
#2050 

Side note: we can switch to the diagnostics pull model later. If so, these changes might only be used in clients not supporting the pull model.


dummdidumm (approved):
👍

PR #2060, opened 2023-06-16
Sync compiled code preview with open svelte file

### LINKED ISSUE #831: Synchronizing view of .svelte code with view of .svelte.js (compiled) code
Now I need to write `svelte.showCompiledCodeToSide` command everytime I switch to another file to be able to see compiled code of that file.

When I open compiled version of code by `svelte.showCompiledCodeToSide` command, it would be nice if it would automatically change that compiled version when I switch to another .svelte file.

**Describe alternatives you've considered**
-

**Additional context**
-

ghostdevv:
Closes #831

I had to change how the preview window worked, previously it created a window for the current svelte file with the uri `svelte-compiled://FILE_PATH` (where FILE_PATH is the file path). Now whenever the command is run it creates an editor with the uri `svelte-compiled://preview.js`. I added comments to the code so it's easier to understand, since the file hasn't been touched in three years.

The only downside to the uri change is that the file path isn't in the editor label anymore. Unfortunately unless it's a webview you can't change the editor label so I added code (4bff895e6fb796b60bb15bfea0e9d87ac557fa4f) to make a comment in the compiled output to indicate what file is actually being compiled. I've based it's behaviour on the Markdown Preview window as much as possible, within the limitations of an editor.

### DIFF
--- a/packages/svelte-vscode/src/CompiledCodeContentProvider.ts
+++ b/packages/svelte-vscode/src/CompiledCodeContentProvider.ts
@@ -71,7 +71,7 @@ export default class CompiledCodeContentProvider implements TextDocumentContentP
     // we can trigger this using the didChangeEmitter
     async provideTextDocumentContent(): Promise<string | undefined> {
         // If there is no selected svelte file, try to get it from the activeTextEditor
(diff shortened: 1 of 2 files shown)


dummdidumm (approved):
Nice!

Issue #2007, opened 2023-05-09
Test failures on  language-server-0.15.10


polyzen:
### Describe the bug

```
  1) CallHierarchyProvider
       can prepare call hierarchy:
     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/build/svelte-language-server/src/language-tools-language-server-0.15.10/packages/language-server/tes
t/plugins/typescript/features/CallHierarchyProvider.test.ts)
      at processTicksAndRejections (node:internal/process/task_queues:95:5)

  2) FindComponentReferencesProvider
       finds component references:
     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/build/svelte-language-server/src/language-tools-language-server-0.15.10/packages/language-server/tes
t/plugins/typescript/features/FindComponentReferencesProvider.test.ts)
      at processTicksAndRejections (node:internal/process/task_queues:95:5)

  3) TypeScript Plugin Performance Tests
       should be fast enough:
     Error: Timeout of 15000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/build/svelte-language-server/src/language-tools-language-server-0.15.10/packages/language-server/te
st/plugins/typescript/typescript-performance.test.ts)
```

`CallHierarchyProvider` passed with Node.js 20.1.0

### Reproduction

This is the build script used:
https://github.com/archlinux/svntogit-community/blob/packages/svelte-language-server/trunk/PKGBUILD

### Expected behaviour

Tests pass

### System Info

- OS: Arch Linux
- Node.js: 19.9.0/20.1.0
- language-server-0.15.10

### Which package is the issue about?

svelte-language-server

jasonlyu123:
These tests are usually timeouts because they literally take too long. Maybe we should increase the timeout since some of it failed in our CI occasionally and might fail consistently in less performant machines. What do you think @dummdidumm?

The CallHierarchyProvider might be a new thought; it only started to fail on my machine recently. Adding a `perf` tag for that.

dummdidumm:
I'm a bit split on increasing the timeout. If it consistently starts happening and did not before then that hints at a performance regression - though we probably rarely are the culprit, so maybe it's ok increasing it? I'm not sure ..

jasonlyu123:
[...]

PR #2064, opened 2023-06-22
(chore) warmup ts service before running test

jasonlyu123:
#2007

Our typescript-related tests usually time out because the typescript's internal building takes too long. This makes the test frequently fail in less performant machines and in `test.only` during development. This PR moves the internal building to the `before` hook with a 5000ms timeout. After moving the building out, most of the tests should be able to finish within 1s. So I also dropped the snapshot test timeout back to default. If we want to test potential performance degradation, we could also drop the timeout to 1500ms or even 1000ms in CI.


dummdidumm:
Nice ideas, let's hope this solves this!

Issue #1988, opened 2023-04-17
Inlay hint in `#each` block at the wrong position


blaumeise20:
### Describe the bug

I am very satisfied with the current state of Svelte tooling in VS Code, but I noticed one little problem in the inlay hints for `#each` blocks.

### Reproduction

Create a new file `test.svelte` and paste this content inside:
```svelte
{#each ["a", "b", "c"] as item, index}
    {index}: {item}
{/each}
```
You can see that the inlay hint for `index` is positioned in the wrong place, and `item` has nothing at all.

### Expected behaviour

The first line should show up as this:
```svelte
{#each ["a", "b", "c"] as item[: string], index[: number]}
```

### System Info

- OS: macOS 12.6
- IDE: VSCode 1.73.0

### Which package is the issue about?

Svelte for VS Code extension

### Additional Information, eg. Screenshots

<img width="407" alt="Bildschirmfoto 2023-04-17 um 12 33 54" src="https://user-images.githubusercontent.com/62756994/232460083-ccba6ba0-9753-46bf-a733-1598a9ec504d.png">

PR #2070, opened 2023-06-26
fix: supress inlay hint for each block's index variable

BlueGreenMagick:
#1988 

Inlay hints for each block's index variable appears at a wrong position, 1 character before the proper position.
```
{#each items as item, inde[: number]x}
```

This PR suppresses the inlay hint by explicitly writing type annotation for the index variable.

<br />

### Side note

I initially attempted to fix the positioning code, but gave up as fixing that part proved to be too complex. Below is my investigation into the cause.

The problem code is in the `transform` function in `packages/svelte2tsx/src/htmlxtojsx_v2/utils/node-utils.ts`. 
https://github.com/sveltejs/language-tools/blob/0f2396e1a567040f5a33a27d6ddd98712a81f0b9/packages/svelte2tsx/src/htmlxtojsx_v2/utils/node-utils.ts#L57-L76

If a range (given in format [number, number]) `tEnd === end - 1`, the following line isn't run: `str.overwrite(tEnd - 1, tEnd, overwrite, { contentOnly: true });`

So the string `= 1;` inserted afterward is mapped to the last character of `index` variable, instead of the character right after. In generated code, inlay hint is positioned at the character right after `index`, which is mapped to `x`.

`tEnd < end - 1` exists because `str.move(tStart, end, end)` is not valid. So to fix it, we'd have to overhaul the entire transform logic, which moves everything to the end.

### DIFF
--- a/packages/language-server/test/plugins/typescript/features/inlayHints/fixtures/each/expectedv2.json
+++ b/packages/language-server/test/plugins/typescript/features/inlayHints/fixtures/each/expectedv2.json
@@ -1,0 +1,11 @@
+[
+  {
+    "kind": 1,
+    "label": ": number[]",
+    "paddingLeft": true,
+    "position": {
+      "character": 11,
+      "line": 1
+    }
+  }
+]
--- a/packages/language-server/test/plugins/typescript/features/inlayHints/fixtures/each/input.svelte
+++ b/packages/language-server/test/plugins/typescript/features/inlayHints/fixtures/each/input.svelte
@@ -1,0 +1,7 @@
+<script lang="ts">
+  let items = [1]
+</script>
+
+{#each items as item, i}
+  {item}
+{/each}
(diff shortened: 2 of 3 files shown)


jasonlyu123:
I would prefer to solve this in the language server. This off-by-one situation exists in multiple places. Don't know if other situations would show up in inlay hints. An easy way is to use regex to check if the kind of the inlay hint is `InlayHintKind.Type` and it's placed before a word that might be a js identifier. Or we can find the AST node and uses its `start + length` to map. Also only for `InlayHintKind.Type`. We are already finding the node in the filtering part https://github.com/sveltejs/language-tools/blob/0f2396e1a567040f5a33a27d6ddd98712a81f0b9/packages/language-server/src/plugins/typescript/features/InlayHintProvider.ts#L154. We can get the info by restructuring the code to make the related node available for mapping.

BlueGreenMagick:
> An easy way is to use regex to check if the kind of the inlay hint is `InlayHintKind.Type` and it's placed before a word that might be a js identifier.

Done. I opened a new PR since it's a completely different implementation. (https://github.com/sveltejs/language-tools/pull/2070)

I don't think there are any other instances of inlay hint position being off by one, but I suppose it doesn't hurt to be sure.

Issue #2065, opened 2023-06-23
Slow update-import with declarationMap 


jasonlyu123:
### Describe the bug

Update import takes a very long time if a library has declaration maps. For example, svelte 4

### Reproduction

Our update import test running with Svelte 4

### Expected behaviour

It should be fast enough. 

### System Info

- OS: [e.g. Windows]
- IDE: [e.g. VSCode, Atom]

FedericoBiccheddu:
FYI, this happens even with Svelte 3.

PR #2071, opened 2023-06-27
perf: prevent multiple recompiles during update import

jasonlyu123:
#2065
We currently implement `LanguageServiceHost.readFile` with `getSnapshot.` But this causes Typescript to recompile multiple times during update import if there're declaration maps. Ideally, this method should be directly reading from the file system. But this would result in way too many IO operations. . tsserver doesn't have this problem because they cache more stuff with some internal APIs. So this PR adds a package.json cache as a middle ground.

### DIFF
--- a/docs/preprocessors/typescript.md
+++ b/docs/preprocessors/typescript.md
@@ -150,7 +150,7 @@ You may need to set `baseUrl` in `tsconfig.json` at the project root to include 
 
 ### I'm using an attribute/event on a DOM element and it throws a type error
 
-If it's a non-experimental standard attribute/event, this may very well be a missing typing from our [HTML typings](https://github.com/sveltejs/svelte/blob/master/elements/index.d.ts). In that case, you are welcome to open an issue and/or a PR fixing it.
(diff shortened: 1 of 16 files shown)


dummdidumm (commented):
Would this make other operations faster as well?

jasonlyu123:
For the readFile part, Probably not. The problem with update import is that TypeScript called `languageService.getProgram` whenever they tried to sourcemap declaration map. If we update the project version when reading map files. The `getProgram` call would invalidate the old program and create another one. And it would continuously add new files and then recompile. 

I also change some parts of the module loader. That change might improve performance. It's while I am trying to make TypeScript read packages.json fewer times. But I still add a cache for package.json content because there's also completion and `@ types` modules check, so TypeScript still reads `package.json` very frequently.

hanszoons:
To add to this: the 'updating imports' is not just 'slow', it just hangs indefinitely when there are more than a few files involved. This makes refactoring a code base virtually impossible without constantly reloading vscode. I hope this fix can be merged soon.

PR #2079, opened 2023-06-29
fix: inlay hint position sometimes off by one

BlueGreenMagick:
Continued from PR https://github.com/sveltejs/language-tools/pull/2070
Fixes issue https://github.com/sveltejs/language-tools/issues/1988

Inlay hint position is sometimes off by 1 as mentioned in https://github.com/sveltejs/language-tools/pull/2070

This code detects if inlay hint position is off by 1 by checking `source[position]`. It is off by one if `kind == InlayHintKind.Type`, and `source[position]` is not a valid character that comes after an identifier.

A valid character that can come after an identifier, is ascii characters that is not alphanumeric A-Z(x41-x5A), a-z(x61-x7A), 0-9(x30-x39), or special characters '$'(x24), '_'(x5F), '\\'(x5C).


BlueGreenMagick:
Fixed lint

dummdidumm (approved):
Thank you!

PR #2100, opened 2023-07-17
feat: support prettier v3

dummdidumm:
Prettier v3 changed some APIs to be asynchronous. This is an easy fix, we just add `await` to some methods which is useless but harmless for v2 invocations.

The more involved problem is that the v3 Prettier API is incompatible with v2, and likewise you can't use prettier plugins that support v2 with Prettier v3 and vice versa. This means we need to add additional code to keep things working when someone has Prettier (or the plugin) installed but not the other (the plugin or Prettier). The current code does ignore the user's Prettier version if it's v3 but `prettier-plugin-svelte` is not found - it falls back to v2 of both Prettier and the plugin. It falls back to v2 for now because based on installs Prettier v2 is still far more popular so that's the safer choice.
There's one problem though: What if someone has other plugins in their node modules but _not_ the prettier plugin? In this case formatting could be done with Prettier v2 using plugins (from the user) that only support v3, and that fails. I don't see a good way around this, but I would also hope this does not happen often.

Thoughts @jasonlyu123 ?

### DIFF
--- a/packages/language-server/src/plugins/svelte/SveltePlugin.ts
+++ b/packages/language-server/src/plugins/svelte/SveltePlugin.ts
@@ -75,7 +75,7 @@ export class SveltePlugin
         }
 
         const filePath = document.getFilePath()!;
-        const prettier = importPrettier(filePath);
+        const prettier = await importFittingPrettier();
         // Try resolving the config through prettier and fall back to possible editor config
         const config = this.configManager.getMergedPrettierConfig(
             await prettier.resolveConfig(filePath, { editorconfig: true }),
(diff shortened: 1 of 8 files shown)


ghostdevv:
Perhaps we could have a "best guess" at prettier version and have a way to override it if it guesses wrong?

jasonlyu123:
I also can't think of ways around it. Probably easier to just show a message alert. But the problem is, can we reliably identify if the problem is caused by a plugin incompatible with a prettier version? Or we can only provide a troubleshooting hint.

dummdidumm:
tested:
- prettier v3 + plugin v3 -> works
- prettier v2 + plugin v2 -> works
- prettier v2 + no plugin -> works
- prettier v3 + no plugin -> works

Should be safe. I needed to adjust the code some more because Prettier v3 has removed the plugin search feature and so we need to load the config to see if `plugins: ['prettier-plugin-svelte']` is in there. Another review would be appreciated.

(will look into if we can test this in the test suite)

packages/language-server/src/plugins/svelte/SveltePlugin.ts
```diff
@@ -131,15 +172,22 @@ export class SveltePlugin
...
-                .getSupportInfo()
-                .languages.some((l) => l.name === 'svelte');
-            return hasPluginLoadedAlready ? [] : [require.resolve('prettier-plugin-svelte')];
+            return !isFallback && (await hasSveltePluginLoaded(prettier, plugins))
+                ? []
+                : [require.resolve('prettier-plugin-svelte')];
+        }
+
+        async function hasSveltePluginLoaded(p: typeof prettier, plugins: string[] = []) {
+            if (plugins.includes('prettier-plugin-svelte')) return true;
```
jasonlyu123, line 186: It's possible that plugins have a resolved full path, but probably an edge case. When we had the problem with plugins not loading in monorepo, someone suggested in the issue using `.prettierrc.cjs` with 
```js
module.exports = {
    plugins: [require.resolve('prettier-plugin-svelte')],
}
```

Issue #2107, opened 2023-07-22
Generics attribute with const modifier on script tag do not type check props


Nick-Mazuk:
### Describe the bug

When you have a generics attribute on a script tag (introduced in https://github.com/sveltejs/language-tools/pull/2020), it turns off type checking anytime that generic is used as the type of a prop.

### Reproduction

Let's say you have a component `items.svelte`

```svelte
<script lang="ts" generics="const T extends readonly string[]">
    export let items: T
    export let currentItem: T[number]
</script>
```

You can now use this component with any nonsense props. For instance, all of the following pass type checking when none of them should:

```svelte
<script lang='ts'>
    import Items from './items.svelte'
</script>

<!-- here currentItem is invalid -->
<Items items="{['item 1', 'item 2']}" currentItem="foobar" />
<Items items="{['item 1', 'item 2']}" currentItem="{3}" />
<Items items="{['item 1', 'item 2']}" currentItem="{{ foo: 'bar' }}" />

<!-- here items is invalid -->
<Items items="{['item 1', 'item 2', { foo: 'bar' }]}" currentItem="item 1" />
<Items items="{{ foo: 'bar' }}" currentItem="item 1" />
<Items items="foo bar" currentItem="item 1" />
```

### Expected behaviour

None of the above reproduction cases should type check.

However the following case should type check:

```svelte
<Items items="{['item 1', 'item 2']}" currentItem="item 2" />
```

[...]

PR #2118, opened 2023-08-05
fix: generic const modifier

jasonlyu123:
#2107

Switching to let TypeScript parse the generic attribute by wrapping it with an arrow function.


dummdidumm (approved):
Smart, just let TS do the work 😄

Issue #2106, opened 2023-07-22
"plugin.includes is not a function" error when formatting document


NateXVI:
### Describe the bug

After updating to Svelte for VS Code version 107.9.0, code formatting is no longer working.

### Reproduction

1. Clone the repo
```bash
git clone https://github.com/NateXVI/plugin-includes-error.git
cd plugin-includes-error
```

2. Install packages
```bash
npm i
```

3. Open VS Code
```bash
code .
```
4. Install Svelte for VS Code extension version 107.9.0

5. Mess up the formatting on `src/routes/+page.svelte`

6. Format `src/routes/+page.svelte` 
<kbd>Shift</kbd> + <kbd>Alt</kbd> + <kbd>F</kbd>

### Expected behaviour

`src/routes/+page.svelte` should be formatted

### System Info

- OS: Windows 11
- IDE: VS Code

### Which package is the issue about?

svelte-language-server

### Additional Information, eg. Screenshots

The error seems to be cause by changes made in #2100 and only happens when using a `prettier.config.cjs` instead of `.prettierrc` config file.

Svelte extension output I get when formatting
```
TypeError: plugin.includes is not a function
    at c:\Users\nated\.vscode\extensions\svelte.svelte-vscode-107.9.0\node_modules\svelte-language-server\dist\src\plugins\svelte\SveltePlugin.js:126:49
    at Array.some (<anonymous>)
    at hasSveltePluginLoaded (c:\Users\nated\.vscode\extensions\svelte.svelte-vscode-107.9.0\node_modules\svelte-language-server\dist\src\plugins\svelte\SveltePlugin.js:126:25)
    at importFittingPrettier (c:\Users\nated\.vscode\extensions\svelte.svelte-vscode-107.9.0\node_modules\svelte-language-server\dist\src\plugins\svelte\SveltePlugin.js:62:40)
    at async SveltePlugin.formatDocument (c:\Users\nated\.vscode\extensions\svelte.svelte-vscode-107.9.0\node_modules\svelte-language-server\dist\src\plugins\svelte\SveltePlugin.js:73:50)
    at async PluginHost.tryExecutePlugin (c:\Users\nated\.vscode\extensions\svelte.svelte-vscode-107.9.0\node_modules\svelte-language-server\dist\src\plugins\PluginHost.js:342:20)
    at async Promise.all (index 0)
[...]

Issue #1763, opened 2022-12-07
Show an info/error card in VS Code whenever required modules are not installed 


JL102:
### Describe the bug

I was quite confused as to why "Format Document" in Svelte files seemed to do nothing. Turns out that I didn't have `prettier-plugin-svelte` installed, and when the Svelte extension attempted to load it, it simply logged an error into the Output tab without notifying me through the UI. I believe the extension should show an error to the user instead of just logging the error.

### Reproduction

1. Set up the Svelte language extension, but EITHER:
  a. Do *not* install `prettier-plugin-svelte` in your project OR
  b. Install `prettier-plugin-svelte` and `prettier` in a subfolder, i.e. not the root of your Code instance (related to https://github.com/sveltejs/prettier-plugin-svelte/issues/311) 
2. Attempt to format a Svelte file via right click -> Format Document

### Expected behaviour

If any required modules are not found, instead of _just_ logging into the output console, the Svelte extension should create an info card in the bottom-right corner of the screen, displaying the error, so the user can understand why it's doing nothing. 

### System Info

- OS: Reproducible on both Windows and Ubuntu
- IDE: VS Code

### Which package is the issue about?

None

### Additional Information, eg. Screenshots

![image](https://user-images.githubusercontent.com/11487870/206276375-821a694d-8ee2-472e-aa1f-ab3502547238.png)
In this screenshot, note that `ScoringApp-Serverless` is the workspace root, while `offline` is the Svelte project. The Svelte project has both `prettier` and `prettier-plugin-svelte` installed.

dummdidumm:
We pushed a fix which hopefully resolved the issue. I don't think a popup window is necessary or helpful since there's nothing you can do about it, therefore closing.

If this happens again in the future, we could think about an icon or sth in the bottom right (similar to how prettier does it when it fails to format a document)

JL102:
> We pushed a fix which hopefully resolved the issue. I don't think a popup window is necessary or helpful since there's nothing you can do about it, therefore closing.
> 
> If this happens again in the future, we could think about an icon or sth in the bottom right (similar to how prettier does it when it fails to format a document)

[...]

PR #2119, opened 2023-08-05
fix: handle more cases of plugin config

jasonlyu123:
#2106

Fix cases where `plugins: [require('prettier-plugin-svelte')]`  and where `plugins ['prettier-plugin-svelte']` but never installed. 

There're still some cases where it would fail, like `plugins: [require.resolve('prettier-plugin-svelte')]` or `plugins: [path.join(prcess.cwd(), 'node_modules/prettier-plugin-svelte']`, but the plugin isn't installed, likely when people copy a prettier config but never pay attention to what the config actually does. Wondering if we should just show a message alert like #1763 suggested. Likely only detect errors with "Cannot find module". Or should we try to detect these two cases and fallback to using bundled version?


dummdidumm (approved):
Thank you!
I think we should hold off from the error modal for now. Let's see if more cases come up where people see it doesn't work as expected. If there's sufficient evidence that there are still many cases which we can't detect, then yes we can do something like that - but maybe through an icon similar to prettier that turns red or sth.

Issue #2140, opened 2023-08-26
VSCode. Cann't collapse #region with multi-line comment and JSDoc comments


narroweyedasianyouknow:
### Describe the bug

Inside tag <style> or in CSS exists only one way to create comment and it's block comment or multi-line comment

So if I use multi-line comment to create some regions in CSS, 
I cant able to collapse this regions because VSCode doesn't recognize it like a region  
It's also doesn't work with JSDoc comment

### Reproduction

It's very simple to reproduce, you need to write CSS like this
```css
	.block {
		/* #region TEST */
		/* some styles.... */
		/* #endregion */
	}
```

Or you can create something like this in JS

```js
	/** #region 100% SAFE FUNCS */
	function delete_downloads() {/** ... */}
	function delete_cache() {/** ... */}
	function create_you_hacked_file() {/** ... */}
	function change_user_password() {/** ... */}
	function fill_computer_with_cat_image() {/** ... */}
	function change_user_name() {/** ... */}
	function hello_world() {
		/* #region CALLER FUNCS TO HACK USER  */
		delete_downloads();
		delete_cache();
		change_user_password();
		change_user_name();
		fill_computer_with_cat_image();
		create_you_hacked_file();
		/* #endregion CALLER FUNCS TO HACK USER */
		return 'Hello world!';
	}
	/** #endregion */
```

### Expected behaviour

You should be able to collapse #region when you used multi-line comment or JSDoc comment in VSCode

### System Info

- OS: Ubuntu 22.04.3 LTS
- IDE: VSCode

### Which package is the issue about?

Svelte for VS Code extension

### Additional Information, eg. Screenshots

![Screencast](https://github.com/sveltejs/language-tools/assets/38946653/cd8c606e-73d6-44e4-b97e-42657df9d658)

ksixen:
Looks like I found where is the problem was.

I replaced Regex inside `language-configuration.json` to like this
![image](https://github.com/sveltejs/language-tools/assets/38946653/b8ccf118-facc-47db-84af-7102214cb3ee)


And seems it's working

[Screencast.webm](https://github.com/sveltejs/language-tools/assets/38946653/c6b4640e-f899-41bc-83fb-09886673f5e1)

PR #2141, opened 2023-08-26
(Fix) VSCode. Cann't collapse #region with multi-line comment and JSDoc comments

narroweyedasianyouknow:
I have created #2140 issue and I found where is the problem was and fixed it


jasonlyu123:
Thanks! 👍

Issue #2132, opened 2023-08-20
svelte-check fails if importing TS in monorepo


jakzo:
### Describe the bug

In a monorepo setup importing a Typescript file from outside a package's tsconfig.json `rootDir` file Typescript will complain with the following when running `svelte-check`:

```
File '/svelte-ts-monorepo-issue/packages/dep/index.ts' is not under 'rootDir' '/svelte-ts-monorepo-issue/packages/app'. 'rootDir' is expected to contain all source files.
  The file is in the program because:
    Imported via "dep" from file '/svelte-ts-monorepo-issue/packages/app/index.ts' with packageId 'dep/index.ts@1.0.0'
    Root file specified for compilation
```

This happens even doing an absolute import of a monorepo dependency (eg. `import "dep"`) which is symlinked into node_modules. Running `tsc` does not complain in this case, only when running `svelte-check` does it complain.

### Reproduction

Clone and run commands from readme: https://github.com/jakzo/svelte-ts-monorepo-issue

### Expected behaviour

svelte-check shows the same Typescript errors as tsc and allows importing from symlinked dependencies in node_modules.

### System Info

- OS: MacOS
- IDE: VSCode

### Which package is the issue about?

svelte-check

jakzo:
I tried debugging this just then and I think the issue stems from here: https://github.com/sveltejs/language-tools/blob/master/packages/language-server/src/svelte-check.ts#L229

The Typescript compiler program when initially parsing the program processes the `import "dep"` statement and finds `dep/index.ts` by searching in `node_modules` and marks it as such which is why tsc doesn't complain because [`isSourceFileFromExternalLibrary()`](https://github.com/microsoft/TypeScript/blob/main/src/compiler/utilities.ts#L6375) returns `true` for `dep/index.ts`, meaning the file will not be emitted and does not need to be under `rootDir`.

However when [svelte-check gets the list of files in the program](https://github.com/sveltejs/language-tools/blob/master/packages/language-server/src/svelte-check.ts#L191) Typescript returns the non-symlinked path to `dep/index.ts` which is not in `node_modules` and then when [svelte-check gets its diagnostics](https://github.com/sveltejs/language-tools/blob/master/packages/language-server/src/svelte-check.ts#L232) using the non-`node_modules` path Typescript sees it as **not** an external library and complains about `rootDir`.

dummdidumm:
[...]

PR #2146, opened 2023-08-31
(fix) only allow client files and project files to be root

jasonlyu123:
#2132

This PR moved the tracking of "document is open by the client" to document so it can be copied over to ts snapshot and create a new `documentManager.openClientDocument`  to replace `openDoucment` in existing places. This will mark the file as client files in one method call, so we don't have to add `markAsOpenByClient` to all existing tests.  The reason that we need client files is that files might be opened by the client and not in the project files. One reason is that the user has an empty include in the tsconfig or has the default one. Also, we didn't watch Svelte files with File Watcher. So the new svelte file won't be in the project files. We probably should do it for something like `git checkout`, but that can be a separate PR.

~~Alternatively, we could also create a `getClientSnapshot` method to replace the current `LanguageServiceContainer.getSnaphsot` method in `getLSAndTSDoc` . This should mark any document that is passed in as a client file. But I haven't checked if this would cause the `dep/index.ts` in #2132 to be marked as a client file.~~

I have to use a TypeScript internal API and add a layer of synchronization because of a module resolution bug.  The way `getSnapshot` prevents the problem before the amount of root files have changed so that TypeScript will check more thoroughly.  The existing test is different from how it'll work in typical cases. It probably worked when it was introduced, but we later changed to not run `getSnapshot` in new ts/js files, so it's not actually preventing the error.


packages/language-server/src/plugins/typescript/LSAndTSDocResolver.ts
```diff
@@ -116,11 +119,31 @@ export class LSAndTSDocResolver {
...
+        const { tsDoc, lsContainer, userPreferences } = await this.getLSAndTSDocWorker(document);
+
+        return { tsDoc, lang: lsContainer.getService(), userPreferences };
+    }
+
+    /**
+     * Retrieves the LS for operations that don't need cross-files information.
+     * can save some time by not synchronizing languageService program
+     */
+    async getLsForSyntheticOperations(document: Document): Promise<{
```
jasonlyu123, line 131: I created this as a separate method because it gave it a more meaningful name, so the code intention is clearer. And, well, tsserver also does that😆,

packages/language-server/src/plugins/typescript/features/CompletionProvider.ts
```diff
@@ -103,9 +103,11 @@ export class CompletionsProviderImpl implements CompletionsProvider<CompletionEn
...
         }
 
-        const { lang, tsDoc, userPreferences } = await this.lsAndTsDocResolver.getLSAndTSDoc(
-            document
-        );
+        const {
+            lang: langForSyntheticOperations,
+            tsDoc,
+            userPreferences
+        } = await this.lsAndTsDocResolver.getLsForSyntheticOperations(document);
```
jasonlyu123, line 110: Reuse completion check and JSDoc template both don't need the language service program to be up to date.

packages/language-server/test/plugins/typescript/features/FindComponentReferencesProvider.test.ts
```diff
@@ -8,16 +8,16 @@ import { LSAndTSDocResolver } from '../../../../src/plugins/typescript/LSAndTSDo
 import { pathToUrl } from '../../../../src/utils';
 import { serviceWarmup } from '../test-utils';
 
-const testDir = path.join(__dirname, '..');
+const testDir = path.join(__dirname, '..', 'testfiles');
```
jasonlyu123, line 11: This warms up the wrong directory. I fixed it while debugging why the test is flaky again.

packages/language-server/test/plugins/typescript/features/DiagnosticsProvider.test.ts
```diff
@@ -43,7 +43,7 @@ describe('DiagnosticsProvider', function () {
         const newFilePath = normalizePath(path.join(testDir, 'doesntexistyet.js')) || '';
         writeFileSync(newFilePath, 'export default function foo() {}');
         assert.ok(existsSync(newFilePath));
-        await lsAndTsDocResolver.getSnapshot(newFilePath);
+        await lsAndTsDocResolver.invalidateModuleCache(newFilePath);
```
jasonlyu123, line 46: This test is the one that failed because `getSnapshot` no longer adds new root files, and it's the reason why I use the internal API `hasInvalidatedResolutions`.

packages/language-server/src/plugins/typescript/service.ts
```diff
@@ -636,6 +704,26 @@ async function createLanguageService(
...
         docContext.onProjectReloaded?.();
     }
+
+    function updateIfDirty() {
+        if (!dirty) {
+            return;
+        }
+
+        languageService.getProgram();
+        svelteModuleLoader.clearPendingInvalidations();
```
jasonlyu123, line 714: This wrapper layer is needed because we need to clear the invalidated files so they won't affect every program update check afterwards. And the hoops you have to jump through are probably why `hasInvalidatedResolutions` is a TypeScript internal API 😅.

packages/language-server/src/plugins/typescript/module-loader.ts
```diff
@@ -49,6 +50,7 @@ class ModuleResolutionCache {
         this.cache.forEach((val, key) => {
             if (val && this.getCanonicalFileName(val.resolvedFileName) === resolvedModuleName) {
                 this.cache.delete(key);
+                this.pendingInvalidations.add(key.split(':::').shift() || '');
```
dummdidumm, line 53: Probably makes sense to have `:::` as `const` at the top and use that at this point

packages/language-server/src/plugins/typescript/module-loader.ts
```diff
@@ -61,16 +63,27 @@ class ModuleResolutionCache {
         const fileNameWithoutEnding =
             getLastPartOfPath(this.getCanonicalFileName(path)).split('.').shift() || '';
         this.cache.forEach((val, key) => {
+            const split = key.split(':::');
```
dummdidumm, line 66: `moduleName` and `containingFile` retrieval can probably be combined, right now they both split.

dummdidumm (commented):
Damn, impressive work!

To summarize in my own words, to double-check I understood this all correctly:
- we're shifting `openedByClient` elsewhere so that we no longer mark all TS/JS files we pick up along the way as open automatically, because that's wrong
- we're now using `openedByClient` to filter the files, which is why we need to be more granular about what is marked as opened in the client
- because we're no longer marking every TS/JS file as opened, we have to do some more work in other places to correctly clear the module resolution cache in situations like creation/deletion/move, which is what we need the failed lookup info etc for
- because we have this additional logic now, we can also be more precise about when to invalidate the program (`updateIfDirty`)
- as a drive-by, we now optimize a bit with the "get language service but don't sync" bit

My guess is this PR might make things a bit more performant in some cases?

jasonlyu123:
> We have to do some more work in other places to correctly clear the module resolution cache in situations like creation/deletion/move
Yeah. But it's also another bug because the previous method doesn't work for ts/js file creation, probably because we changed to not load new files unless it's already loaded. 

>as a drive-by, we now optimize a bit with the "get language service but don't sync" bit 
Nope. this is added because `updateIfDirty` might called at the wrong time and might have prevented previous optimizations from working. The reason why we need to manually trigger updates is because TypeScript will automatically check it in most language features. And that internal API is called when TS is checking it they need to rebuild. If we manually trigger an update we can then clear the invalidation as done.

Issue #2053, opened 2023-06-12
VSCode shows error/`any` for generated types


probablykasper:
### Describe the bug

SvelteKit generated types often don't work correctly with VSCode. It could be a bigger issue than just with `Locals`, I'm not sure.

In my production repo I have a `+layout.ts` which receives
In the reproduction repo I have a `+layout.ts` which receives data from `+layout.server.ts`.

If I open `+layout.ts` in VSCode, it shows the `data` type as `any`:
```
{
    session: any;
}
```

Then if I open the `+layout.server.ts`, it shows the reason for that:
<img width="754" alt="image" src="https://github.com/sveltejs/language-tools/assets/11315492/0e83260a-cae3-42e8-b300-c23041054b91">

If I just update `+layout.server.ts`, the error goes away and the types are correct in both files.

### Reproduction

Repo: https://github.com/probablykasper/sveltekit-generated-any

https://github.com/sveltejs/language-tools/assets/11315492/5f15b805-5b94-4796-9889-a9a0a27dbff6

### Expected behaviour

The type error shouldn't happen. If some error like this does happen, it would be nice to avoid silently using `any` and instead provide some hint that there's an error somewhere

### System Info

- OS: [e.g. Windows]
- IDE: [e.g. VSCode, Atom]

### Which package is the issue about?

Svelte for VS Code extension

### Additional Information, eg. Screenshots

Tested with a fresh VSCode instance, with only the Svelte for VSCode extension

mvartuc:
Any update on this? I'm having the same issue, haven't found a workaround so far.

MystPi:
I consistently have this problem every time I open a file that uses auto-generated types. While the issue goes away after editing the file, it would be better if this didn't happen at all.

einarpersson:
I wrote about this here https://github.com/sveltejs/kit/discussions/10703

I am having this problem on a daily basis to such an extent that it is unusable. My impression is that it has increased in frequency (or maybe it is just my app that has grown, which amplifies the problem).

Any update / work on this?

ESchouten:
Any update on this? The bug makes the development of my app very hard since my +page.svelte files don't know what data is in the data export

PR #2160, opened 2023-09-17
fix: ensure SvelteKit file is patched when first open

jasonlyu123:
#2053

If a file is patched the first time, typescript might have cached the unpatched version. It won't update even if we return the patched version in `getScriptSnapshot`. This should only happen to files the client opens after the first compilation of the proxy language service. In the first compilation, there won't be any files in the cache. And once there are any updates TypeScript will reparse the file with the patched snapshot.


dummdidumm (approved):
Nice!

michal-weglarz:
Thank you for fixing this 🙏

PR #2162, opened 2023-09-21
feat: ts5.2 clickable inlay parameter hints

jasonlyu123:
This also fixed some existing configs.  The `suppressWhenArgumentMatchesName` config is the opposite of what it should be. `propertyDeclarationTypes` is also not passed in. I also forgot to add the support to disable the type hint when it matches the name.


dummdidumm (approved):
nice!

Issue #1704, opened 2022-10-25
folding range support


BigBoulard:
Hi guys,

In a Svelte file, when params of a function are not on the same line, I can't fold the whole function. I can instead only fold the function params. 

```javascript
async function lambda(
    param1,
    param2,
    param3
  ) {
	const a = 1;
	const b = 2;
}
```

becomes

```javascript
async function lambda(
  ) {
  const a = 1;
  const b = 2;
}
```
instead of

```javascript
async function lambda() {}
```
Unlike `.svelte` files, it works perfectly on `.js` ones.

jasonlyu123:
We'll need to implement the folding range feature for this. The current one is an indent-based solution provided by VSCode. 

Haven't tried it yet. But I am not sure if the result would be great given we have some limitations with our current approach for typescript features. We might not be able to calculate the folding range when there are syntax errors in the markup. Not sure if this would cause the folded range suddenly expand. If it did it is worse than not being able to fold a function.

Issue #1120, opened 2021-07-28
Folding the script tag hides its closing tag


juancpgo:
When editing **.svelte** files, the script closing tag `</script>` gets hidden by **VS Code** when I fold it. 
![Screen Shot 2021-07-28 at 12 42 23](https://user-images.githubusercontent.com/1007847/127355401-0cca15d4-5722-497c-b54f-1482476bbaed.png)

It doesn't happen with any other html tag, and it doesn't happen if I switch the file extension to **.html**.

If I rename `script` to anything else, the closing tag shows up:
![Screen Shot 2021-07-28 at 12 57 20](https://user-images.githubusercontent.com/1007847/127355844-a5b28225-7ec7-4262-b56c-d86daa93617a.png)

Does anyone know why? Is there a solution?

juancpgo:
ps: My VS Code settings folding strategy is set to "indentation". But switching it to "auto" doesn't solve the problem.

jasonlyu123:
We have [special treatment](https://github.com/sveltejs/language-tools/blob/20ade7cf5cdabc4c1fa05dcd158481db5012583a/packages/svelte-vscode/language-configuration.json#L30) with the script and style tag so that it can be folded without indentation. I don't know if it's possible to show the end tag.

juancpgo:
> We have [special treatment](https://github.com/sveltejs/language-tools/blob/20ade7cf5cdabc4c1fa05dcd158481db5012583a/packages/svelte-vscode/language-configuration.json#L30) with the script and style tag so that it can be folded without indentation. I don't know if it's possible to show the end tag.

So that must be it, because I realize those other tags (`style` and `template`) also behave the same way.

Can I remove that special treatment? How do I do it?
(I'd like them to behave as the others tags, I use indentation the same way)

dummdidumm:
Unfortunately it's not possible to remove that special treatment (well, you could go into the folder where the extension is installed, find the JSON file and change it there, but that's not a viable workaround I guess). We are stuck in an either-or-situation here. If we remove the special handling for these three tags, #523 is broken again, which would be worse than living with the minor inconvenience that the folded tags don't show their end tag.

quoid:
> well, you could go into the folder where the extension is installed, find the JSON file and change it there, but that's not a viable workaround I guess

[...]

PR #2169, opened 2023-09-26
feat: folding range support

jasonlyu123:
#1704 #1120

This adds the syntactic folding range support instead of the VSCode's default indentation-based and regex-based folding. For embedded languages like Pug and Sass, I added a simplified version of indentation folding. The indentation folding is also a fallback for svelte blocks if there is a parser error. 

For users who prefer the speed of the indent-based folding, You can still roll back with the following config.

```json
{
    "[svelte]": {
        "editor.foldingStrategy": "indentation"
    }
}
```

### Implementation

The commits might be messy because I started this nearly a year ago and have done part of this every few months 😆. The process is "the TypeScript result is good enough that's used it for HTML and svelte block as well" to "the TypeScrip result has a lot of weird cases, that's use HTML language service and svelte ast instead". 

Now, the TypeScript folding handles everything inside the script tags and any folding ranges that are one-to-one in the generated code, i.e. it's untouched or only being moved around. Component tag and HTML tag are handled by HTML language service.  Svelte blocks like each, if and await are extracted from Svelte AST. And folded from the end of the start tag to the start of the end tag. For example,

```svelte
{#if hi}<!-- start here-->
  {hi}
<!-- end here-->{/if}
```

VSCode only supports line folding, so it'll start at line 0 and end at line 1, leaving line 2 visible after fold. 

```svelte
{#if hi}...
{/if}
```


dummdidumm (approved):
Damn, impressive work!

General thought: We're using various `getText()` operations like `indexOf`, `split(/\n\r?/)` etc in many places - maybe it's worth having some of these operations move into the document in a more performant manner (for example, keeping the splitted lines cached etc).

packages/language-server/src/plugins/css/CSSPlugin.ts
```diff
@@ -453,6 +518,18 @@ function shouldExcludeColor(document: CSSDocument) {
     }
 }
 
+function shouldUseIndentBasedFolding(kind?: string) {
+    switch (kind) {
+        case 'postcss':
+        case 'sass':
+        case 'stylus':
+        case 'styl':
```
Xeronic, line 526: style?

Issue #2168, opened 2023-09-25
VSCode suggestions are too broad with Svelte extension


bitkidd:
### Describe the bug

For some reason Svelte to VSCode extension does not provide optimal TS experience in terms of suggestions, sometimes I get too many of them, sometimes none. I narrowed down the issue to round braces, so if some value is inside of them, suggestions do not work, if not, work okay.

### Reproduction

Here suggestions do not work for me.

```svelte
<script lang="ts">
  const someFunction = (input: string) => input

  export let data: { title: string; excerpt?: string } | null = null
</script>

{#if data}
  <h1>{someFunction(data.title)}</h1> <!-- Here suggestions after . do not work -->
  <p>{data.excerpt}</p>
{/if}
```

### Expected behaviour

Suggestions that are strictly tied to the data type provided in script tag.

### System Info

- OS: MacOS 13.6
- IDE: VSCode 1.82.2

### Which package is the issue about?

Svelte for VS Code extension

### Additional Information, eg. Screenshots

![Screenshot 2023-09-25 at 15 05 59](https://github.com/sveltejs/language-tools/assets/2136460/cfadb3bd-5789-43cf-bd99-595ccd6331d5)

jasonlyu123:
https://github.com/sveltejs/language-tools/issues/1302#issuecomment-1002358234

dummdidumm:
Closing as duplicate of #1302

PR #2173, opened 2023-10-02
fix: enhance precompile blank for parser error workaround

jasonlyu123:
For the specific case of #2168. The blanking pre-process currently will stop whenever it encounters a char that is not something we want to blank. But it's probably safe to skip `]` and `)` and continue checking operators to blank. Haven't thought of cases where it might be not safe


dummdidumm (approved):
Also can't think of a case where it breaks - let's try this 👍

PR #2184, opened 2023-10-17
feat: add a default types in the $$Generic

fedorovvvv:
Hi!✨
A common problem with $$Generic is that there was no way to set a default type

I tried to add more functionality and now it is possible to set the default type as the second argument to $$Generic

```ts
type T = $$Generic<number, 2>
interface $$Props {
    value?:T
    result?: T extends 2 ? 'AAAAAA' : 'OOOOO'
}
export let value:$$Props['value'] = 2
```

I think this will take types in svelte one step further and open up a lot of new possibilities
I can't live without default types))🥲

### DIFF
--- a/packages/svelte2tsx/src/svelte2tsx/nodes/Generics.ts
+++ b/packages/svelte2tsx/src/svelte2tsx/nodes/Generics.ts
@@ -51,13 +51,22 @@ export class Generics {
                     'Invalid $$Generic declaration: $$Generic definitions are not allowed when the generics attribute is present on the script tag'
                 );
             }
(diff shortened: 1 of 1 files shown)


fedorovvvv:
I fixed the condition, error output and ran all the tests. Everything's fine.

I hope you will accept this PR or we will improve it together💛

fedorovvvv:
No one has used this syntax yet because of the error condition, so the chance of affecting projects is minimal if not 0

fedorovvvv:
Whoa!
I wasted an evening digging up $$$Generic😂

Is `generics` written about in the documentation?

Thanks💜

PR #2185, opened 2023-10-19
feat: add option to force-start svelte plugin

dummdidumm:
This adds the assumeIsSvelteProject project for the typescript plugin. Some editors like Intellij may use their own detection mechanisms for whether or not a plugin should run, and this allows them to do that and skip our detection mechanism.

@tomblachut does this solve your use case?

### DIFF
--- a/packages/typescript-plugin/src/config-manager.ts
+++ b/packages/typescript-plugin/src/config-manager.ts
@@ -4,12 +4,15 @@ const configurationEventName = 'configuration-changed';
 
 export interface Configuration {
     enable: boolean;
+    /** Skip the Svelte detection and assume this is a Svelte project */
+    assumeIsSvelteProject: boolean;
 }
 
 export class ConfigManager {
     private emitter = new EventEmitter();
     private config: Configuration = {
-        enable: true
+        enable: true,
+        assumeIsSvelteProject: false
     };
 
     onConfigurationChanged(listener: (config: Configuration) => void) {
--- a/packages/typescript-plugin/README.md
+++ b/packages/typescript-plugin/README.md
@@ -24,7 +24,10 @@ Then add it to your `tsconfig.json` or `jsconfig.json`:
     "compilerOptions": {
         ...
         "plugins": [{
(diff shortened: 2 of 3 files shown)


packages/typescript-plugin/src/config-manager.ts
```diff
@@ -4,12 +4,15 @@ const configurationEventName = 'configuration-changed';
...
+    /** Skip the Svelte detection and assume this is a Svelte project */
+    assumeIsSvelteProject: boolean;
 }
 
 export class ConfigManager {
     private emitter = new EventEmitter();
     private config: Configuration = {
-        enable: true
+        enable: true,
+        assumeIsSvelteProject: false
```
tomblachut, line 15: @dummdidumm looks great, thanks!

Just to double check, `isConfigChanged` in this class still only reacts to `enable`, is that  alright?

dummdidumm: Yes that's right. I implemented this way because `assumeIsSvelteProject` is the very first thing that is checked, and everything else is just skipped else, so no point in reacting to something that can never be reacted to you.

Is that a problem for the Intellij integration because you create the TypeScript service eagerly, and therefore all plugins eagerly, and what you need instead is some kind of "sleep" mode where things are already prepared but not turned on yet in the plugin? I imagine that being tricky, I don't know how TypeScript would react if we lazily patch stuff. Probably the closest to this is the `enable` flag which turns off all the intellisense-related work when it's `false`, so it's almost like the sleep mode I described.

tomblachut: On second take your current code is fine. No, we create TypeScript service when the first TS file is opened and we control which commands are sent in what order. We never had to use "configurePlugin" before actually and my initial understanding was wrong, so I was getting undefined at creation time. After a bit of investigation and a bug fix in protocol handling, now it works as expected 👍

Type checks for Svelte compiled classes work without node_modules installed. Zero-effort types are not injected but after `npm install` everything refreshes properly.

Thanks again!

PR #2194, opened 2023-11-06
chore: remove use of soon to be deprecated method

benmccann:
This has been removed from Svelte 5


dummdidumm:
Doing this as part of a larger Svelte 5 compat PR, closing in favor of that

PR #2200, opened 2023-11-12
test: increase test timeout for riscv64

phanen:
Some tests fails on slow RISC-V architecture.
Fix by increasing the timeout.

### DIFF
--- a/packages/language-server/package.json
+++ b/packages/language-server/package.json
@@ -5,7 +5,7 @@
     "main": "dist/src/index.js",
     "typings": "dist/src/index",
     "scripts": {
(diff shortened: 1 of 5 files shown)


jasonlyu123:
We discussed this before and want to avoid increasing the timeout to catch potential performance problems. Which test failed because of time out in your environment? We'd prefer to separate out the demanding part into a warm-up step instead of picking a random number on all tests. If it's not possible, I'd prefer you add a utility function that takes an environmental variable or a command line flag so you can increase it by yourself.

phanen:
Thanks for advices. I can use command flags `pnpm test -- --timeout 100000` to pass all tests with default timeout. But most tests with override hardcode may fail (like serviceWarmup and performance test..).

I've opened a new pr (https://github.com/sveltejs/language-tools/pull/2205) to make it possible to handle all tests using cli only.

Issue #2179, opened 2023-10-14
VSCode extension/intellisense slow in .svelte files with a lot of HTML


ChrolloDeveloper:
### Describe the bug

VSCode's intellisense is slow when writing in .svelte files with significant HTML portions (>2500 lines). Waiting for it to suggest variables/properties can take many seconds, and pressing Ctrl+Space will show a "Loading..." indicator as the extension/VSCode attempts to report the available values.

However, commenting out the HTML portion of the .svelte files makes intellisense almost instant again. Even if that HTML code is replaced with equivalent amounts of JS code, intellisense continues to be fine, so it seems to be an issue with how the extension handles the HTML code.

In addition, the slowness happens even when the HTML code is completely basic - just a simple repeating div + span + input for 2500 lines with no svelte-specific bindings (on:click, bind:value, etc) to <script> variables. So maybe the issue is with how the HTML is being parsed for values by the extension rather than the actual bindings? Who knows.

### Reproduction

I was able to recreate this just by putting together some random functions/variables and some HTML that bind to them. 
Note that it's attached here as "testComponent.txt" since GitHub doesn't allow .svelte extensions, so it should be renamed back. In the file, I generated a bunch of random functions and values, and then appended a bunch of basic HTML underneath. There's a section of commented out functions `extraF0 ... extraF999` which I use during the reproduction steps to demonstrate that extra JS code is handled without problems, it's just HTML code that slows down the IDE's intellisense.

1. Open the attached file in VSCode with the Svelte for VS Code extension enabled
2. Anywhere in the script tag, type something to see intellisense completion, e.g. `val0.`
3. Observe that it takes multiple seconds for the suggestion list to appear, including the one for "a: (property) number".
4. Comment out the HTML code and retry step 2
5. Observe that intellisense now shows suggestions instantly
6. Un-comment the lines of code with "extraFn" definitions (twice as long as the commented-out HTML).
7. Repeat step 2
8. Observe that intellisense continues to show suggestions immediately, despite there being even more code to parse through.
[testComponent.txt](https://github.com/sveltejs/language-tools/files/12905348/testComponent.txt)

### Expected behaviour

Intellisense should suggest relevant variables/properties quickly.

### System Info

- OS:macOS 13.5.2 - Apple Silicon (M1 Pro)-  
- IDE: VSCode Version: 1.83.1 with svelte extension enabled

### Which package is the issue about?

Svelte for VS Code extension

PR #2216, opened 2023-11-24
perf: cancel diagnostics

jasonlyu123:
Another part of #2179 is that TypeScript took a long time to type-check the file. When the file is ts or check-js, it'll be `getSemanticDiagnostics`, and for non-checked js, it's `getSuggestionDiagnostics`. There is probably not much we can do about it. But we can instead cancel the check when the file is updated. The cancellation can only be checked if there is an async operation. I added a small delay in the typescript check so that the update notification queue might run in that time frame. And we could return early for the svelte and typescript diagnostic. CSS and HTML are pure sync code and generally not that heavy, so it's probably not worth the overhead.

### DIFF
--- a/packages/language-server/src/plugins/typescript/features/DiagnosticsProvider.ts
+++ b/packages/language-server/src/plugins/typescript/features/DiagnosticsProvider.ts
@@ -80,11 +80,23 @@ export class DiagnosticsProviderImpl implements DiagnosticsProvider {
             ];
         }
 
-        let diagnostics: ts.Diagnostic[] = [
-            ...lang.getSyntacticDiagnostics(tsDoc.filePath),
-            ...lang.getSuggestionDiagnostics(tsDoc.filePath),
-            ...lang.getSemanticDiagnostics(tsDoc.filePath)
-        ];
+        let diagnostics: ts.Diagnostic[] = lang.getSyntacticDiagnostics(tsDoc.filePath);
+        const checkers = [lang.getSuggestionDiagnostics, lang.getSemanticDiagnostics];
+
+        for (const checker of checkers) {
+            if (!cancellationToken) {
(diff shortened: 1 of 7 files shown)


packages/language-server/src/plugins/typescript/features/DiagnosticsProvider.ts
```diff
@@ -80,11 +80,23 @@ export class DiagnosticsProviderImpl implements DiagnosticsProvider {
...
+                diagnostics.push(...checker.call(lang, tsDoc.filePath));
+                continue;
+            }
+
+            // wait a bit so the event loop can check for cancellation
+            // or let completion go first
+            await new Promise((resolve) => setTimeout(resolve, 10));
+            if (cancellationToken.isCancellationRequested) {
+                return [];
+            }
```
dummdidumm, line 97:
```suggestion
            if (cancellationToken) {
                // wait a bit so the event loop can check for cancellation
                // or let completion go first
                await new Promise((resolve) => setTimeout(resolve, 10));
                if (cancellationToken.isCancellationRequested) {
                    return [];
                }
            }
```

dummdidumm (commented):
We have cancellation tokens elsewhere with abort logic. What's different to this one, or rather, why can't we reuse the logic from the other cancellation?

jasonlyu123:
It's the same since vscode-language-server uses this as well. But because we manage the diagnostic, we need to manage it and cancel ourselves. And since diagnostics have already been debounced, adding two small delays might not make a huge difference anyway. 

LSP 3.17 has a diagnostics pull model where the client manages the diagnostics. With that, it would be similar to what we have in other features. We can also try switching to the pull model in supported clients. But it probably still makes sense to add it to the existing push model so unsupported clients can benefit as well.

Issue #2222, opened 2023-12-04
Typescript Language Server Error on "Move to File" in Svelte Project


ian-viri:
### Describe the bug

The new VSCode `Move to file` refactor throws an error in a Svelte project that has `*.svelte` component files.

### Reproduction

1. Create a new SvelteKit project
2. Add a new component file, e.g. `src/lib/Component.svelte`
3. Use that component in a `+page.svelte file`
4.  Open any `.ts` file, find a function, and attempt to `Refactor > Move to File`
5. See the following error from the TS lanugage server
```
<semantic> TypeScript Server Error (5.2.2) Debug Failure. File /Users/iangillis/code/blog/prisma-sveltekit-seeding/src/lib/Component.svelte has unknown extension. Error: Debug Failure. File /Users/iangillis/code/blog/prisma-sveltekit-seeding/src/lib/Component.svelte has unknown extension. at extensionFromPath (/Users/iangillis/code/blog/prisma-sveltekit-seeding/node_modules/typescript/lib/tsserver.js:21064:39) at /Users/iangillis/code/blog/prisma-sveltekit-seeding/node_modules/typescript/lib/tsserver.js:144823:221 at mapDefined (/Users/iangillis/code/blog/prisma-sveltekit-seeding/node_modules/typescript/lib/tsserver.js:2671:22) at Proxy.getMoveToRefactoringFileSuggestions (/Users/iangillis/code/blog/prisma-sveltekit-seeding/node_modules/typescript/lib/tsserver.js:144823:19) at IpcIOSession.getMoveToRefactoringFileSuggestions (/Users/iangillis/code/blog/prisma-sveltekit-seeding/node_modules/typescript/lib/tsserver.js:184487:41) at getMoveToRefactoringFileSuggesti...
```

### Expected behaviour

Ideally, be able to use the `Move to File` refactor to move files between any combination of `.ts` and `.svelte` files. At the very least, be able to us it to move files between two `.ts` files without error.

### System Info

- OS: Mac
- IDE: VSCode

### Which package is the issue about?

Svelte for VS Code extension

### Additional Information, eg. Screenshots

Moved from https://github.com/microsoft/vscode/issues/199785

jasonlyu123:
[...]

PR #2227, opened 2023-12-10
fix: workaround for move to file refactor restriction

jasonlyu123:
#2222
 
When using the move to existing file refactor, TypeScript currently throws a `Debug.assert` error when any file in the program is not a supported file extension. Working around it by monkey patching the `program.getSourceFiles` method during the `getMoveToRefactoringFileSuggestions` function call. Let's hope we can convince the TypeScript team to remove this restriction, and we'll no longer need this workaround.


jasonlyu123:
Find one problem. What if other ts-plugins were returning a new language service and didn't support `getProjectVersion`? That way, the program object might be updated so that the workaround won't work, and it might also patch the new program with the old `getSourceFiles`. Might need to find a way to detect that.

dummdidumm (approved):

Issue #2228, opened 2023-12-10
Svelte VS Code extension keeps crashing (TypeError: Cannot read properties of undefined (reading 'sourceFile'))


marianmeres:
### Describe the bug

VSCode Svelte Extension keeps crashing on a npm workspaces based Sveltekit typescript project. On other non workspaces Svelte/kit projects it seems to be running fine.

I'm seeing this (or very similar) error output:
```
Node.js v18.15.0
[Info  - 12:16:16] Connection to server got closed. Server will restart.
true
Initialize language server at  file:///<...omitted...>
Initialize new ts service at  <...omitted...>/tsconfig.json
Trying to load configs for <...omitted...>
Loaded config at  <...omitted...>/svelte.config.js
SnapshotManager File Statistics:
Project files: 22
Svelte files: 9
From node_modules: 0
Total: 22
Reloading ts service at  <...omitted...>/tsconfig.json  due to config updated
Trying to load configs for <...omitted...>
SnapshotManager File Statistics:
Project files: 22
Svelte files: 9
From node_modules: 1
Total: 23
/Users/mm/.vscode/extensions/svelte.svelte-vscode-108.1.0/node_modules/typescript/lib/typescript.js:134135
    return !!entry.sourceFile;
                   ^

TypeError: Cannot read properties of undefined (reading 'sourceFile')
    at isDocumentRegistryEntry (/Users/mm/.vscode/extensions/svelte.svelte-vscode-108.1.0/node_modules/typescript/lib/typescript.js:134135:20)
    at getDocumentRegistryEntry (/Users/mm/.vscode/extensions/svelte.svelte-vscode-108.1.0/node_modules/typescript/lib/typescript.js:134211:21)
    at Object.releaseDocumentWithKey (/Users/mm/.vscode/extensions/svelte.svelte-vscode-108.1.0/node_modules/typescript/lib/typescript.js:134306:21)
    at /Users/mm/.vscode/extensions/svelte.svelte-vscode-108.1.0/node_modules/typescript/lib/typescript.js:142983:67
    at forEach (/Users/mm/.vscode/extensions/svelte.svelte-vscode-108.1.0/node_modules/typescript/lib/typescript.js:55:24)
    at cleanupSemanticCache (/Users/mm/.vscode/extensions/svelte.svelte-vscode-108.1.0/node_modules/typescript/lib/typescript.js:142983:9)
[...]

Issue #1965, opened 2023-03-26
issues with `"moduleResolution": "Node16"` inside `tsconfig.json`


ivanhofer:
### Describe the bug

When setting `moduleResolution` to `Node16` inside `tsconfig.json` I encounter two issues:

1. the "zero-effort typesafety" functionality stops working
2. if I run `svelte-check:watch` and safe the `tsconfig.json` file, the process crashes with following error message

	```
	Watching for file changes.../home/my-app/node_modules/.pnpm/typescript@4.9.5/node_modules/typescript/lib/typescript.js:139552
	        return !!entry.sourceFile;
	                       ^
	
	TypeError: Cannot read properties of undefined (reading 'sourceFile')
	    at isDocumentRegistryEntry (/home/my-app/node_modules/.pnpm/typescript@4.9.5/node_modules/typescript/lib/typescript.js:139552:24)
	    at getDocumentRegistryEntry (/home/my-app/node_modules/.pnpm/typescript@4.9.5/node_modules/typescript/lib/typescript.js:139612:25)
	    at Object.releaseDocumentWithKey (/home/my-app/node_modules/.pnpm/typescript@4.9.5/node_modules/typescript/lib/typescript.js:139718:25)
	    at /home/my-app/node_modules/.pnpm/typescript@4.9.5/node_modules/typescript/lib/typescript.js:169519:45
	    at Object.forEach (/home/my-app/node_modules/.pnpm/typescript@4.9.5/node_modules/typescript/lib/typescript.js:181:30)
	    at Object.dispose (/home/my-app/node_modules/.pnpm/typescript@4.9.5/node_modules/typescript/lib/typescript.js:169518:20)
	    at dispose (/home/my-app/node_modules/.pnpm/svelte-check@3.1.4_svelte@3.57.0/node_modules/svelte-check/dist/src/index.js:90308:25)
	    at watchConfigCallback (/home/my-app/node_modules/.pnpm/svelte-check@3.1.4_svelte@3.57.0/node_modules/svelte-check/dist/src/index.js:90343:9)
	    at /home/my-app/node_modules/.pnpm/typescript@4.9.5/node_modules/typescript/lib/typescript.js:6724:17
	    at /home/my-app/node_modules/.pnpm/typescript@4.9.5/node_modules/typescript/lib/typescript.js:6439:180
	```

### Reproduction

demo repository: https://github.com/ivanhofer/svelte-check-moduleresolution-node16

[...]

PR #2230, opened 2023-12-13
fix: prevent crash in moduleResolution Node16+

jasonlyu123:
#1965  #2228

Patching the `releaseDocument` methods of  `DocumentRegistry` so it doesn't error. From my understanding, TypeScript doesn't use the `impliedNodeFormat` inside the `DocumentRegistry`. It'll always be treated as undefined for the svelte files. So, patching the `releaseDocument` part should be enough.

While we were here, I also changed to reuse the `documentRegistry` across language services. It should reduce some memory usage for files reused across language services. While checking if we can optimize the restart as well, I found that TypesScript added the support for `scriptKind` changes three years ago. So we no longer need to restart it because of `scriptKind` changes.

### DIFF
--- a/packages/language-server/src/plugins/typescript/service.ts
+++ b/packages/language-server/src/plugins/typescript/service.ts
@@ -18,7 +18,8 @@ import {
     ensureRealSvelteFilePath,
     findTsConfigPath,
     getNearestWorkspaceUri,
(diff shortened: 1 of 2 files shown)


dummdidumm (approved):
Great findings, thank you!

Issue #2232, opened 2023-12-19
Poor auto-complete performance when Drizzle ORM is installed


cdebotton:
### Describe the bug

When installing Drizzle ORM and setting up a single table in a brand new SvelteKit project (no other libraries), the language server slows down to the point where it's almost unusable, however only within the svelte template syntax. Intellisense is still instantaneous in `.ts`, `.svelte.ts`, and `<script lang="ts">` blocks.

When setting up a similar example project in Next.js (only using ts-server in the IDE), there are no performance issues.

### Reproduction

```bash
$ pnpm create svelte@latest example
$ cd example
$ pnpm add drizzle-orm
$ mkdir -p src/lib/server && touch src/lib/server/db.ts
```

```ts
// db.ts
import { drizzle } from 'drizzle-orm/vercel-postgres';
import { sql } from '@vercel/postgres';
import { pgTable, serial, timestamp, varchar } from 'drizzle-orm/pg-core';

export const db = drizzle(sql);

export const objects = pgTable('objects', {
  id: serial('id').primaryKey(),
  name: varchar('name', { length: 255 }),
  createdAt: timestamp('created_at').defaultNow().notNull(),
  updatedAt: timestamp('updated_at'),
});
```

Editing an `*.svelte` files results in unusable autocomplete with HTML elements, script/style blocks.

Autocomplete for typescript in TS files and within script blocks performs as expected.

When I uninstall `drizzle-orm`, the performance issues disappear and autocomplete preforms as expected.

### Expected behaviour

There should ideally be no extra delay when working within the svelte template syntax

### System Info

- OS: MacOS 14.2
- IDE: VSCode

```
Version: 1.85.1 (Universal)
Commit: 0ee08df0cf4527e40edc9aa28f4b5bd38bbff2b2
Date: 2023-12-13T09:48:06.308Z (6 days ago)
Electron: 25.9.7
ElectronBuildId: 25551756
Chromium: 114.0.5735.289
Node.js: 18.15.0
V8: 11.4.183.29-electron.0
OS: Darwin arm64 23.2.0
```

### Which package is the issue about?

I believe it's an issue with the `language-server`?

### Additional Information, eg. Screenshots

Here's an example repo:
https://github.com/cdebotton/svelte-drizzle-example-language-tools

I've tested this on both an M3 Pro and an M1 Max and found the same results.

jasonlyu123:
[...]

Issue #2193, opened 2023-11-05
Intellisense for never before imported libraries doesnt work


EldinHb:
### Describe the bug

When you install a Svelte library and then try to use a Svelte component in any .svelte file the intellisense cant find the library. When you add `import {} from 'library';` in any other file intellisense is able to find any component of said library.

Screenshots with https://tabler.io/docs/icons/svelte as the library. Could also reproduce with https://www.skeleton.dev/.

No intellisense for any icon.
<img width="940" alt="image" src="https://github.com/sveltejs/language-tools/assets/8807015/ccc1d7f8-71fc-4f3c-a688-53f97d8f120a">

Intellisense works with `import {} from '@tabler/icons-svelte';` in same file.
<img width="832" alt="image" src="https://github.com/sveltejs/language-tools/assets/8807015/4a0eabfc-ba3e-431a-bfd2-a2207aa5360a">

Intellisense works with `import {} from '@tabler/icons-svelte';` in another unused .svelte file.
<img width="929" alt="image" src="https://github.com/sveltejs/language-tools/assets/8807015/0ae33475-fedd-40d1-a8ed-f8821999bb3e">

### Reproduction

Create new sveltekit project. 

Pick empty project with typescript and no other libraries.

Install https://tabler.io/docs/icons/svelte

See the screenshots in the above description.

### Expected behaviour

Intellisense to work with never before imported libraries.

### System Info

- OS: macOS
- IDE: vscode

### Which package is the issue about?

Svelte for VS Code extension

jasonlyu123:
This is expected because the type definition is only loaded if there is an import to it. Loading arbitrary d.ts files from node_modules might cause some performance problems. In TypeScript projects, the auto-import also behaves like this. 

But in the js projects, the automatic [type acquisition](https://www.typescriptlang.org/tsconfig#typeAcquisition) will also discover some of the direct dependencies in a child process. This is a feature that we didn't support, so I am marking this issue as a feature request instead of a bug.

EldinHb:
Do you mean in regular typescript projects that aren't svelte projects? I've never seen this in regular typescript projects or react projects. It works smoothly and I haven't seen performance issues in those projects.

jasonlyu123:
Yes. It's the same. I tried in a ts file in a plain typescript project with only a tsconfig and in a SvelteKit project. They both don't provide auto-import for newly installed packages. 

[...]

PR #2237, opened 2023-12-21
perf: auto import cache

jasonlyu123:
#2232 #2193

Implementing TypeScript's internal APIs for cache with the help of Volar's typescript-auto-import-cache. It also adopted the packag.json auto-import added in TS 4.0 for TSServer. The package.json auto-import has some limitations, though. It only works for dependencies and not dev dependencies. And it seems like the first dependencies also don't work until a server restarts. This also happens in ts/js files.

For implementation, the original text-based package.json cache was removed and now relies on typescript-auto-import-cache and ts.ModuleResolutionCache for the cache, and we only watch the files for invalidation.

The reason I didn't use the default API of typescript-auto-import-cache is that there is some memory leak with the implementation. One is that an old program object is stored, and the other is that there is no API to clear the project object stored in a map during the tsconfig update. The former is a bit hard to fix without changing the public API. But we definitely will need to try to resolve this when we eventually migrate to Volar. The `projectService` object also caused problems with the virtual file system test because it is a singleton with the `ts.sys` passed in earlier. It should not be a problem if we handle this by ourselves. 

Although migrating to Volar, which also uses `typescript-auto-import-cache`, also improves the auto import performance, it seems like we'll still need to put a lot of work into migrating existing post-processing and extra features for Volar. So, it probably still makes sense to add this now.

### DIFF
--- a/packages/language-server/src/plugins/typescript/features/CompletionProvider.ts
+++ b/packages/language-server/src/plugins/typescript/features/CompletionProvider.ts
@@ -236,7 +236,9 @@ export class CompletionsProviderImpl implements CompletionsProvider<CompletionEn
             offset,
             {
                 ...userPreferences,
-                triggerCharacter: validTriggerCharacter
+                triggerCharacter: validTriggerCharacter,
+                // Just in case, if there is a parser error. Force ts to not use the cache
+                triggerKind: tsDoc.parserError ? undefined : completionContext?.triggerKind
             },
             formatSettings
         );
--- a/packages/language-server/src/plugins/typescript/features/CompletionProvider.ts
+++ b/packages/language-server/src/plugins/typescript/features/CompletionProvider.ts
@@ -236,7 +236,7 @@ export class CompletionsProviderImpl implements CompletionsProvider<CompletionEn
             offset,
             {
                 ...userPreferences,
-                triggerCharacter: validTriggerCharacter,
+                triggerCharacter: validTriggerCharacter
             },
             formatSettings
         );
--- a/packages/language-server/src/plugins/typescript/features/CompletionProvider.ts
+++ b/packages/language-server/src/plugins/typescript/features/CompletionProvider.ts
@@ -237,8 +237,6 @@ export class CompletionsProviderImpl implements CompletionsProvider<CompletionEn
             {
                 ...userPreferences,
                 triggerCharacter: validTriggerCharacter,
-                // Just in case, if there is a parser error. Force ts to not use the cache
-                triggerKind: tsDoc.parserError ? undefined : completionContext?.triggerKind
             },
             formatSettings
         );
--- a/packages/language-server/src/plugins/typescript/service.ts
+++ b/packages/language-server/src/plugins/typescript/service.ts
@@ -719,8 +719,13 @@ async function createLanguageService(
 
         dirty = false;
 
+        if (!oldProgram) {
+            changedFilesForExportCache.clear();
+            return;
+        }
+
         for (const fileName of changedFilesForExportCache) {
-            const oldFile = oldProgram?.getSourceFile(fileName);
+            const oldFile = oldProgram.getSourceFile(fileName);
             const newFile = program?.getSourceFile(fileName);
 
             // file for another tsconfig
--- a/packages/language-server/src/plugins/typescript/service.ts
+++ b/packages/language-server/src/plugins/typescript/service.ts
@@ -21,6 +21,7 @@ import {
     hasTsExtensions,
     isSvelteFilePath
 } from './utils';
+import { createLanguageService as createLanguageServiceWithCache } from 'typescript-auto-import-cache';
 
 export interface LanguageServiceContainer {
     readonly tsconfigPath: string;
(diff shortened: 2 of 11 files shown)


packages/language-server/src/plugins/typescript/service.ts
```diff
@@ -294,22 +280,25 @@ async function createLanguageService(
...
-    const languageService = ts.createLanguageService(host, documentRegistry);
     const transformationConfig: SvelteSnapshotOptions = {
         parse: svelteCompiler?.parse,
         version: svelteCompiler?.VERSION,
         transformOnTemplateError: docContext.transformOnTemplateError,
         typingsNamespace: raw?.svelteOptions?.namespace || 'svelteHTML'
     };
 
+    const project = initLsCacheProject();
+    const languageService = ts.createLanguageService(host, documentRegistry);
```
dummdidumm, line 300: `initLsCacheProject` also creates a language service, why are we creating another one?

packages/language-server/src/plugins/typescript/serviceCache.ts
```diff
@@ -0,0 +1,93 @@
...
+) {
+    const version = ts.version.split('.');
+    const major = parseInt(version[0]);
+    const minor = parseInt(version[1]);
+
+    if (major < 5) {
+        return undefined;
+    }
+
+    const factory = minor < 3 ? createProject50 : createProject53;
```
dummdidumm, line 73: This makes me a bit nervous that in the future we potentially need to adjust this for new TS versions - but I guess there's no other way around this since this is using internal methods. Ideally the cache becomes part of the TS service at some point...

packages/language-server/src/plugins/typescript/features/CompletionProvider.ts
```diff
@@ -236,7 +236,9 @@ export class CompletionsProviderImpl implements CompletionsProvider<CompletionEn
             offset,
             {
                 ...userPreferences,
-                triggerCharacter: validTriggerCharacter
+                triggerCharacter: validTriggerCharacter,
+                // Just in case, if there is a parser error. Force ts to not use the cache
+                triggerKind: tsDoc.parserError ? undefined : completionContext?.triggerKind
```
dummdidumm, line 241: Does this mean we got no advantage from this change in broken states, e.g when being in the middle of a template completion like `<Comp`?

dummdidumm (commented):
Impressive work, left a few comments

packages/language-server/src/plugins/typescript/serviceCache.ts
jasonlyu123: Yeah. This might break in the future. Maybe it'll be better to let the `typescript-auto-import-cache` handle it, but there is the integration problem I mentioned above. I talked with @/johnsoncodehk, but he's not sure if the new API we added now in `typescript-auto-import-cache` can be available later since the potential changes in the future and with Volar 2.

dummdidumm: Just to be sure: This doesn't mean that `typescript-auto-import-cache` isn't going away at some point, it's just that the APIs we need on top of it cannot be built into the package itself right now because of the potential Volar 2 changes.

packages/language-server/src/plugins/typescript/features/CompletionProvider.ts
jasonlyu123: I was wrong. The incompletion cache requires another internal API. And it seems only rarely to trigger. So I think it's not worth the effort to add it now. Maybe we can add it later or add it to Volar when we switch. 

In the module resolution "node", TypeScript only loads 100 of them and marks the completion as incomplete. That's when the incomplete cache can be handy. The completion is reused and only resolves more completion label detail, the faded module specifier text after the completion label. But In most of the cases I tested, It only marks it as incomplete in global completion, and most of the time, the first letter can already narrow the number down. 

In module resolution node16+ and bundler, TypeScript will check all the module specifiers in auto import. It is probably to filter out auto import blocked by the export map. And create-svelte recently switched to bundler by default. That explains why, all of a sudden, many people started to complain about slow completion. 

In the case of `<Comp`. Do you remember why we'll reuse the completion in case of `triggerKind === CompletionTriggerKind?TriggerForIncompleteCompletions`? It seems like we'll reuse the completion in this situation before asking ts for a new list. Is that the intended behaviour?

packages/language-server/src/plugins/typescript/serviceCache.ts
jasonlyu123: Yeah. Not saying the package is going away. I checked the chat history. What he said is that he might "bundle" the code responsible for the cache.

packages/language-server/src/plugins/typescript/features/CompletionProvider.ts
dummdidumm: I think it was intentional because completion is rather expensive in this situation and if people type one character after the other they should be able to narrow down the existing completions.
Is this resulting in undesired behavior now?

jasonlyu123: I think it's mostly unavoidable. If there is a cache, it at least only happens the first time and after various caches are invalidated. https://github.com/sveltejs/language-tools/issues/2244 might have something to optimize on the TypeScript side, but I think It is an edge case. I'm not sure if there is something less extreme and also having performance issues.

Issue #2155, opened 2023-09-12
Destructuring inline event parameters breaks IntelliSense


ignatiusmb:
### Describe the bug

Having an inline event handler inside an element (doesn't have to be a special element) breaks `<style>` tag IntelliSense.

### Reproduction

Copy and paste the snippet to an empty `.svelte` file and trigger the autocomplete box in the styles tag

```svelte
<svelte:window
	on:mousemove={({}) => {
		// destructuring the parameter above breaks autocompletion
		// remove the empty brackets and the css autocompletion below works
	}}
/>

<div><slot /></div>

<style>
	div {
		/* press space, gets JS autocomplete instead of CSS */
	}
</style>
```

### Expected behaviour

CSS autocompletion works without having to move destructuring inside the function body.

### System Info

- OS: Windows
- IDE: VSCode

### Which package is the issue about?

Svelte for VS Code extension

ShadiestGoat:
Its not just for destructuring (though I can replicate that as well). In this snippet, it breaks with the same symptoms:

```svelte
<svelte:window
    on:mousemove={() => {
        if (true) {return}
        if (2 >= 1) {}
    }}
/>

<style lang="scss">
    /* js suggestions here :// */
</style>
```

PR #2259, opened 2024-01-13
Fix #2175 - Support multiple style tags in one document

### LINKED ISSUE #2175: Intelisense Is Not Working in ``<svelte:head>`` Tag
### Describe the bug

Not sure if this is expected behavior, but I found that inside the <svelte:head> tag, intelisense just not working when typing or hit ``Ctrl + Space``.

### Reproduction

```svelte
<svelte:head>
<style>
    .test-class {
        color: red;
    }
</style>
</svelte:head>
```

### Expected behaviour

Intellisense should work normally.

### System Info

- OS: Windows 10
- IDE: VSCode 1.80.0

_No response_

_No response_

amjadorfali:
## Outline
This PR aims to fix #2175, by providing support to have multiple `<style>` tags in one document. The solution we're looking for should cover the following cases:

- [ ] Need to support multiple style tags in one doc
- [ ] Need to support autocompletion
- [ ] Need to support diagnostics
- [ ] Need to add test cases for svelte:head <style> tag

## The how? 🤔 

Initially I was thinking we simply need to select the appropriate style tag when we're at the appropriate position. It turns out it ain't that simple, as we have support for auto completions such as HTML & TS, and they sometimes overlap.

I do not have a clear solution in my head (pun intended) ATM, I certainly have to get familiarized with the code more in order to provide a better quality solution that covers all the use cases we have.

### DIFF
--- a/packages/language-server/src/lib/documents/utils.ts
+++ b/packages/language-server/src/lib/documents/utils.ts
@@ -59,7 +59,10 @@ function extractTags(
     html?: HTMLDocument
 ): TagInformation[] {
     const rootNodes = html?.roots || parseHtml(text).roots;
-    const matchedNodes = rootNodes
+    const headIdx = tag === 'style' ? rootNodes.findIndex((node) => node.tag === 'svelte:head') : -1;
+    const styleWithinHead = headIdx !== -1 ? rootNodes[headIdx]?.children?.find((node) => node.tag === 'style') : undefined;
+
+    const matchedNodes = [...rootNodes, ...(styleWithinHead ? [styleWithinHead] : [])]
         .filter((node) => node.tag === tag)
         .filter((tag) => {
             return isNotInsideControlFlowTag(tag) && isNotInsideHtmlTag(tag);
@@ -147,15 +150,18 @@ export function extractScriptTags(
     return { script, moduleScript };
 }
 
-export function extractStyleTag(source: string, html?: HTMLDocument): TagInformation | null {
+
+type ReturnType<T> = T extends true ? TagInformation[] | null : TagInformation | null
+
+export function extractStyleTag<T extends boolean = false>(source: string, html?: HTMLDocument, returnAll?: T): ReturnType<T> {
     const styles = extractTags(source, 'style', html);
     if (!styles.length) {
         return null;
     }
-
-    // There can only be one style tag
-    return styles[0];
+   
+    return (returnAll? styles : styles[0]) as ReturnType<T>;
 }
+
 
 export function extractTemplateTag(source: string, html?: HTMLDocument): TagInformation | null {
     const templates = extractTags(source, 'template', html);
--- a/packages/language-server/src/plugins/PluginHost.ts
+++ b/packages/language-server/src/plugins/PluginHost.ts
@@ -125,7 +125,7 @@ export class PluginHost implements LSProvider, OnWatchFileChanges {
     ): Promise<CompletionList> {
         const document = this.getDocument(textDocument.uri);
 
-        const completions = await Promise.all(
+        let completions = await Promise.all(
             this.plugins.map(async (plugin) => {
                 const result = await this.tryExecutePlugin(
                     plugin,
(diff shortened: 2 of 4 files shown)


amjadorfali:
@jasonlyu123 PR is ready for discussion, please feel free to check it out whenever possible 🙏🏼

jasonlyu123 (commented):
Thanks for the PR. I prefer a more specialized approach to this instead of changing shared functions. I like the idea of extracting in `getCompletions` using the position info. We could have something similar to the style attribute. 

Were you intending to add this to more CSS features, or is completion and hover enough? I thought we could also add the support for the style tag nested under any element instead of `svelte:head`. This might help #2155, which is because the top level check is flawed but hard to fix. But for performance, completion and hover should be enough for the nested under any element situation.

amjadorfali:
@jasonlyu123 Thank you for your feedback 🙏🏼 . I will address the first issue first then move on to the requested changes as they might change based on this discussion 

> Were you intending to add this to more CSS features, or is completion and hover enough? I thought we could also add the support for the style tag nested under any element instead of svelte:head. This might help https://github.com/sveltejs/language-tools/issues/2155, which is because the top level check is flawed but hard to fix. But for performance, completion and hover should be enough for the nested under any element situation.

That was exactly my concern. Based on my current trajectory, I can maybe fix the auto completion, but the hover and diagnostics will require a whole different approach, we essentially need to support having 2 `styleInfo` in one CssDocument
I tested this theory with `.html` files yesterday, and apparently the support having as many style tags as needed, all with autoCompletion and diagnostics like `Do not use empty rulesetscss(emptyRules)`

So how do you think we should tackle this? Should we just  support AutoCompletions, or think of a complete solution that covers all edge cases?

jasonlyu123:
Html doesn't have to support css preprocessor so they just replace the js/html part with whitespace and line break. But we'll have to extract them into multiple sub documents. 

https://github.com/microsoft/vscode/blob/70328b1b48dcc4e0f35fe72c592f25057bd82096/extensions/html-language-features/server/src/modes/embeddedSupport.ts#L163

They are using a scanner to extract the style tags. 
https://github.com/microsoft/vscode/blob/789c802e75813bfb38ff2e13e4b322bd7e90124b/extensions/html-language-features/server/src/modes/embeddedSupport.ts#L37

We also have access to the API but we probably could extract it from HTML AST like you did. Since we'll always parse it after file opened or updated. We just need to walk further down the tree to find the nested one. As for performance, You can also try if using a scanner is faster. Another optimization I can think of is to check if there is another "<style" in document before walking the AST. string.include is usually fast and it pretty rare that there is multiple style tag so it's worth trying if it is faster.

amjadorfali:
@jasonlyu123 Sorry to bother you too much on this and taking your time, but I'm lost on what needs to be done.

I do find this issue interesting and want to fix it, but i can't wrap my head around the requirements. Can we maybe sync up on a chat or call? 

If that's not feasible, a simple list of requirements/stuff to look out for would suffice, and i can take it from there. 

Some of questions in mind: 

- Do we need to support all types of checks for all style tags? (auto completion, diagnostics ...)
  - If that's the case, can we change the code to support an array of `styleInfo` objects per doc?

jasonlyu123:
Nested styles are uncommon, so I think supporting only a few features is fine. But if you want and can get it to work, You can make it work in all features. One concern I have is the performance impact of managing multiple style tags per file. Maybe it isn't big but you'll only know how big of a difference it makes when you implement it. 

> Can we change the code to support an array of styleInfo objects per doc?

It doesn't make sense. The styleInfo is used to map the position relative to the start of the style tag to the position of the document. You can change it to allow injecting the style info object from the constructor so the logic can be reused. The logic to maintain and extract nested style tags can be within the `CSSPlugin` or the `Document`. And the `getCSSDoc` can probably be split into `getCSSDocAtPosition` and `getAllCSSDoc`.

amjadorfali:
Thanks a lot @jasonlyu123 
Okay so based on what you suggested, I will first try to solve the issue for Completions, and afterwards try to think of a solution that covers all the features.

amjadorfali:
I'm sorry i couldn't solve this issue, I'm closing this PR and letting someone else handle it

PR #2279, opened 2024-02-05
chore: bump svelte-preprocess

sajaghori:
Currently I get a dependency warning for `svelte-check 3.6.3` which according to pnpm still does not support `postcss-load-config 5.0.2`. With [v5.1.2](https://github.com/sveltejs/svelte-preprocess/releases/tag/v5.1.2) they added support for postcss load config v5.

### DIFF
--- a/packages/language-server/package.json
+++ b/packages/language-server/package.json
@@ -55,7 +55,7 @@
         "prettier": "~3.1.0",
         "prettier-plugin-svelte": "~3.1.1",
         "svelte": "^3.57.0",
-        "svelte-preprocess": "~5.1.0",
+        "svelte-preprocess": "^5.1.3",
         "svelte2tsx": "workspace:~",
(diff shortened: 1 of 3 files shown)


dummdidumm:
`svelte-preprocess` is a dependency of `svelte-check`, so you should be able to bump it yourself (probably through `npm update svelte-preprocess`).
As for bumping it here: Please also bump the `svelte-preprocess` versions in the other packages so that their versions stay in sync.

sajaghori:
> `svelte-preprocess` is a dependency of `svelte-check`, so you should be able to bump it yourself (probably through `npm update svelte-preprocess`). As for bumping it here: Please also bump the `svelte-preprocess` versions in the other packages so that their versions stay in sync.

You are right `pnpm update svelte-preprocess` did it for me. I bumped the version in the other package as well. :)

Issue #2233, opened 2023-12-19
File watching outside of the directory of tsconfig


datstarkey:
### Describe the bug

When working in a monorepo, you can use path alias and includes from the tsconfig to import svelte files from shared libraries.

This works fine until you add a new file inside a shared package, it just does not appear in the intellisense until you restart the svelte language server.

example svelte.config.js and tsconfig.json

```javascript
import adapter from '@sveltejs/adapter-auto';
import { vitePreprocess } from '@sveltejs/vite-plugin-svelte';

/** @type {import('@sveltejs/kit').Config} */
const config = {
    // Consult https://kit.svelte.dev/docs/integrations#preprocessors
    // for more information about preprocessors
    preprocess: vitePreprocess(),

    kit: {
        // adapter-auto only supports some environments, see https://kit.svelte.dev/docs/adapter-auto for a list.
        // If your environment is not supported or you settled on a specific environment, switch out the adapter.
        // See https://kit.svelte.dev/docs/adapters for more information about adapters.
        adapter: adapter(),

        alias: {
            "$ui": "../ui/src/lib",
            "$ui/*": "../ui/src/lib/*",
        }
    }
};

export default config;

```

And inside the tsconfig.json it has the ui src/lib folder in includes

```json
{
  "extends": "./.svelte-kit/tsconfig.json",
  "compilerOptions": {
    "allowJs": true,
    "checkJs": true,
    "esModuleInterop": true,
    "forceConsistentCasingInFileNames": true,
    "resolveJsonModule": true,
    "skipLibCheck": true,
    "sourceMap": true,
    "strict": true,
    "moduleResolution": "bundler"
  },
  "include": [
    "./.svelte-kit/ambient.d.ts",
    "./.svelte-kit/non-ambient.d.ts",
    "./.svelte-kit/types/**/$types.d.ts",
    "./vite.config.js",
    "./vite.config.ts",
    "./src/**/*.js",
    "./src/**/*.ts",
    "./src/**/*.svelte",
    "../ui/src/**/*.js",
    "../ui/src/**/*.ts",
    "../ui/src/**/*.svelte"
  ]
}

```

### Reproduction

https://github.com/datstarkey/svelte-monorepo-import-issue

Reproduction instructions given in readme of repo.

### Expected behaviour

Any files added should appear in the auto-complete list.

I should note this does work in typescript files just fine, its only .svelte files that it doesn't work.

### System Info

- OS: Windows
- IDE: VS Code
[...]

PR #2299, opened 2024-02-24
fix: watch svelte files and project files outside workspace

jasonlyu123:
#2233 #2393

update project files(tsconfig.include) when a new client file is opened. So files included in both tsocnfig.json will be loaded into the respecting language service.


packages/language-server/src/lib/documents/DocumentManager.ts
```diff
@@ -48,14 +48,15 @@ export class DocumentManager {
         if (this.documents.has(textDocument.uri)) {
             document = this.documents.get(textDocument.uri)!;
             document.setText(textDocument.text);
+            document.openedByClient = openedByClient;
         } else {
             document = this.createDocument(textDocument);
             this.documents.set(textDocument.uri, document);
+            document.openedByClient = openedByClient;
```
dummdidumm, line 55:
```suggestion
```

packages/language-server/src/lib/documents/DocumentManager.ts
```diff
@@ -48,14 +48,15 @@ export class DocumentManager {
...
             this.notify('documentOpen', document);
         }
 
         this.notify('documentChange', document);
```
dummdidumm, line 59:
```suggestion
        document.openedByClient = openedByClient;
        this.notify('documentChange', document);
```

packages/language-server/src/plugins/typescript/LSAndTSDocResolver.ts
```diff
@@ -66,7 +66,13 @@ export class LSAndTSDocResolver {
         // where multiple files and their dependencies
         // being loaded in a short period of times
         docManager.on('documentOpen', (document) => {
+            const path = document.getFilePath();
+            const isNewFile = path && !this.globalSnapshotsManager.get(path);
             handleDocumentChange(document);
+            if (isNewFile && document.openedByClient) {
```
dummdidumm, line 72: Does this mean we would kick off the costly "update project files" logic on _every_ new file created through the editor UI? Is there a way to do this in less cases? "update project files" reads the file system, creates a new program version etc, which seems costly.

jasonlyu123: I changed to detection to file watch since #2393 also needs it. I also delay the check to the next time the language service is used. TypeScript also has some directory watching info in the config file parse result. So we can skip the project file check if the file isn't in the target directories. This info is also used to watch directories outside the root using the fallback watch or when the client supports it. This is added because the optimization check might break a way to workaround directory watching outside the workspace, not sure if anyone actually uses it though 😅.

dummdidumm (approved):

Issue #2245, opened 2024-01-04
Completion for custom element tag name


break-stuff:
### Description

There doesn't seem to way to extend the language tool for custom elements.

### Proposed solution

Vue and JSX templates have an interface that can be extended to include custom elements/web components. This would be a nice way for library authors to provide better type integration and descriptions within svelte templates.

jasonlyu123:
Yes. There is .

https://svelte.dev/docs/typescript#enhancing-built-in-dom-types

break-stuff:
Oh, this is fantastic! Thank you!

break-stuff:
@jasonlyu123 do you have a working example of this? It doesn't seem to work in my sample app.

dummdidumm:
Can you provide a link to that example app?

break-stuff:
Yeah, here's a simple repo.

https://github.com/break-stuff/svelte-custom-elements-demo

jasonlyu123:
It does work in your repo. Are you expecting to have completion for it as well?

break-stuff:
Yes, I don't see any difference in having it vs not having it in my repo.

jasonlyu123:
The difference is it will be type-check for the element.

jasonlyu123:
There is another way to add custom tag name completion. It's only a local config, so it's not something library authors can leverage, but it's still a good workaround until we can figure out a way to provide the completion. 

https://github.com/microsoft/vscode-html-languageservice/blob/main/docs/customData.md

break-stuff:
Yeah, I created a package to generate that, but it would be nice if devs would only have to set one thing for types and autocomplete.

https://www.npmjs.com/package/custom-element-vs-code-integration

Issue #1560, opened 2022-07-16
on:touch events are not in the intellisense list


Jojoshua:
### Describe the bug

I started typing on:touch... and no intellisense showed up so I assumed touch events were not available to use. I went ahead and fully typed it out and found no error showed up and found these did exist in svelte-jsx.d.ts after all.

   // Touch Events
      ontouchcancel?: TouchEventHandler<T> | undefined | null;
      ontouchend?: TouchEventHandler<T> | undefined | null;
      ontouchmove?: TouchEventHandler<T> | undefined | null;
      ontouchstart?: TouchEventHandler<T> | undefined | null;

### Reproduction

- Start typing a touch event `<div on:tou `
- Observe no intellisense shows up

### Expected behaviour

All available events are shown for intellisense

### System Info

- OS: Win 10
- IDE: VSCode

Jojoshua:
Did the new transformation land yet?

PR #2300, opened 2024-02-25
feat: completion for custom element tag name and event

jasonlyu123:
For #1560 and #2245. Adding completion for element tag name from custom type definition and some small performance tweaks with completion. 

#1560 is already showing but needed some `TextEdit` so it doesn't end up with on:on:input


dummdidumm (approved):
Nice!

PR #2313, opened 2024-03-11
feat: TypeScript 5.4 support for svelte-check

jasonlyu123:
Auto import seems to be broken. Need to check what the problem is. We also need to check if the [Preserved Narrowing in Closures Following Last Assignments](https://devblogs.microsoft.com/typescript/announcing-typescript-5-4/#preserved-narrowing-in-closures-following-last-assignments) changes the narrowing in any unexpected ways (#2316). 

**Auto Import in non-ts svelte file has some regression in 5.4. So this only adds the support for svelte-check. Editor support waiting on if the upstream fix is merged. So 5.4 might be skipped.**


jasonlyu123:
Seems like auto-import is broken in js/non-ts-svelte files importing `export const`, `export class`... but not `class A {}; export { A }`

jasonlyu123:
Yeah. it also happens in js files. Also found that the problem with `onMount` in svelte 3 isn't because it has a `declare` modifier. It's because of the barrel export. Barrel export seems like a way bigger problem than the default export. Upstream issue https://github.com/microsoft/TypeScript/issues/57740

jasonlyu123:
Let's wait until VSCode releases with TypeScript 5.4 to see if enough people complain about it and see if the fix can made into a 5.4 patch release. If not, we can probably either

1. Adjust the test and add a dedicated test for auto-import reexport/default export. Skipped it and enabled it when the fix was released.
2. Pin the version in 5.3 but release the fixes with binding control flow and transition type-check.

dummdidumm:
The bug was put into the backlog, it doesn't look like it has any priority. We may have to dig into the code base ourselves and make a TS pull request. Would you be willing to try that?
In the meantime doing option 2 sounds good.

packages/svelte2tsx/src/htmlxtojsx_v2/nodes/Binding.ts
```diff
@@ -94,6 +94,11 @@ export function handleBinding(
         return;
     }
 
+    const expressionStr = str.original.substring(attr.expression.start, getEnd(attr.expression));
```
dummdidumm, line 97:
```suggestion
    // add reassignment to force TS to widen the type of the declaration (in case it's never reassigned anywhere else)
    const expressionStr = str.original.substring(attr.expression.start, getEnd(attr.expression));
```

dummdidumm (approved):

DePasqualeOrg:
It looks like the TypeScript version in the svelte-check package still needs to be updated. I'm getting errors when running svelte-check that don't appear in VS Code, where I'm running the latest TypeScript.

Issue #2307, opened 2024-03-06
Volar hybrid mode compatibility with typescript-svelte-plugin


jasonlyu123:
### Describe the bug

Thread for tracking Volar hybrid mode compatibility issues. Documenting some issues currently found and continuing the conversation of https://github.com/microsoft/TypeScript/issues/57630.

1. Typescript plugin is enabled in all projects without a tsconfig/jsconfig.json
2. The plugin is enabled by default. The config request to disable it only happens when a ts/js file is opened. So the `"svelte.enable-ts-plugin": false` config doesn't do anything until a ts/js file is opened. 
3. Debug.assert in ts 5.5 nightly: https://github.com/microsoft/TypeScript/blob/353ccb7688351ae33ccf6e0acb913aa30621eaf4/src/server/project.ts#L1160. Doesn't seem to happen when opening a svelte project.

For 2, we might need to disable the ts plugin when the config is `{ global: true, ... }` and enable it when the config request we sent arrives.

Probably Volar problems:

1. Debug.assert in ts 5.3: https://github.com/microsoft/TypeScript/blob/63717cf44a64585175c76c336c8f8d87d5384d16/src/server/project.ts#L1518. This is because vue typescript plugin patched `getScriptVersion` to be a compound version id but TypeScript expect update should only happen through `Project.updateGraph`.  Might need to check if this only happens if the Svelte extension is enabled.

@johnsoncodehk, Also, I can't reliably reproduce the problem reported in https://github.com/microsoft/TypeScript/issues/57630. Can you provide a more detailed step?

### Reproduction

https://github.com/microsoft/TypeScript/issues/57630

### Expected behaviour

compatible

### System Info

- OS: [e.g. Windows]
- IDE: [e.g. VSCode, Atom]

PR #2317, opened 2024-03-14
fix: volar ts plugin compatibility

jasonlyu123:
#2307 

1. Improve the svelte project check for `InferredProject`. Start from the current directory and go down 2 levels in the directory tree to search for package.json. Then we use the directories with package.json to search for svelte modules. If your ts-plugin is no longer enabled, you can create a `jsconfig.json` where your `package.json` is. This alone should resolve most problems since the plugin won't be applied at all. 

2. Disable the plugin in the config level until we receive the ` _typescript.configurePlugin ` request. This should decrease the chance of the project update because of the config toggle.

As for the Debug.assert in ts 5.5 nightly, this happens because ts clears `.d.ts` loaded by open files in `InferredProject`. So in non-svelte projects, the `.d.ts` from svelte might be clear whenever a file is closed. Haven't found a good way to resolve this but a better svelte project detection should already lower the chance happens.

@johnsoncodehk could you check if this resolves the compatibility issues?


jasonlyu123:
> As for the Debug.assert in ts 5.5 nightly, this happens because ts clears .d.ts loaded by open files in InferredProject.

This actually happens in 5.3 as well. It also happens to Svelte projects without tsconfig/jsconfig. We can probably fix it in a separate PR.

PR #2330, opened 2024-04-05
feat: add editor label and formatter configuration defaults

### LINKED ISSUE #15139: Default VS Code label values
### Describe the problem

Folks are busy with async right now, but when we switch back to working on SvelteKit we should close out the VS Code labels PR. While it's technically part of language tools, it's for SvelteKit support and is hugely requested, which would help make the SvelteKit 3.0 launch a bit splashier

### Describe the proposed solution

Review https://github.com/sveltejs/language-tools/pull/2330

### Alternatives considered

_No response_

### Importance

nice to have

### Additional Information

_No response_

teemingc:
Uses the VS Code extension [configuration defaults](https://code.visualstudio.com/api/references/contribution-points#contributes.configurationDefaults) contribution point to:
* set the default formatter for svelte files [as mentioned in the README](https://github.com/sveltejs/language-tools/tree/master/packages/svelte-vscode#setup) (which some people may have missed)
* uses the new VS Code [custom labels](https://code.visualstudio.com/updates/v1_88#_custom-labels-for-open-editors) setting to append the directory name to the editor label for SvelteKit +page and +server files.

### Default

default editor labels

### Label format

```json
"workbench.editor.labelFormat": "short"
```
label format: short

### Custom labels

```json
  "workbench.editor.customLabels.patterns": {
    "**/+page.*": "${dirname}/${filename}.${extname}",
    "**/+server.*": "${dirname}/${filename}.${extname}",
  },
```
custom vscode editor labels


benmccann:
I wonder if we could drop `routes` from the example. It makes sense to include the directory for URLs like `/about`, but we don't necessarily need it for the root directory where we could show just `/+page.js`. Maybe we can have another entry earlier without the leading `**`?

I also wonder if we could show `/about/+page.js` if it's only one level deep and `.../[id]/+page.js` if it's multiple levels deep

packages/svelte-vscode/package.json
```diff
@@ -396,6 +396,16 @@
                 }
             }
         },
+        "configurationDefaults": {
+            "[svelte]": {
```
dummdidumm, line 400: I'm not 100% sure we want this because people can also use the prettier extension, if that ones installed I'd actually prefer that one was used assuming people have both prettier and the plugin installed.
The label thing is a super nice addition.

benmccann:
> I had this initially too but realised it could break if anyone changed the files.routes config and it no longer points to src/routes. But this should be a minority of users? https://kit.svelte.dev/docs/configuration#files

We could potentially duplicate the rules and remove the `src/routes` from the second set as a fallback for users who have changed that config setting

packages/svelte-vscode/package.json
eltigerchino: What's the difference between the two? I've been using the svelte formatter as default although I have both installed. Mainly because setting prettier as the default formatter prompts me to change the default formatter when I try to format a svelte file in a project that has no prettier installed or prettier config.

eltigerchino:
> > I had this initially too but realised it could break if anyone changed the files.routes config and it no longer points to src/routes. But this should be a minority of users? [kit.svelte.dev/docs/configuration#files](https://kit.svelte.dev/docs/configuration#files)
> 
> We could potentially duplicate the rules and remove the `src/routes` from the second set as a fallback for users who have changed that config setting

I'm not sure how to get the fallback rules to recognise the nesting level since without knowing the `files.routes` configuration. I did add a very general fallback rule to just append the directory name which isn't as nice but maybe it's better than nothing

eltigerchino:
Another suggestion regarding layout groups:

If you had layout groups, your tabs wouldn't give you the context of what route the file if for:
CleanShot 2024-04-06 at 3  42 14@2x

So maybe we can show the directory before the layout group?

CleanShot 2024-04-06 at 3  43 12@2x

```json
"src/routes/(*)/+[layout|page|server]*": "/${filename}.${extname} ${dirname}",
"src/routes/*/**/(*)/+[layout|page|server]*": "/${dirname(1)}/${filename}.${extname} ${dirname(0)}",
```

paoloricciuti:
> I wonder if we could drop `routes` from the example. It makes sense to include the directory for URLs like `/about`, but we don't necessarily need it for the root directory where we could show just `/+page.js`. Maybe we can have another entry earlier without the leading `**`?
> 
> I also wonder if we could show `/about/+page.js` if it's only one level deep and `.../[id]/+page.js` if it's multiple levels deep

I tried exactly this in my configuration but removed it because:

1. With very nested stuff it can get very long and it will still not reflect the actual route
2. With not so nested stuff it will include `routes` which I didn't like
3. In the end I got already used to the current version and I prefer being precise and not changing stuff and maybe make it more confusing

ottomated:
This is what I came up with:

```
"workbench.editor.customLabels.patterns": {
	// Root
	"**/routes/+layout.svelte": "/ [layout]",
	"**/routes/+page.server.ts": "/ [server]",
	"**/routes/+layout.server.ts": "/ [layout.server]",
	"**/routes/+server.ts": "/ [endpoint]",
	"**/routes/+page.svelte": "/ [page]",
	// 1st level
	"**/routes/*/+layout.svelte": "/${dirname} [layout]",
	**/routes/*/+page.server.ts": "/${dirname} [server]",
	"**/routes/*/+layout.server.ts": "/${dirname} [layout.server]",
	"**/routes/*/+server.ts": "/${dirname} [endpoint]",
	"**/routes/*/+page.svelte": "/${dirname} [page]",
	// Nested
	"**/routes/*/*/**/+layout.svelte": "${dirname(1)}/${dirname} [layout]", 
	"**/routes/*/*/**/+page.server.ts": "${dirname(1)}/${dirname} [server]",
	"**/routes/*/*/**/+layout.server.ts": "${dirname(1)}/${dirname} [layout.server]",
	"**/routes/*/*/**/+server.ts": "${dirname(1)}/${dirname} [endpoint]",
	"**/routes/*/*/**/+page.svelte": "${dirname(1)}/${dirname} [page]",
}
```

Very convoluted but this makes things like `src/routes/+page.svelte` -> `/ [page]` but `src/routes/api/nested/endpoint/+server.ts` -> `nested/endpoint [endpoint]`.

benmccann:
>I tried exactly this in my configuration but removed it because:
>With not so nested stuff it will include routes which I didn't like

@paoloricciuti I'm not sure you did try the same thing then. My suggestion was that `routes` should never be shown

@ottomated your setup can't differentiate between `+page.svelte` and `+page.js`

packages/svelte-vscode/package.json
```diff
@@ -396,6 +396,17 @@
...
+                "editor.defaultFormatter": "svelte.svelte-vscode"
+            },
+            "workbench.editor.customLabels.patterns": {
+                "src/routes/+[layout|page|server]*": "/${filename}.${extname}",
+                "src/routes/*/+[layout|page|server]*": "/${dirname}/${filename}.${extname}",
+                "src/routes/*/*/**/+[layout|page|server]*": ".../${dirname}/${filename}.${extname}",
```
benmccann, line 406: this didn't work for me in a monorepo because `src` was contained within another directory. I think we might need a leading `**/`
```suggestion
                "**/src/routes/+[layout|page|server]*": "/${filename}.${extname}",
                "**/src/routes/*/+[layout|page|server]*": "/${dirname}/${filename}.${extname}",
                "**/src/routes/*/*/**/+[layout|page|server]*": ".../${dirname}/${filename}.${extname}",
```

ottomated:
> @ottomated your setup can't differentiate between `+page.svelte` and `+page.js`

Good catch, here's a more comprehensive version:

```json5
"workbench.editor.customLabels.patterns": {
	// Root
	"**/routes/+layout.svelte": "/ [layout]",
	"**/routes/+layout.server.[jt]s": "/ [layout.server.load]",
	"**/routes/+layout.[jt]s": "/ [layout.load]",
	"**/routes/+page.svelte": "/ [page]",
	"**/routes/+page.server.[jt]s": "/ [server.load]",
	"**/routes/+page.[jt]s": "/ [load]",
	"**/routes/+server.[jt]s": "/ [endpoint]",
	// 1st level
	"**/routes/*/+layout.svelte": "/${dirname} [layout]",
	"**/routes/*/+layout.server.[jt]s": "/${dirname} [layout.server.load]",
	"**/routes/*/+layout.[jt]s": "/${dirname} [layout.load]",
	"**/routes/*/+page.svelte": "/${dirname} [page]",
	"**/routes/*/+page.server.[jt]s": "/${dirname} [server.load]",
	"**/routes/*/+page.[jt]s": "/${dirname} [load]",
	"**/routes/*/+server.[jt]s": "/${dirname} [endpoint]",
	// Nested
	"**/routes/*/*/**/+layout.svelte": "/${dirname(1)}/${dirname} [layout]",
	"**/routes/*/*/**/+layout.server.[jt]s": "/${dirname(1)}/${dirname} [layout.server.load]",
	"**/routes/*/*/**/+layout.[jt]s": "/${dirname(1)}/${dirname} [layout.load]",
	"**/routes/*/*/**/+page.svelte": "/${dirname(1)}/${dirname} [page]",
	"**/routes/*/*/**/+page.server.[jt]s": "/${dirname(1)}/${dirname} [server.load]",
	"**/routes/*/*/**/+page.[jt]s": "/${dirname(1)}/${dirname} [load]",
	"**/routes/*/*/**/+server.[jt]s": "/${dirname(1)}/${dirname} [endpoint]",
},
```

sacrosanctic:
what about `+error.svelte`?

teemingc:
Adding a reminder for myself to try elstalky’s config shared on discord:

i'm using these:
```json
"workbench.editor.customLabels.patterns": {
    "**/routes/{+server,+layout,+page,+error}.{svelte,ts,cs,server.ts,js,server.js}": "/ (${filename}.${extname})",
    "**/routes/*/{+server,+layout,+page,+error}.{svelte,ts,cs,server.ts,js,server.js}": "/${dirname} (${filename}.${extname})",
    "**/routes/*/*/{+server,+layout,+page,+error}.{svelte,ts,cs,server.ts,js,server.js}": "/${dirname(1)}/${dirname} (${filename}.${extname})",
    "**/routes/*/*/*/{+server,+layout,+page,+error}.{svelte,ts,cs,server.ts,js,server.js}": "/${dirname(2)}/${dirname(1)}/${dirname} (${filename}.${extname})",
    "**/routes/**/*/*/*/*/{+server,+layout,+page,+error}.{svelte,ts,cs,server.ts,js,server.js}": "…/${dirname(2)}/${dirname(1)}/${dirname} (${filename}.${extname})"
},
```
i.e. `/route/here (filename)`

packages/svelte-vscode/package.json
```diff
@@ -415,6 +415,14 @@
                 }
             }
         },
+        "configurationDefaults": {
+            "workbench.editor.customLabels.patterns": {
+                "**/src/routes/+[layout|page|server|error]*": "/${filename}.${extname}",
+                "**/src/routes/*/+[layout|page|server|error]*": "/${dirname}/${filename}.${extname}",
+                "**/src/routes/*/*/**/+[layout|page|server|error]*": ".../${dirname}/${filename}.${extname}",
+                "**/+[layout|page|server|error]*": "${dirname}/${filename}.${extname}"
```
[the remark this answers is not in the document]
teemingc: Personally, I've been using `..` as a separator such as `/about .. +page.svelte` rather than the parenthesis suggested here

PR #2339, opened 2024-04-19
(fix) add field-sizing

tkhs0813:
closed https://github.com/sveltejs/language-tools/issues/2340


tkhs0813:
@jasonlyu123 
Could you please confirm?

jasonlyu123:
We don't usually add experiment CSS feature with limited browser support and this is an experiment property that only chromium support.

tkhs0813:
@jasonlyu123 
Understood. In that case, are there any options to avoid this issue?

jasonlyu123:
No. In the editor, there is a `css.customData` option but it's not available for svelte-check. But since this is a CSS property with poor browser support I'll say this is a valid warning.

tkhs0813:
Understood. Thank you.

PR #2347, opened 2024-04-24
fix: Preserve event attribute casing in svelte 5

### LINKED ISSUE #2345: don't lowercase event attributes in Svelte 5
### Describe the bug

Event attributes are treated differently in Svelte 5, we shouldn't lowercase them.

### Reproduction

onClick is not a type error

### Expected behaviour

onClick is a type error

### System Info

Irrelevant 

svelte2tsx

_No response_

jjones315:
Fixes #2345 

When running svelte 5+ any attributes beginning with "on*" after left in their original casing. 

may be better to check the knownevents.ts list, but not sure its ideal to maintain the event list in multiple places in addition to the performance of searching the array for each attribute.


dummdidumm (approved):
Not checking a list of events is the right approach as this applies to _any_ event. Thank you!

PR #2356, opened 2024-04-29
Bump TypeScript version in svelte-check

DePasqualeOrg:
I'm getting errors with svelte-check that don't appear in VS Code, where I'm running the latest TypeScript. I think this should solve them.


jasonlyu123:
The version here meant the minimum support version. You should bump the version in your project instead. You can install typescript as a dependency so it's easier to control which version you're using.

DePasqualeOrg:
https://github.com/sveltejs/language-tools/issues/2357

PR #2372, opened 2024-05-15
feat: allow `as` expressions for bindable props

paoloricciuti:
As per [discussion on discord](https://discord.com/channels/457912077277855764/1153350350158450758/1240321241290899507) this change allow for declaring the type of a bindable prop using the `as type` expression

eg.

```svelte
<!-- Parent Component -->
<script lang=ts>
  import Child from './Child.svelte';
  let amount = $state(10);
</script>
<Child bind:amount />

<!-- Child Component -->
<script lang=ts>
  import Child from './Child.svelte';
  let {
    amount = $bindable(0) as number,
  } = $props();
</script>
{amount}
```

i would just need a bit of guidance to determine where to write the test.

### DIFF
--- a/packages/language-server/test/plugins/typescript/features/diagnostics/fixtures/svelte-element-error/expected_svelte_5.json
+++ b/packages/language-server/test/plugins/typescript/features/diagnostics/fixtures/svelte-element-error/expected_svelte_5.json
@@ -3,7 +3,7 @@
         "range": { "start": { "line": 4, "character": 0 }, "end": { "line": 4, "character": 0 } },
         "severity": 1,
         "source": "ts",
-        "message": "`<svelte:element>` must have a 'this' attribute",
+        "message": "`<svelte:element>` must have a 'this' attribute with a value",
         "code": -1
     }
 ]
--- a/packages/svelte2tsx/src/svelte2tsx/nodes/ExportedNames.ts
+++ b/packages/svelte2tsx/src/svelte2tsx/nodes/ExportedNames.ts
@@ -153,7 +153,12 @@ export class ExportedNames {
                         : element.name.text;
 
                     if (element.initializer) {
-                        const call = element.initializer;
+                        let call = element.initializer;
+                        // if it's an as expression we need to check wether the as
+                        // expression expression is a call
+                        if (ts.isAsExpression(call)) {
+                            call = call.expression;
+                        }
                         if (ts.isCallExpression(call) && ts.isIdentifier(call.expression)) {
                             if (call.expression.text === '$bindable') {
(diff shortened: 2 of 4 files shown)


paoloricciuti:
@dummdidumm can i ask you where i should put the test here?

P.s. if you can give a general rule of thumb on how to determine where to put them it would be great for the next times too 😄

jasonlyu123:
Personally unsure about this because the type assert looks redundant. The behaviour also slightly deviates from a ts file where the `amount` is typed as any. But again, we did transform `export let hi: string | number` to eliminate the control flow narrowing. That also deviates from a ts file.

paoloricciuti:
> Personally unsure about this because the type assert looks redundant. The behaviour also slightly deviates from a ts file where the `amount` is typed as any. But again, we did transform `export let hi: string | number` to eliminate the control flow narrowing. That also deviates from a ts file.
> 
> https://www.typescriptlang.org/play/?ssl=11&ssc=39&pln=1&pc=1#code/DYUwLgBA3gUBEEMC2B7ArgO0gXggEgCMBLDAEwQNAAoAGASkQGcIM0kCQAnAGhgF8IuPAAdOKYYyp0A3DBjJ0WGAHplEAHoB+OTFIgAxsAScQEAGaZ9YIigz5R4yXQBciDAE85ew8dMWMVjZ2hCTklCAAPAAqAHxUAG6uUS4QUUA

Is not something I would personally do but it was pointed out and I think is worth have a safeguard against those who does it.

dummdidumm:
While I agree with @jasonlyu123 that it's weird to do that, it's still valid TS syntax and we should support it so this property is properly marked as bindable. As for the test: I propose to enhance the `ts-runes-bindable` test case with a `c = $bindable(0) as number`

paoloricciuti:
> While I agree with @jasonlyu123 that it's weird to do that, it's still valid TS syntax and we should support it so this property is properly marked as bindable. As for the test: I propose to enhance the `ts-runes-bindable` test case with a `c = $bindable(0) as number`

Added the test...for the future how can i pick between `svelte2tsx` and `htmlx2jsx`?

paoloricciuti:
@dummdidumm the test failing is unrelated to this change...should i update that here?

dummdidumm:
Yeah you can update it here.
How to differentiate: `htmlx2jsx` is for when this is purely about template transformations, `svelte2tsx` is for everything on top

paoloricciuti:
> How to differentiate: htmlx2jsx is for when this is purely about template transformations, svelte2tsx is for everything on top

Oh got it! Thanks

dummdidumm (approved):
Thank you!

PR #2378, opened 2024-05-21
feat: reference/implementation codelens

jasonlyu123:
#2130 Porting VSCode's implementation/reference code lens. This is not enabled by default. To enable this, you'll need to enable the `typescript.implementationsCodeLens.enabled` config or the `typescript.referencesCodeLens.enabled` config for `lang="ts"` and `javascript.referencesCodeLens.enabled` for js components. Note that we reuse config for ts/js files so it'll also enable the feature in ts/js files. 
 
### For non-VSCode users
Due to protocol limitations, this feature only partly works on other LSP clients. A client middleware is required to map reference info to references UI. So LSP client author might need to add support for it. For neovim-lsp users, you'll need a similar solution for the typescript language server. 

### Screenshots
圖片
(Yes. this will work for Svelte 5 props if it's a separate interface)


jasonlyu123:
No idea why, I guess they just thought it was simple enough and could be achieved with existing API.

dummdidumm (approved):

PR #2380, opened 2024-05-21
feat: Svelte 5 component class/function interop

dummdidumm:
Svelte 5 uses functions to define components under the hood. This should be represented in the types. We can't just switch to using functions though because `d.ts` files created from Svelte 4 libraries should still work, and those contain classes. So we need interop between functions and classes. The idea is therefore:
- Svelte 5 creates a default export which is both a function and a class constructor
- Various places are adjusted to support the new default exports

### DIFF
--- a/packages/svelte2tsx/src/svelte2tsx/addComponentExport.ts
+++ b/packages/svelte2tsx/src/svelte2tsx/addComponentExport.ts
@@ -206,7 +206,7 @@ function addSimpleComponentExport({
                 // `const ${className || '$$Component'} = __sveltets_2_isomorphic_component($$Component_, ${exportedNames.createBindingsStr2()});\n` +
                 `const ${className || '$$Component'} = __sveltets_2_isomorphic_component${usesSlots ? '_slots' : '2'}(${propDef});\n` +
                 surroundWithIgnoreComments(
-                    `type ${className || '$$Component'} = typeof ${className || '$$Component'};\n`
+                    `type ${className || '$$Component'} = InstanceType<typeof ${className || '$$Component'}>;\n`
                 ) +
                 `export default ${className || '$$Component'};`;
         } else {
--- a/packages/svelte2tsx/src/svelte2tsx/addComponentExport.ts
+++ b/packages/svelte2tsx/src/svelte2tsx/addComponentExport.ts
@@ -4,6 +4,7 @@ import MagicString from 'magic-string';
 import { ExportedNames } from './nodes/ExportedNames';
 import { ComponentDocumentation } from './nodes/ComponentDocumentation';
 import { Generics } from './nodes/Generics';
+import { surroundWithIgnoreComments } from '../utils/ignore';
 
 export interface AddComponentExportPara {
     str: MagicString;
@@ -21,6 +22,7 @@ export interface AddComponentExportPara {
     mode: 'ts' | 'dts' | 'tsx';
     generics: Generics;
     usesSlots: boolean;
+    isSvelte5: boolean;
     noSvelteComponentTyped?: boolean;
 }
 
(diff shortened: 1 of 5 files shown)


packages/language-server/src/plugins/typescript/features/CompletionProvider.ts
```diff
@@ -910,7 +914,16 @@ export class CompletionsProviderImpl implements CompletionsProvider<CompletionRe
...
         }
 
-        parts.push(changeSvelteComponentName(ts.displayPartsToString(displayParts)));
+        let text = changeSvelteComponentName(ts.displayPartsToString(displayParts));
+        if (
+            tsDoc.isSvelte5Plus &&
+            text.includes('(alias)') &&
+            text.includes('__SvelteComponent_')
+        ) {
+            // The info contains both the const and type export along with a bunch of gibberish we want to hide
```
jasonlyu123, line 923: The `__sveltets_2_IsomorphicComponent` type also appears in hover and the completion documentation when the component is already imported. I am wondering if it makes sense to create a type for it and use it to annotate the `Component__SvelteComponent_` variable. That way the type will be named so it won't show the internal interface and the large generic parameters. But that'll be a bit more complex because we can't use a helper function and need to write JSDoc when in js instead. And also, does having this new type and name actually make things easier to understand?

```ts
const __propsDef = render();
type Input_SvelteComponent_Function = __sveltets_2_IsomorphicComponent<typeof __propsDef['props']>;
const Input__SvelteComponent_: Input_SvelteComponent_Function = __sveltets_2_any({});
```

https://www.typescriptlang.org/play/?#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXwH0CBnANxAgxA2IICYCticBbHGABwAsswCxWHPCFQYAPACh40+AAUYODsXggAHlVTBlAJXDtgY4hhhZUAcwA08KKgCeAPisBRcqOVqNW+LoEwDRk3MrGwcrAGUIHBoVdREvH31DY1NLaztHeCdVIRhojzidPT8kwNSQjIAhU2AU91jNZQCUiXsACgBraGJiAC54AG8OBSU++UViKxBXGj6XERorYkiZ+AioiZicmgB+Wez2BfgAI2ra3fgqzVr4AF8ASj6iMgoqGnoCAElmNk4eMABhQTCURiMZKZzTDZrQ5ZLYbS41czEewAbgkElMVBgiCgYAQT3IlGotAYX1Y7G4vEBLCEqHmoOGdU8hV8-mSQTSDngAF5ORD5kyCt4imzSsF0jy+atloKGsLWSUUuKubyQs59rllLz+jcrAjrrymuZ7AMpDI6QB3eCtRTYPC9eBYGkHVoAcmeRNddwAdNTafNAagAshMOwAPIcO1BhnjewPR3O3Juj1UL3esKEqh+4HiMEbOZucIyk0AMgG8AAJBWTlcked9UjbvAy7CDsQ0TJrZiQDBUFAIH00O1UDgLagrENxqNGTFmfLEkbUnTyDATdsBlWpgLzgXDlWlutztDiDd4NPxs2NxWt24d5CrPuZUeZfc9nCOzIAF4EKs1xFmYh61OJE0RudEBCDDBCAISclAIUBEElfp4Fgh0dUmSE+nQ+ADxWbC-1qPpXVdSYNTwm4bjRDBbA4BAABUQCMbM6VEAAxNBMFwfBeQJF5iXeMkfkpAEgRY8RqNonBEKIVD4JARAAG1XVQ10AF1UXRCQIKMeAAEE+gYpjRPmdj0CjSVUGQCAIGsZQQg-eAAHpHPgAA9bZwPtKCKklXiiTeBgmHJX5eH4YzRFaGTGTkxA7jNJyXPcoA

dummdidumm: Good find, I removed it from hover and completion when already imported

Issue #2384, opened 2024-06-02
Cannot find name 'variable' when using a default value in a snippet


Keagel:
### Describe the bug

A warning appears on variables assigned a default value within a snippet, indicating that the name cannot be found. The warning disappears if the default value is removed. The content of the default value doesn't change the warning, ie. `''` and `'foo'` all give the same warning - it doesn't come from it being an empty string.

The error is independent from the type of the variable - the warning remains even if the type is removed.

### Reproduction

Using [this repo](https://stackblitz.com/~/github.com/Keagel/svelte5-snippet-bug?file=src/routes/TextField.svelte), install the `Svelte for VS Code` extension and enable the Typescript plugin when prompted.

### Expected behaviour

There should be no warning.

### System Info

- IDE: VSCode
- Extension: Svelte for VS Code v108.5.0

### Which package is the issue about?

Svelte for VS Code extension

dummdidumm:
Fixed in Svelte itself, update to the latest version and it should start to work

PR #2385, opened 2024-06-03
fix: support default parameter in snippet block

jasonlyu123:
#2384

### DIFF
--- a/packages/svelte2tsx/src/htmlxtojsx_v2/nodes/SnippetBlock.ts
+++ b/packages/svelte2tsx/src/htmlxtojsx_v2/nodes/SnippetBlock.ts
@@ -93,8 +93,11 @@ export function handleSnippet(
         ];
 
         if (snippetBlock.parameters?.length) {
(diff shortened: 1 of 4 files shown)


jasonlyu123:
Still need to find a way to infer the type from the default value.

dummdidumm:
What do you mean by "infer the type from the default value"?
Also, this feels like something we could fix in the Svelte parser instead - feels wrong that start/end isn't set for it.

jasonlyu123:
> What do you mean by "infer the type from the default value"?

Because of the type annotation, the type for the parameter would be any instead of inferring from the default value like functions usually do. 

> This feels like something we could fix in the Svelte parser instead. 
Sure. We can deal with this type of problem later. I can't really think of a way to let ts infer the type now.

dummdidumm:
I pushed a fix to Svelte to make the default parameter appear correctly. What's left is the default type inference problem.

dummdidumm:
Closing in favor of #2449

Issue #2397, opened 2024-06-12
Drop `fast-glob` dependency from `svelte-check`


benmccann:
### Description

Removing it would remove 18 dependencies: https://npmgraph.js.org/?q=fast-glob

It would still be pulled in via chokidar in the short-term, but chokidar 4 will remove it.

`svelte-check` pulling in `fast-glob` is responsible for 16% of the dependencies in a new SvelteKit project

### Proposed solution

The usage looks pretty simple and I don't think we need a library to help

### Alternatives

Use `tiny-glob`, which SvelteKit already uses. That won't add dependencies since the user is already downloading `tiny-glob`. Or maybe switch to `fdir`

jasonlyu123:
It actually should be a devDependencies because it's bundled by rollup. But It can also be removed as part of https://github.com/sveltejs/language-tools/issues/2364 since fast-glob doesn't handle it.

jasonlyu123:
Found one problem. Although we didn't mention the `--ignore` option can be a glob in the doc. But because it is directly passed to fast-glob, it can be. 

https://github.com/sveltejs/language-tools/blob/b49270391d22d6d1cd12f6a473d4b3b4cc6f3787/packages/svelte-check/src/index.ts#L35

Not sure if we consider this as a breaking change. `tiny-glob` also doesn't have options for ignore so we can't replace it with it. Or we can replace it with typescript's `ts.sys.readDirectory`. it only supports limited glob patterns but I doubt anyone would use complex glob patterns here.

PR #2400, opened 2024-06-12
chore: move fast-glob to devDependencies

benmccann:
ref https://github.com/sveltejs/language-tools/issues/2397

feel free to close that issue when this is merge. I didn't automatically link it since multiple solutions were being proposed there


jasonlyu123 (approved):
Yeah. This sounds like the best solution for now. Removing it from the bundle doesn't feel like a big gain.

dummdidumm (commented):
It was always bundled, we just needlessly had it as a dependency since the beginning of time ... oops

Issue #2391, opened 2024-06-08
Remove dependency on `svelte-preprocess`


benmccann:
### Describe the bug

This dependency is unnecessary. If the user doesn't have it installed in their project, then they aren't using it and there's no reason for `svelte-check` to add it. This is one of multiple issues contributing to https://github.com/sveltejs/kit/issues/12258. But even without that issue, the dependency is just adding extra weight we don't need in most projects. Even most TypeScript-enabled SvelteKit projects don't use it as they generally use vite preprocess from v-p-s

### Reproduction

https://github.com/sveltejs/language-tools/blob/c18615292110d7a945547a71609c4af03f971484/packages/svelte-check/package.json#L32

### Expected behaviour

Try to dynamically import it instead

### System Info

N/A

### Which package is the issue about?

svelte-check

jasonlyu123:
svelte-preprocess is used as a fallback when there isn't a preprocess config or config can't be loaded so I think this would be a breaking change.

benmccann:
> svelte-preprocess is used as a fallback when there isn't a preprocess config or config can't be loaded so I think this would be a breaking change

Yeah, potentially. Might not be bad to do with the next major whenever that is

> One example is that the preprocess config is directly passed into the svelte plugin in the vite.config.js/rollup.config.js instead of in the svelte.config.js

I've never seen anyone do that for TypeScript :laughing: `@sveltejs/enhanced-img` and I think a couple custom user preprocessors might do that, but the fallback isn't going to help there

dummdidumm:
I'm wondering why we're having this dependency at all and are not bundling with the other packages. Was it because of some transitive dependency not playing ball? Can't remember.

dummdidumm:
We can't really remove it because it's a dependency of `svelte-language-server`, which is also powering the VS Code extension.

benmccann:
I just meant remove it from `svelte-check`

jasonlyu123:
> I'm wondering why we're having this dependency at all and are not bundling with the other packages.

It might be because of the dynamic import in `svelte-preprocess`. 

`svelte-language-sever` is bundled by rollup so we should be able to remove it but we'll need to adjust the `useFallbackPreprocessor` to handle the module not found error. It might also make sense to add a step to try loading the vite preprocess as the fallback.

dummdidumm:
[...]

Issue #2243, opened 2024-01-02
svelte-check breaking changes v4


dummdidumm:
Loose collection of things we should do for v4:

- make TypeScript a peer dependency
- Possibly chokidar v4 is out by then and we should check whether or not to switch to it. By the looks of it it will be ESM only, so we need change our output to be ESM, too. But that sounds like a good thing to do either way; and likely harmless since svelte-check is meant to be used as a command line tool where that doesn't matter.
- Check which minimum Node version to support. Probably 18

Issues to take a closer look at:

- #2154
- #1976 
- #1234
- #1074

Timing:

- after we investigated Volar? no
- with Svelte 5? no, shorty before it

dummdidumm:
- Possibly chokidar v4 is out by then and we should check whether or not to switch to it. By the looks of it it will be ESM only, so we need change our output to be ESM, too. But that sounds like a good thing to do either way; and likely harmless since `svelte-check` is meant to be used as a command line tool where that doesn't matter.
- Checks which minimum Node version to support. Probably 18

dummdidumm:
#2420

dummdidumm:
Idea for `svelte-preprocess`:
- keep it in language-server, but load it through `await` with try-catch
- make it external in rollup build config in svelte-check, which means the fallback doesn't kick in when using svelte-check

PR #2401, opened 2024-06-12
chore: upgrade svelte-preprocess to v6

benmccann:
this will fix the deprecation warnings that users are getting


dummdidumm:
preprocess has node 18 and Svelte 4 minimum, svelte-check has 16 and 3 - so this would be a major release. Need to check what else we want to do for svelte-check 4.0 (#2243) - @jasonlyu123 anything that springs to mind we want to change?

benmccann:
I had just filed https://github.com/sveltejs/language-tools/issues/2391, which Jason suggested would be a breaking change

benmccann:
@dummdidumm would you be able to give a look at https://github.com/sveltejs/svelte-preprocess/pull/645 before this PR? Then I could update this one to use it

dummdidumm:
Closing in favor of #2452

PR #2409, opened 2024-06-21
Avoid warning issues with peer dependencies

delight:
Allow  svelte-preprocess ^6 to avoid warnings when using the latest dependencies.

```
 WARN  Issues with peer dependencies found
.
└─┬ svelte-check 3.8.1
  └─┬ svelte-preprocess 5.1.4
    └── ✕ unmet peer postcss-load-config@"^2.1.0 || ^3.0.0 || ^4.0.0 || ^5.0.0": found 6.0.1
```

### DIFF
--- a/packages/svelte-check/package.json
+++ b/packages/svelte-check/package.json
@@ -29,7 +29,7 @@
         "import-fresh": "^3.2.1",
         "picocolors": "^1.0.0",
         "sade": "^1.7.4",
-        "svelte-preprocess": "^5.1.3",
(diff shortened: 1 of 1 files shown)


dummdidumm (approved):

dummdidumm (changes requested):
We can't do this because it's a dependency, and that dependency requires node 18, while we claim node 16 still. But we'll do a new major of svelte-check soon where we can include this.

dummdidumm:
Closing in favor of #2401

PR #2412, opened 2024-06-22
fix: snippets with TS incorrect transformation

paoloricciuti:
After https://github.com/sveltejs/svelte/pull/12070 the AST for the SnippetBlock has changed (to be more correct).

The problem is that `svelte2tsx` was using that wrong information for the transformation. This means that currently last version of svelte produces wrong ts code from `svelte2tsx`.

The problem is that before parameter `end` was including the typeAnnotation
image
while now it stops where it should (at the parameter)
image

this PR fixes this problem by looking at the end of typeAnnotation if there's one and falling back to the end of the parameter otherwise.

I tried to update a test but despite messing with the expected to make it fail it was still passing so i think i need a bit of guidance over this (sorry).

~Also currently there's one failing test for the language server (in the SemanticTokenProvider so i have to check what's that.~

~Ok by debugging the language server it seems like the type semantic tokens are lost in the snippet...will try to fix~

Lol scratch last one it was failing in master too, i just needed a `pnpm i`

### DIFF
--- a/packages/svelte2tsx/src/htmlxtojsx_v2/nodes/SnippetBlock.ts
+++ b/packages/svelte2tsx/src/htmlxtojsx_v2/nodes/SnippetBlock.ts
@@ -47,7 +47,7 @@ export function handleSnippet(
     const startEnd =
         str.original.indexOf(
             '}',
-            lastParameter?.typeAnnotation.end ?? lastParameter?.end ?? snippetBlock.expression.end
+            lastParameter?.typeAnnotation?.end ?? lastParameter?.end ?? snippetBlock.expression.end
         ) + 1;
 
     if (isImplicitProp) {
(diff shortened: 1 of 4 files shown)


paoloricciuti:
Ok this should be ready...however ci is failing here and not locally for something seemingly unrelated. Maybe it's a flaky?

jasonlyu123:
Yes, it's a flaky test. I saw it once or twice locally so haven't figured out what the problem is 😅.

packages/svelte2tsx/src/htmlxtojsx_v2/nodes/SnippetBlock.ts
```diff
@@ -42,19 +42,20 @@ export function handleSnippet(
...
     );
 
+    const lastParameter = snippetBlock.parameters?.at(-1);
+
     const startEnd =
         str.original.indexOf(
             '}',
-            // context was the first iteration in a .next release, remove at some point
-            snippetBlock.parameters?.at(-1)?.end || snippetBlock.expression.end
+            lastParameter?.typeAnnotation.end ?? lastParameter?.end ?? snippetBlock.expression.end
```
dummdidumm, line 50:
```suggestion
            lastParameter?.typeAnnotation?.end ?? lastParameter?.end ?? snippetBlock.expression.end
```

PR #2420, opened 2024-06-26
chore: remove undocumented filePathsToIgnore option

benmccann:
ref https://github.com/sveltejs/language-tools/issues/2397

I did a search on GitHub and can't find anyone using this. If no one complains about it that will free us up to swap `fast-glob` with a lighter library. For now I left `fast-glob` in place, but we can follow up to remove it later on assuming no one misses this option

### DIFF
--- a/packages/svelte-check/src/index.ts
+++ b/packages/svelte-check/src/index.ts
@@ -26,6 +26,7 @@ type Result = {
 };
 
 async function openAllDocuments(workspaceUri: URI, svelteCheck: SvelteCheck) {
+    // TODO: replace glob with smaller implementation
     const files = await glob('**/*.svelte', {
         cwd: workspaceUri.fsPath,
         ignore: ['node_modules/**']
--- a/packages/svelte-check/src/index.ts
+++ b/packages/svelte-check/src/index.ts
@@ -25,14 +25,10 @@ type Result = {
     fileCountWithProblems: number;
 };
 
-async function openAllDocuments(
-    workspaceUri: URI,
-    filePathsToIgnore: string[],
-    svelteCheck: SvelteCheck
-) {
+async function openAllDocuments(workspaceUri: URI, svelteCheck: SvelteCheck) {
     const files = await glob('**/*.svelte', {
         cwd: workspaceUri.fsPath,
-        ignore: ['node_modules/**'].concat(filePathsToIgnore.map((ignore) => `${ignore}/**`))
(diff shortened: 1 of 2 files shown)


dummdidumm:
@jasonlyu123 do you remember why we ever added that? This probably predated the "load the tsconfig" times. I would be ok with removing it in the next svelte-check major.

benmccann:
When I searched all of GitHub for `filePathsToIgnore` I saw several other projects such as Astro with the same code. I'm guessing it was probably in some example that got copied

dummdidumm:
I think Astro did copy svelte-check for their initial iteration, probably just an oversight to not remove it - sounds like it's safe to remove in the next major then

jasonlyu123:
`filePathsToIgnore` is the internal name for the `--ignore` flag. Yeah. it predated the "load the tsconfig". We could probably just remove the ability for it to be a glob pattern and keep the option. The undocumented part should be that it can be a glob pattern.

packages/svelte-check/src/index.ts
```diff
@@ -25,14 +25,10 @@ type Result = {
     fileCountWithProblems: number;
 };
 
-async function openAllDocuments(
-    workspaceUri: URI,
-    filePathsToIgnore: string[],
-    svelteCheck: SvelteCheck
-) {
+async function openAllDocuments(workspaceUri: URI, svelteCheck: SvelteCheck) {
     const files = await glob('**/*.svelte', {
```
benmccann, line 29:
```suggestion
    // TODO: replace glob with smaller implementation
    const files = await glob('**/*.svelte', {
```

dummdidumm:
Did a quick github search, only one archived repository used a glob pattern. A bunch of others use it but without any patterns. @benmccann I'm therefore closing this PR, but feel free to open a new one that replaces the glob library with a simpler one that also fits our needs, as long as said library can deal with simple patterns (and we need to keep some form of it anyway because of our built-in ignores)

benmccann:
@dummdidumm could you share some example repos? I had done a search and couldn't find any consumers, so I'm not quite sure what patterns need to be supported

dummdidumm:
This is the search I did: https://github.com/search?q=%22svelte-check+--ignore%22+language%3Ajson&type=code

benmccann:
ohhh. I see. I misunderstood this code and searched for `filePathsToIgnore`. thanks!

Issue #2419, opened 2024-06-25
Latest release causing types and intellisense as well as variable usages to not function correctly.


Rotinx:
### Describe the bug

While using the extension types as well as intellisense and variable usages don't function so your unable to determine that a import or variable has been used or types not loading.

### Reproduction

Update to the latest release utilize in a component as you normally would. 

### Expected behaviour

Types intellisense and variable usages accurately display.

### System Info

Windows
VS Code
Extension version v108.5.2

Svelte ^4.2.17

### Which package is the issue about?

Svelte for VS Code extension

### Additional Information, eg. Screenshots

Downgrading one version solved the issue.

Rotinx:
Downgrading can be done here. Just a note for people running into this issue also.
![image](https://github.com/sveltejs/language-tools/assets/39930952/3ecdfc94-7ec9-4fa1-a657-affc430e6a36)

jasonlyu123:
Does this happen to all the files or on a specific file? If it's a particular file, Can you recall if you have renamed the casing of a file? For example, `Test.svelte`  to `test.svelte`.

kjk:
Happened to me too, on two different projects.

Can even repro by creating the simplest svelte javascript project with `bun create vite` (pick `svelte`, then `javascript`) .

Then open in VSCode, make an error somewhere and save the file. You should get an error immediately but you don't.

This is on latest Windows 11.

Downgrading to `108.5.1` fixed the problem.

https://github.com/sveltejs/language-tools/commit/dbfb47acde50e2b2c9bbcd864c3ffeadb06ceb5d could be the reason (just guessing, if file watching is broken then lsp won't try to re-check the files).

kjk:
Since both me and @Rotinx are on Windows, this could be Windows only: paths and file watching is something that could break on Windows but not on Mac or Linux.

jasonlyu123:
Please provide a reproduction that isn't bun.

Rotinx:
> Does this happen to all the files or on a specific file? If it's a particular file, Can you recall if you have renamed the casing of a file? For example, `Test.svelte` to `test.svelte`.

It wasn't specific to only one file that's for sure. While I was developing before downgrading I was mainly in components so I cannot tell you if it also effected pages for sure. All generally components and pages are in lowercase and components with dashes test-component.svelte etc that I was working with.

I tried replicating it a few minutes ago but no luck so far.  I'll give it another go tomorrow and see if I can find a pattern.

[...]

PR #2421, opened 2024-06-26
fix: handle issues with svelte file watching 

jasonlyu123:


jasonlyu123:
Yeah. This should fix #2419. The trigger is file save, once the file is saved it'll be marked as closed so further diagnostic is skipped.

dummdidumm:
From my testing the issue was also the missing `pathToUrl` in `onWatchFileChanges`. It wrongfully marked the document as not being opened because no document was found

Issue #2371, opened 2024-05-12
`.d.ts` files missing silently after emit


jeannemas:
### Describe the bug

Related issues:

- [emitDts should print diagnostic messages](https://github.com/sveltejs/language-tools/issues/1835)

Hey! I'd like to maybe bring back the above issue to life, since I believe to have found a consistent way to reproduce it while keeping `svelte-check` saying everything is fine, which results in missing `.d.ts` files for some components.

I originally stumbled upon this while working with [Shadcn-svelte](https://www.shadcn-svelte.com/) and [vaul-svelte](https://www.vaul-svelte.com/) but I believe it affects a lot of other third-party libraries.

## TL;DR

The issue is that the `.d.ts` files are not being generated for some components, while `svelte-check` says everything is fine, which means that the issue is not being caught by the linter.

## My current solution (Gist link)

In order to find the issues (which basically needs to print the `.emit()` output), I needed to create a modified version of the package script which I've assembled into a Gist.

[Gist link `scripts/package.js`](https://gist.github.com/jeannemas/c23093c0979dcec516e61ddb147308ce)

All of this to say that I believe it is a good idea to handle the emit result and log it out in case of an error, since it might be the only way to find out what's wrong with the `.d.ts` files generation.

## Quick repro using Vaul-Svelte (example)

- Clone the repo
  ```zsh
  git clone https://github.com/huntabyte/vaul-svelte.git
  ```
- Install the dependencies
  ```zsh
  pnpm i
  ```
- Package the library
  ```zsh
  pnpm package
  ```
- Notice the output saying everything is fine
  ```raw
  src/lib -> dist
  vaul-svelte lint results:
  All good!
  ```
- List the emmitted files inside `dist/vaul/components`
  ```zsh
  ls -1 dist/vaul/components
  ```
- Notice the missing `.d.ts` files for some components (`close.svelte`, `content.svelte`, `trigger.svelte`)
  ```raw
  close-wrapper.svelte
  close-wrapper.svelte.d.ts
  close.svelte
  content.svelte
  index.d.ts
  index.js
  nested-root.svelte
  nested-root.svelte.d.ts
  overlay.svelte
  overlay.svelte.d.ts
  root.svelte
  root.svelte.d.ts
  trigger-wrapper.svelte
  trigger-wrapper.svelte.d.ts
  trigger.svelte
  types.d.ts
  types.js
  visible.svelte
  visible.svelte.d.ts
  ```
- Check using `svelte-check`
[...]

PR #2428, opened 2024-07-02
feat: warn on likely failed d.ts file generation

dummdidumm:
emitDts now warns when it detects diagnostics that likely mean the `d.ts` file was not generated

### DIFF
--- a/packages/svelte2tsx/src/emitDts.ts
+++ b/packages/svelte2tsx/src/emitDts.ts
@@ -13,7 +13,34 @@ export async function emitDts(config: EmitDtsConfig) {
     const { options, filenames } = loadTsconfig(config, svelteMap);
     const host = await createTsCompilerHost(options, svelteMap);
     const program = ts.createProgram(filenames, options, host);
-    program.emit();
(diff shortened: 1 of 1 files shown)


jasonlyu123:
The code for #2371 is `2742` but judging from https://github.com/microsoft/TypeScript/blob/main/src/compiler/transformers/declarations.ts there are 10+ more exceptions outside the 4000-4108 range. Do you think we should write a script to fetch it from this file or maybe try something like https://github.com/sveltejs/language-tools/issues/2371#issuecomment-2128839670?

dummdidumm:
Sounds good - let's try that! As for the diagnostic: TBH even for me, and even if the code wasn't transpiled, it's super confusing to really know what I should do after getting such an error message. So I wouldn't invest _too_ much time into making them nicer.

SystemParadox:
I literally ran into this last week and this patch would have saved a lot of time! I agree that just knowing that it failed (and which file) is the main thing. I think it's highly like that in practice you'll have to open it up in vscode to be able to work out what's going on anyway.

PR #2433, opened 2024-07-04
chore: switch from fast-glob to fdir

### LINKED ISSUE #2364: VSCode Plugin does not handle symlinks correctly
### Describe the bug

The plugin won't initialize when the project folder contains two or more symbolic links that point to the root of the project and as a result no svelte related features like autocompletion work.

This happens for example when using svelte in a [tauri](https://github.com/tauri-apps/tauri) project together with the [cxx](https://github.com/dtolnay/cxx) library.
When the project is compiled, cxx creates multiple symbolic links that point to the project root inside the target folder, as described [here](https://github.com/dtolnay/cxx/issues/1004).

### Reproduction

The issue can be reproduced with a newly created svelte project (using the default options):
```
npm create svelte@latest my-app
cd my-app
npm install
code .
```
Ensure svelte vs code plugin is on, then open e.g. Counter.svelte and try using autocompletion on "display_counter." and notice how it still works as expected.

Svelte plugin output at this point:
```
Initialize language server at  file:///r%3A/my-app
Initialize new ts service at  r:/my-app/jsconfig.json
Trying to load configs for r:/my-app
Loaded config at  r:\my-app\svelte.config.js
SnapshotManager File Statistics:
Project files: 22
Svelte files: 7
From node_modules: 0
Total: 22
Initialize new ts service at  
Trying to load configs for r:\my-app
```

Next create two symbolic link to the project root using powershell for example:
```
New-Item -Path test1 -ItemType SymbolicLink -Value .
New-Item -Path test2 -ItemType SymbolicLink -Value .
```
Afterwards restart the extension host in vscode and try autocompletion again.
It no longer works and only shows "Loading..." forever. The Svelte plugin output looks like this:
```
Initialize language server at  file:///r%3A/my-app
Initialize new ts service at  r:/my-app/jsconfig.json
Trying to load configs for r:/my-app
```

After the symbolic links are removed and the extension host is restarted, it works again.

### Expected behaviour

The plugin should operate normally, regardless of any symbolic links to the project root.

### System Info

- OS: Windows 11
- IDE: VSCode

Svelte for VS Code extension

When only one symbolic link exists in the project, The autocompletion works, but the plugin contains some very interesting output, that may explain why it fails to load:

```
Initialize language server at  file:///r%3A/my-app
Initialize new ts service at  r:/my-app/jsconfig.json
Trying to load configs for r:/my-app
Loaded config at  r:\my-app\svelte.config.js
Loaded config at  r:\my-app\test1\svelte.config.js
Loaded config at  r:\my-app\test1\test1\svelte.config.js
Loaded config at  r:\my-app\test1\test1\test1\svelte.config.js
Loaded config at  r:\my-app\test1\test1\test1\test1\svelte.config.js
Loaded config at  r:\my-app\test1\test1\test1\test1\test1\svelte.config.js
Loaded config at  r:\my-app\test1\test1\test1\test1\test1\test1\svelte.config.js
Loaded config at  r:\my-app\test1\test1\test1\test1\test1\test1\test1\svelte.config.js
Loaded config at  r:\my-app\test1\test1\test1\test1\test1\test1\test1\test1\svelte.config.js
Loaded config at  r:\my-app\test1\test1\test1\test1\test1\test1\test1\test1\test1\svelte.config.js
Loaded config at  r:\my-app\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\svelte.config.js
Loaded config at  r:\my-app\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\svelte.config.js
Loaded config at  r:\my-app\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\svelte.config.js
Loaded config at  r:\my-app\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\svelte.config.js
Loaded config at  r:\my-app\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\svelte.config.js
Loaded config at  r:\my-app\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\svelte.config.js
Loaded config at  r:\my-app\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\svelte.config.js
Loaded config at  r:\my-app\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\svelte.config.js
Loaded config at  r:\my-app\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\svelte.config.js
Loaded config at  r:\my-app\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\svelte.config.js
Loaded config at  r:\my-app\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\svelte.config.js
Loaded config at  r:\my-app\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\svelte.config.js
Loaded config at  r:\my-app\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\svelte.config.js
Loaded config at  r:\my-app\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\svelte.config.js
Loaded config at  r:\my-app\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\svelte.config.js
Loaded config at  r:\my-app\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\svelte.config.js
Loaded config at  r:\my-app\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\svelte.config.js
Loaded config at  r:\my-app\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\svelte.config.js
Loaded config at  r:\my-app\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\svelte.config.js
Loaded config at  r:\my-app\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\svelte.config.js
Loaded config at  r:\my-app\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\svelte.config.js
Loaded config at  r:\my-app\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\test1\svelte.config.js
SnapshotManager File Statistics:
Project files: 22
Svelte files: 7
From node_modules: 0
Total: 22
Initialize new ts service at  
Trying to load configs for r:\my-app
```

It seems that it looks for svelte.config.js files inside any folder it can find and runs into some infinite loop when there are multiple symbolic links.

### LINKED ISSUE #2397: Drop `fast-glob` dependency from `svelte-check`

benmccann:
closes #2397
fixes #2364

fast-glob:
https://npmgraph.js.org/?q=fast-glob
https://bundlephobia.com/package/fast-glob@3.3.2

fdir:
https://npmgraph.js.org/?q=fdir
https://bundlephobia.com/package/fdir@6.1.1

docs: https://github.com/thecodrr/fdir/blob/master/documentation.md

### DIFF
--- a/packages/language-server/src/lib/documents/configLoader.ts
+++ b/packages/language-server/src/lib/documents/configLoader.ts
@@ -4,7 +4,7 @@ import { CompileOptions } from 'svelte/types/compiler/interfaces';
 // @ts-ignore
 import { PreprocessorGroup } from 'svelte/types/compiler/preprocess';
 import { importSveltePreprocess } from '../../importPackage';
-import _glob from 'fast-glob';
+import { fdir } from 'fdir';
 import _path from 'path';
 import _fs from 'fs';
 import { pathToFileURL, URL } from 'url';
@@ -46,6 +46,8 @@ const _dynamicImport = new Function('modulePath', 'return import(modulePath)') a
     modulePath: URL
 ) => Promise<any>;
 
+const configRegex = /\/svelte\.config\.(js|cjs|mjs)$/;
+
 /**
  * Loads svelte.config.{js,cjs,mjs} files. Provides both a synchronous and asynchronous
  * interface to get a config file because snapshots need access to it synchronously.
@@ -60,7 +62,7 @@ export class ConfigLoader {
     private disabled = false;
 
     constructor(
-        private globSync: typeof _glob.sync,
+        private globSync: typeof fdir,
         private fs: Pick<typeof _fs, 'existsSync'>,
         private path: Pick<typeof _path, 'dirname' | 'relative' | 'join'>,
         private dynamicImport: typeof _dynamicImport
@@ -83,12 +85,19 @@ export class ConfigLoader {
         Logger.log('Trying to load configs for', directory);
 
         try {
-            const pathResults = this.globSync('**/svelte.config.{js,cjs,mjs}', {
-                cwd: directory,
-                // the second pattern is necessary because else fast-glob treats .tmp/../node_modules/.. as a valid match for some reason
-                ignore: ['**/node_modules/**', '**/.*/**'],
-                onlyFiles: true
-            });
+            const pathResults = new this.globSync({})
+                .withPathSeparator('/')
+                .exclude((_, path) => {
+                    // no / at the start, path could start with node_modules
+                    return path.includes('node_modules/') || path.includes('/.');
+                })
+                .filter((path, isDir) => {
+                    return !isDir && configRegex.test(path);
+                })
+                .withRelativePaths()
+                .crawl(directory)
+                .sync();
+
             const someConfigIsImmediateFileInDirectory =
                 pathResults.length > 0 && pathResults.some((res) => !this.path.dirname(res));
             if (!someConfigIsImmediateFileInDirectory) {
@@ -270,4 +279,4 @@ export class ConfigLoader {
     }
 }
 
-export const configLoader = new ConfigLoader(_glob.sync, _fs, _path, _dynamicImport);
+export const configLoader = new ConfigLoader(fdir, _fs, _path, _dynamicImport);
--- a/packages/svelte-check/src/index.ts
+++ b/packages/svelte-check/src/index.ts
@@ -49,7 +49,7 @@ async function openAllDocuments(
         return (path) => path.startsWith(i);
     });
     const isIngored = (path: string) => {
-        path = path.slice(offset).replace(/\\/g, '/');
+        path = path.slice(offset);
         for (const i of ignored) {
             if (i(path)) {
                 return true;
@@ -63,6 +63,7 @@ async function openAllDocuments(
             path = path.slice(offset);
             return path.startsWith('.') || path.startsWith('node_modules');
         })
+        .withPathSeparator('/')
         .withFullPaths()
         .crawl(workspaceUri.fsPath)
         .withPromise();
--- a/packages/svelte-check/src/index.ts
+++ b/packages/svelte-check/src/index.ts
@@ -53,7 +53,7 @@ async function openAllDocuments(
 
         return (path) => path.startsWith(i);
     });
-    const isIngored = (path: string) => {
+    const isIgnored = (path: string) => {
         path = path.slice(offset);
         for (const i of ignored) {
             if (i(path)) {
@@ -63,7 +63,7 @@ async function openAllDocuments(
         return false;
     };
     const absFilePaths = await new fdir()
-        .filter((path) => path.endsWith('.svelte') && !isIngored(path))
+        .filter((path) => path.endsWith('.svelte') && !isIgnored(path))
         .exclude((_, path) => {
             path = path.slice(offset);
             return path.startsWith('.') || path.startsWith('node_modules');
--- a/packages/svelte-check/src/index.ts
+++ b/packages/svelte-check/src/index.ts
@@ -4,7 +4,7 @@
 
 import { watch } from 'chokidar';
 import * as fs from 'fs';
-import glob from 'fast-glob';
+import { fdir } from 'fdir';
 import * as path from 'path';
 import { SvelteCheck, SvelteCheckOptions } from 'svelte-language-server';
 import { Diagnostic, DiagnosticSeverity } from 'vscode-languageserver-protocol';
@@ -30,10 +30,20 @@ async function openAllDocuments(
     filePathsToIgnore: string[],
     svelteCheck: SvelteCheck
 ) {
-    const files = await glob('**/*.svelte', {
-        cwd: workspaceUri.fsPath,
-        ignore: ['node_modules/**'].concat(filePathsToIgnore.map((ignore) => `${ignore}/**`))
-    });
+    const ignored = ['node_modules'].concat(filePathsToIgnore);
+    const isIngored = (path: string) => {
+        for (const i of ignored) {
+            if (path.startsWith(i)) {
+                return true;
+            }
+        }
+        return false;
+    };
+    const files = await new fdir()
+        .withBasePath()
+        .filter((path) => path.endsWith('.svelte') && !isIngored(path))
+        .crawl(workspaceUri.fsPath)
+        .withPromise();
     const absFilePaths = files.map((f) => path.resolve(workspaceUri.fsPath, f));
 
     for (const absFilePath of absFilePaths) {
--- a/packages/svelte-check/src/index.ts
+++ b/packages/svelte-check/src/index.ts
@@ -31,21 +31,26 @@ async function openAllDocuments(
     svelteCheck: SvelteCheck
 ) {
     const offset = workspaceUri.fsPath.length + 1;
-    // We support a very limited subset of glob patterns: You can have a * at the end or the start
+    // We support a very limited subset of glob patterns: You can only have  ** at the end or the start
     const ignored: Array<(path: string) => boolean> = filePathsToIgnore.map((i) => {
-        if (i.endsWith('*')) i = i.slice(0, -1);
-        if (i.startsWith('*')) {
-            i = i.slice(1);
+        if (i.endsWith('**')) i = i.slice(0, -2);
+
+        if (i.startsWith('**')) {
+            i = i.slice(2);
+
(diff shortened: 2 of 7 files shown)


dummdidumm:
Tweaked a few things and switched to fdir in `svelte-language-server` aswell - @jasonlyu123 anything that speaks against that?

packages/svelte-check/src/index.ts
```diff
@@ -30,11 +30,43 @@ async function openAllDocuments(
...
+            return (path) => path.includes(i);
+        }
+        if (i.includes('*'))
+            throw new Error(
+                'Invalid svelte-check --ignore pattern: Only * at the start or end is supported'
+            );
+        return (path) => path.startsWith(i);
     });
-    const absFilePaths = files.map((f) => path.resolve(workspaceUri.fsPath, f));
+    const isIngored = (path: string) => {
```
benmccann, line 51:
```suggestion
    const isIgnored = (path: string) => {
```

packages/svelte-check/src/index.ts
```diff
@@ -30,11 +30,43 @@ async function openAllDocuments(
     filePathsToIgnore: string[],
     svelteCheck: SvelteCheck
 ) {
-    const files = await glob('**/*.svelte', {
-        cwd: workspaceUri.fsPath,
-        ignore: ['node_modules/**'].concat(filePathsToIgnore.map((ignore) => `${ignore}/**`))
+    const offset = workspaceUri.fsPath.length + 1;
+    // We support a very limited subset of glob patterns: You can have a * at the end or the start
```
benmccann, line 34: why?

benmccann:
Glad you were able to figure out how to call `fdir` directly as it can be tricky to figure out the combination of options needed! Since I created this PR there's been a new `tinyglobby` package created, which is a drop-in replacement for `fast-glob` that has only `fdir` and `picomatch` as dependencies, so if it's helpful you could also take a look at that. I didn't quite follow what some of the new code was doing...

packages/svelte-check/src/index.ts
dummdidumm: What do you mean by "why"?

dummdidumm:
think going with `fdir` is fine, we don't really need the glob syntax support (and I'm fine with only supporting a very limited glob syntax subset for the `--ignore` option)

packages/svelte-check/src/index.ts
benmccann: ah, I misunderstood what this was doing

packages/svelte-check/src/index.ts
```diff
@@ -30,11 +30,43 @@ async function openAllDocuments(
...
+    const ignored: Array<(path: string) => boolean> = filePathsToIgnore.map((i) => {
+        if (i.endsWith('*')) i = i.slice(0, -1);
```
benmccann, line 36: I think this should be `**`. because the `path.includes` call below won't care about path separators

packages/svelte-check/src/index.ts
```diff
@@ -30,11 +30,43 @@ async function openAllDocuments(
...
+        if (i.startsWith('*')) {
+            i = i.slice(1);
+            if (i.includes('*'))
+                throw new Error(
+                    'Invalid svelte-check --ignore pattern: Only * at the start or end is supported'
+                );
+            return (path) => path.includes(i);
```
benmccann, line 43: I think this needs to be posixified. If someone checks in something like `"check": "svelte-check --ignore packages/fixtures/**"` and then their Windows teammate runs it I think this could break

jasonlyu123:
fast-glob default to follow symlinks but `fdir` didn't. Do we want to support it? The default is to read the symlinked file but not walk down the symlinked directory. After writing down this I think this is good enough. Although it might be a minor breaking change but I doubt there are many people does this. In this case, this also fixes #2364

dummdidumm:
Interesting! I think the behavior of not following symlinks is better, and instead walk the tree like the user sees it visually. In that sense I would acutally count this as a nice side-effect bugfix.

packages/svelte-check/src/index.ts
dummdidumm: this is handled by the `withPathSeparator('/')`, which does the posification for us

packages/svelte-check/src/index.ts
dummdidumm: what do you mean by that? Can you give an example where it would break / be wrong?

benmccann: `*` vs `**` have different meaning in glob syntax where `*` excludes path separators and `**` includes them. See https://github.com/micromatch/picomatch?tab=readme-ov-file#basic-globbing

`path.includes` will not stop at path separators like `*` will

So if you want to ignore all of `node_modules` the glob syntax `node_modules/*` should not work whereas `node_modules/**` should

dummdidumm: ah ok so this is a case of me not knowing the globbing syntax well enough 😄

packages/language-server/src/lib/documents/configLoader.ts
```diff
@@ -83,12 +85,19 @@ export class ConfigLoader {
...
-                cwd: directory,
-                // the second pattern is necessary because else fast-glob treats .tmp/../node_modules/.. as a valid match for some reason
-                ignore: ['**/node_modules/**', '**/.*/**'],
-                onlyFiles: true
-            });
+            const pathResults = new this.globSync({})
+                .withPathSeparator('/')
+                .exclude((_, path) => {
+                    // no / at the start, path could start with node_modules
+                    return path.includes('node_modules/') || path.includes('/.');
```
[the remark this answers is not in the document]
dummdidumm: It's necessary to not traverse into directories such as `.git` etc (directories starting with a dot are generally considered to be private), and I can't think of a situation where you would have a `svelte.config.js` in one of those directories - do you have such a case or why do you ask=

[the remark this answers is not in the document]
brunnerh: I have used `.` directories for caching/generated content but don't think they needed checking/processing so far.
Was just wondering if this could be simplified, but it's probably also faster this way.

Issue #2442, opened 2024-07-18
Declaration maps have incorrect `"sources"` paths for `.svelte.ts|js` files


ryanatkn:
### Describe the bug

The generated `.d.ts.map` files from `svelte-package` have incorrect `"sources"` paths for `.svelte.ts` and `.svelte.js` files. The problem appears to be [a false-positive detection](https://github.com/sveltejs/language-tools/blob/027ab23e3e59939ad1f018a34ca5e0386cbaf829/packages/svelte2tsx/src/emitDts.ts#L176) that maps the path for real Svelte files. I think I have the right fix in #2443.

### Reproduction

Repo with reproduction and instructions - https://github.com/ryanatkn/svelte-declaration-map-repro

From a new `npm create svelte@latest` library with TypeScript:

1 - Add to tsconfig:

```json
"declaration": true,
"declarationMap": true,
"sourceRoot": "../src/lib/" // not actually needed but makes the paths correct
```

2 - Add a `.svelte.ts` module to `src/lib`:

```ts
export class Fails {
  s = $state();
}
```

3 - Run `npm run package` and inspect `dist/fails.svelte.d.ts.map`.

### Expected behaviour

`dist/fails.svelte.d.ts.map` from `npm run package`

Actual:

```json
{"version":3,"file":"fails.svelte.d.ts","sourceRoot":"../src/lib/","sources":["fails.svelte"]
```

Expected:

```json
{"version":3,"file":"fails.svelte.d.ts","sourceRoot":"../src/lib/","sources":["fails.svelte.ts"]
```

### System Info

- OS: Linux
- IDE: VSCode

### Which package is the issue about?

svelte2tsx

PR #2443, opened 2024-07-18
fix: correctly detect real Svelte files for declarationMap

ryanatkn:
I opened #2442 - `declarationMap` `"sources"` being incorrect for `.svelte.ts` and `.svelte.js` files - and this appears to fix it.

I struggled with the tests, I'm sorry for not getting them right first. It was working in my reproduction but the paths are different in the tests because of `pathPrefix`, and moving the check after `sourcePath` is adjusted for `pathPrefix` seems to be correct both in the tests here and [my bug reproduction](https://github.com/ryanatkn/svelte-declaration-map-repro).

### DIFF
--- a/packages/svelte2tsx/src/emitDts.ts
+++ b/packages/svelte2tsx/src/emitDts.ts
@@ -173,8 +173,8 @@ async function createTsCompilerHost(options: any, svelteMap: SvelteMap) {
             if (fileName.endsWith('d.ts.map')) {
                 data = data.replace(/"sources":\["(.+?)"\]/, (_, sourcePath: string) => {
                     // Due to our hack of treating .svelte files as .ts files, we need to adjust the extension
-                    if (sourcePath.endsWith('.svelte.ts')) {
-                        sourcePath = sourcePath.slice(0, -3);
+                    if (svelteMap.get(path.join(options.rootDir, toRealSvelteFilepath(sourcePath)))) {
+                        sourcePath = toRealSvelteFilepath(sourcePath);
                     }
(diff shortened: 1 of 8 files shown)


packages/svelte2tsx/test/emitDts/samples/typescript-declarationMap/tsconfig.json
```diff
@@ -9,7 +9,8 @@
         "esModuleInterop": true,
         "allowJs": true,
         "checkJs": true,
-        "declarationMap": true
+        "declarationMap": true,
+        "declaration": true
```
ryanatkn, line 13: TypeScript complains when `declarationMap` is enabled without `declaration`, the other tsconfig with `declarationMap` enabled in language-server has it already.

packages/svelte2tsx/src/emitDts.ts
```diff
@@ -184,6 +180,10 @@ async function createTsCompilerHost(options: any, svelteMap: SvelteMap) {
                                   sourcePath.indexOf(pathPrefix) + pathPrefix.length + 1
                               )
                             : sourcePath;
+                    // Due to our hack of treating .svelte files as .ts files, we need to adjust the extension
+                    if (svelteMap.get(path.join(options.rootDir, toRealSvelteFilepath(sourcePath)))) {
+                        sourcePath = toRealSvelteFilepath(sourcePath);
+                    }
```
ryanatkn, line 186: If this code isn't moved after the `sourcePath` inversion, it fails the lookup.

dummdidumm (approved):
Thank you!

Issue #2425, opened 2024-06-27
svelte-check shows `Cannot find name '<snippet>'` error when snippet is defined outside component it's being used for


sheijne:
### Describe the bug

When running svelte-check on a file which defines a snippet "outside" an component/element, svelte-check will show the following error:

```
Error: Cannot find name 'snippet'. (ts)
	<Button children={hasChildren ? renderChildren : undefined}></Button>
```

### Reproduction

Running `svelte-check` against the following code will trigger the error:

```svelte
<script lang="ts">
  import SomeComponent from './SomeComponent.svelte';
  import Button from './Button.svelte';

  let hasChildren = $state(false);
</script>

<SomeComponent>
  {#snippet renderChildren()}Hello world!{/snippet}

  <!-- All these trigger the error -->
  <Button children={hasChildren ? renderChildren : undefined}></Button>
  <Button>{@render renderChildren()}</Button>
  <Button>{#snippet children()}{@render renderChildren()}{/snippet}</Button>
</SomeComponent>
```

Where `SomeComponent` and `Button` are arbitrary.

```svelte
<script>
  let { children } = $props();
</script>

<button>{@render children?.()}</button>
```

### Expected behaviour

[...]

Issue #2417, opened 2024-06-25
False positive snippet usage before assignment errors


brunnerh:
### Describe the bug

Snippets are available anywhere in their scope.
The language tools falsely surface errors if the order of definition and usage is reversed on the same level.

### Reproduction

```svelte
<script>
    const address = null;
</script>

<div>
	{@render (address ?? defaultAddress)()}
	{#snippet defaultAddress()}Unknown address{/snippet}
</div>
```

> Variable 'defaultAddress' is used before being assigned. js(2454)

![screenshot of error](https://github.com/sveltejs/language-tools/assets/834235/5dcf71d8-c5cc-4ddc-82f6-36fec7394cff)

### Expected behaviour

No error.

### System Info

- OS: Win 10
- IDE: VSCode
- Extension version: v108.5.2

dummdidumm:
Top level snippets are hoisted to the top, guess we need to do that for nested snippets, too (top of their scope)

jasonlyu123:
Leaving here as a record. I thought it would be easier to just transform it into a function declaration but snippets are actually block-scoped so the following won't be flagged as an error if it was transformed to a `var` or a function declaration but would result in a runtime error. This might also mean, at least nested one, `const` is more correct than the current var. 

```svelte
<div>
  {#snippet hi()}{/snippet}
</div>

{@render hi()}
```

PR #2449, opened 2024-07-30
fix: various snippet fixes

### LINKED ISSUE #2444: Should it be possible to type snippet args using JSDoc?
### Describe the bug

Snippets that are passed to components can be typed as parts of `$props` but how about snippets defined and used locally?

**Workaround:** Just use TS.

### Reproduction

```svelte
{#snippet test(/** @type {number} */ arg)}
	{arg}
{/snippet}
```
> Expected identifier or destructure pattern

### Logs

_No response_

### System Info

```shell
System:
    OS: Windows 10 10.0.19045
    CPU: (12) x64 Intel(R) Core(TM) i7-8700K CPU @ 3.70GHz
    Memory: 10.15 GB / 31.92 GB
  Binaries:
    Node: 18.19.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.19 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 9.2.0 - ~\AppData\Roaming\npm\npm.CMD
    pnpm: 8.6.9 - ~\AppData\Local\pnpm\pnpm.EXE
  Browsers:
    Chrome: 123.0.6312.124
    Edge: Spartan (), Chromium (123.0.2420.97), ChromiumDev ()
    Internet Explorer: 11.0.19041.3636
```

### Severity

annoyance

jasonlyu123:
#2444 #2417 #2425 https://github.com/sveltejs/language-tools/pull/2385#issuecomment-2152542282

The snippet is now transformed into a function with the return type so that The default parameter type can be inferred:

```ts
const foo = (a: string, b = 1): ReturnType<import('svelte').Snippet> => {

}
```
圖片

Another solution I considered using is an IIFE function so the inferred type is still `Snippet<[...]>`. But seems like typescript almost always expanded the type in hover. And

```ts
const foo = (() => {
    const $$_func = (a: string, b = 1) => { }

    return __sveltets_2_any() as Snippet<Parameters<typeof $$_func>>;
})()
```
圖片

I went for the `ReturnType` because the hover is shorter and the code generation is slightly simpler but OK with both so I could also change it to IIFE. 

### Breaking change
Since it is now possible to type snippets with JSDoc I removed explicit `any` in the generated code while changing the generation. This is a breaking change for js components with type-check but we can also add the `any` back for parameter without a default.

### Scoping issues
The `var` is changed to `const` because this result in a runtime error but Typescript give a green light if it's var.

```svelte
<div>
    {#snippet hi2()} {/snippet}
</div>

{@render hi2()}
```

The generated const is now hoist to before the first non-snippet sibling for #2417. And for implicit props, it's still passed directly to the component so the type can be inferred but another variable is generated for #2425

```ts
const {foo} = $$_tnenopmoC0.$$prop_def
```

A limitation is that we'll have to map the definition/reference/rename back but we can probably do this in a separate issue. Not even sure if many people actually rewrite something like this.


dummdidumm (approved):
Nice!

Issue #1683, opened 2022-10-15
Assume an empty svelte.config.js instead of requiring it?


ciscoheat:
### Describe the bug

I was pointed here after reporting withastro/astro#5065 , the problem seems to be that the VS Code extension requires a `svelte.config.js` that isn't always available, for example when using Sass in a Svelte component in an Astro project, that doesn't have that file in its root folder.

### Reproduction

When using a Svelte component with Sass styling (<style lang="scss">) in an Astro project, I get a warning in VS Code:

```
Cannot find any of modules: sass,node-sass

Error: Cannot find module 'node-sass'
Require stack:
- d:\program\VSCode\data\extensions\svelte.svelte-vscode-106.2.0\node_modules\svelte-preprocess\dist\modules\utils.js
- d:\program\VSCode\data\extensions\svelte.svelte-vscode-106.2.0\node_modules\svelte-preprocess\dist\autoProcess.js
- d:\program\VSCode\data\extensions\svelte.svelte-vscode-106.2.0\node_modules\svelte-preprocess\dist\index.js
- d:\program\VSCode\data\extensions\svelte.svelte-vscode-106.2.0\node_modules\svelte-language-server\dist\src\importPackage.js
- d:\program\VSCode\data\extensions\svelte.svelte-vscode-106.2.0\node_modules\svelte-language-server\dist\src\lib\documents\configLoader.js
- d:\program\VSCode\data\extensions\svelte.svelte-vscode-106.2.0\node_modules\svelte-language-server\dist\src\lib\documents\Document.js
- d:\program\VSCode\data\extensions\svelte.svelte-vscode-106.2.0\node_modules\svelte-language-server\dist\src\lib\documents\index.js
- d:\program\VSCode\data\extensions\svelte.svelte-vscode-106.2.0\node_modules\svelte-language-server\dist\src\server.js
- d:\program\VSCode\data\extensions\svelte.svelte-vscode-106.2.0\node_modules\svelte-language-server\bin\server.js

The file cannot be parsed because style requires a preprocessor that doesn't seem to be setup or failed during setup. Did you setup a `svelte.config.js`? If you use SCSS, it may be necessary to add the path to your NODE runtime to the setting `svelte.language-server.runtime`, or use `sass` instead of `node-sass`. 

See https://github.com/sveltejs/language-tools/tree/master/docs#using-with-preprocessors for more info.
```

[...]

PR #2452, opened 2024-07-31
feat: replace svelte-preprocess with barebones TS preprocessor

dummdidumm:
Removes the need for a dependency on svelte-preprocess in svelte-language-server, since that was all we used it for. Also tightens up the warning filter to account for situations where those could be wrong.

closes:
- #1683 and #1259, because the situation no longer arises (because we no longer have svelte-preprocess inside the VS Code extension)
- #2391

@jasonlyu123 any situation I'm not correctly antipicating here which would make this a bad change? The only cases I see where this could have any change in behavior compared to before is if you're opening a Svelte file with `lang="ts"` and no config and...
-  using import statements that don't adhere to `verbatimModuleSyntax`, but even then I think it will not make a difference because Svelte does not care which imports are left in the file.
- using decorators, because they are not transpiled and lead to a compiler error, but then the "filter out this error" logic should kick in, so it's fine

### DIFF
--- a/packages/language-server/package.json
+++ b/packages/language-server/package.json
@@ -55,7 +55,6 @@
         "prettier": "~3.2.5",
         "prettier-plugin-svelte": "^3.2.2",
         "svelte": "^3.57.0",
-        "svelte-preprocess": "^5.1.3",
         "svelte2tsx": "workspace:~",
         "typescript": "^5.5.2",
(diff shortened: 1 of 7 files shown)


jasonlyu123 (approved):
Looks good. I can't think of another situation either.

Issue #1074, opened 2021-06-24
[svelte-check] `--ignore` flag gets overwritten by `--tsconfig`


macmillen:
**Describe the bug**
When i use both `--ignore` and `--tsconfig` flags in the `svelte-check` cli, the `--ignore` flag doesn't do anything anymore. It only works when I omit the `--tsconfig` flag.

**To Reproduce**
``` bash
yarn svelte-check --ignore "src" --tsconfig ./tsconfig.json
```

**Expected behavior**
If that's expected throw a warning or something to make the user aware. Otherwise it should work with both flags.

dummdidumm:
It's expected, the tsconfig exclude setting should be used in that case. Throwing makes sense.

dummdidumm:
In svelte-check, setting both will be an error

PR #2453, opened 2024-07-31
breaking: svelte-check v4

### LINKED ISSUE #2243: svelte-check breaking changes v4

dummdidumm:
- breaking: make TypeScript a peer dependency
- breaking: require node 18 or later
- breaking: require Svelte 4 or later (devDependencies pinned to Svelte 3 because other packages in this repo still use it. Theoretically we still support Svelte 3 with svelte-check v4 but this gives us the opportunity to adjust that later without a major)
- breaking: throw an error when `--ignore` is used without `--no-tsconfig` #1074
- feat: remove dependency on `svelte-preprocess` (also see #2452)

closes #2243

WIP, will finish after my vacation.

### DIFF
--- a/packages/language-server/src/lib/documents/configLoader.ts
+++ b/packages/language-server/src/lib/documents/configLoader.ts
@@ -4,7 +4,7 @@ import { CompileOptions } from 'svelte/types/compiler/interfaces';
 // @ts-ignore
 import { PreprocessorGroup } from 'svelte/types/compiler/preprocess';
 import { importSveltePreprocess } from '../../importPackage';
-import _glob from 'fast-glob';
+import { fdir } from 'fdir';
 import _path from 'path';
 import _fs from 'fs';
 import { pathToFileURL, URL } from 'url';
@@ -47,6 +47,8 @@ const _dynamicImport = new Function('modulePath', 'return import(modulePath)') a
     modulePath: URL
 ) => Promise<any>;
 
+const configRegex = /\/svelte\.config\.(js|cjs|mjs)$/;
+
 /**
  * Loads svelte.config.{js,cjs,mjs} files. Provides both a synchronous and asynchronous
  * interface to get a config file because snapshots need access to it synchronously.
@@ -61,7 +63,7 @@ export class ConfigLoader {
     private disabled = false;
 
     constructor(
-        private globSync: typeof _glob.sync,
+        private globSync: typeof fdir,
         private fs: Pick<typeof _fs, 'existsSync'>,
         private path: Pick<typeof _path, 'dirname' | 'relative' | 'join'>,
         private dynamicImport: typeof _dynamicImport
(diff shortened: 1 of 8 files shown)


packages/svelte-check/package.json
```diff
@@ -27,11 +27,11 @@
...
         "picocolors": "^1.0.0",
         "sade": "^1.7.4",
-        "svelte-preprocess": "^5.1.3",
-        "typescript": "^5.0.3"
+        "svelte-preprocess": "^5.1.3"
     },
     "peerDependencies": {
-        "svelte": "^3.55.0 || ^4.0.0-next.0 || ^4.0.0 || ^5.0.0-next.0"
+        "svelte": "^3.55.0 || ^4.0.0-next.0 || ^4.0.0 || ^5.0.0-next.0",
+        "typescript": ">=5.0.0"
```
benmccann, line 34: Should we support some versions of Typescript 4? It might be nice for users who haven't been able to upgrade yet if it's not too hard for us

dummdidumm: Thanks for reminding me to bump the Svelte peer dependency to v4 😄 And since v4 has an implicit dependency requirement on TS >=5 we should, too. Better to nudge users towards upgrading

benmccann: I think users will have plenty of reason to upgrade on their own since Svelte 5 is so much better. But it'd probably be appreciated if we let them do it in their own time instead of forcing them. It can make things easier to have wider support as you have more flexibility around what order to upgrade various libraries

dummdidumm: This isn't forcing people to jump to Svelte 5, only Svelte 4 - which you need to do before upgrading to 5 anyway.

benmccann: yeah, I know. I was just saying people would want to upgrade to Svelte 4 in order to get to 5 so we don't need to push them too hard

benmccann:
Is `@jridgewell/trace-mapping` unused? It looks like maybe we could remove it?

dummdidumm:
No it's used within the language server

benmccann:
> No it's used within the language server

It should just be a dependency in that project then and not svelte-check, right?

dummdidumm:
No, the setup here is a bit different, we're bundling most of the language-server but externalize a few dependencies, which we then need to require separately.

benmccann:
Just curious, how is it decided which are bundled or why can't we bundle `@jridgewell/trace-mapping`?

packages/language-server/src/lib/documents/configLoader.ts
```diff
@@ -84,12 +86,19 @@ export class ConfigLoader {
...
-                cwd: directory,
-                // the second pattern is necessary because else fast-glob treats .tmp/../node_modules/.. as a valid match for some reason
-                ignore: ['**/node_modules/**', '**/.*/**'],
-                onlyFiles: true
-            });
+            const pathResults = new this.globSync({})
+                .withPathSeparator('/')
+                .exclude((_, path) => {
+                    // no / at the start, path could start with node_modules
+                    return path.includes('node_modules/') || path.includes('/.');
```
dummdidumm, line 93:
```suggestion
                    return path.includes('node_modules/') || path.includes('/.') || path[0] === '.';
```

packages/svelte-check/src/index.ts
```diff
@@ -30,11 +30,48 @@ async function openAllDocuments(
...
+                return true;
+            }
+        }
+        return false;
+    };
+    const absFilePaths = await new fdir()
+        .filter((path) => path.endsWith('.svelte') && !isIgnored(path))
+        .exclude((_, path) => {
+            path = path.slice(offset);
+            return path.startsWith('.') || path.startsWith('node_modules');
```
dummdidumm, line 69:
```suggestion
            return path.includes('/node_modules/') || path.includes('/.');
```

Issue #2148, opened 2023-09-03
Project References support


ziggi:
### Describe the bug

I'm using  [TypeScript's Project References](https://www.typescriptlang.org/docs/handbook/project-references.html) and it's looks like Svelte extension didn't work with it.

Extension output:
```
Initialize language server at  file:///d%3A/dev/svelte-ts-projects
Initialize new ts service at  d:/dev/svelte-ts-projects/tsconfig.json
Trying to load configs for d:/dev/svelte-ts-projects
Loaded config at  d:\dev\svelte-ts-projects\proj1\svelte.config.js
SnapshotManager File Statistics:
Project files: 0
Svelte files: 1
From node_modules: 0
Total: 1
```

### Reproduction

Clone https://github.com/ziggi/svelte-ts-projects
Open it in VS Code and open `svelte-ts-projects\proj1\svelte\TestComponent.svelte` file.
You will see `Cannot find module '@shared/SharedTest' or its corresponding type declarations.`

### Expected behaviour

It should just work.

### System Info

Svelte for VS Code v107.10.0

Version: 1.78.2 (user setup)
Commit: b3e4e68a0bc097f0ae7907b217c1119af9e03435
Date: 2023-05-10T14:39:26.248Z
Electron: 22.5.2
Chromium: 108.0.5359.215
Node.js: 16.17.1
V8: 10.8.168.25-electron.0
OS: Windows_NT x64 10.0.22621
Sandboxed: No

### Which package is the issue about?

Svelte for VS Code extension

### Additional Information, eg. Screenshots

<img width="728" alt="image" src="https://github.com/sveltejs/language-tools/assets/1020099/d3d083f4-4930-4808-abb8-e477a1a09e83">

jasonlyu123:
Some updates. I am making this the mega thread for project reference support since we didn't support it at all. Leaving some notes here as an update on the progress and also as a reminder in case I forgot about it, which I often do 😅.

The problem with this one is that project reference is used to load different tsconfig that doesn't use the name "tsconfig.json". Vite templates also often use this to differentiate server files and web files. So module resolution needs to use the redirected tsconfig. 

[...]

Issue #1976, opened 2023-04-05
Svelte language server forces too many tsconfig options


dummdidumm:
This came up in a conversation with @TehShrike - assume you have a project which you slowly want to convert from JS to TS. You have `"allowJs": false` and you use `svelte-check` - now you have a mismatch in reported errors, because the Svelte language server forces `allowJs` to `true`, which means JS file imports are followed, its types are picked up and used etc.

This got me thinking about the current list of things we force, and if we should no longer force some of them in the next major version of `svelte-check` (I feel like this should be seen as a breaking change).

These are the compiler options we currently force:

```ts
const forcedCompilerOptions: ts.CompilerOptions = {
    allowNonTsExtensions: true,
    target: ts.ScriptTarget.Latest,
    allowJs: true
    noEmit: true,
    declaration: false,
    skipLibCheck: true
}
```

#### allowNonTsExtensions

This is some internal compiler option - AFAIK we need this for basic functionality, but I'm not sure anymore.

#### target

The idea is to not allow CJS etc, but I'm no longer sure we should just flat out force it to latest. You might use globals in your `.svelte` files which do not exist in the version yet that you set in your `tsconfig.json` - `target` influences the default value of `lib`. Then again, something like Vite might auto-transpile it so it's fine. Not sure what's better.

#### allowJs

We need this so we can go to `.svelte` files from within TS files, get some typings etc. There may be one situation where we shouldn't set it though in my opinion: When running `svelte-check` with a referenced `tsconfig.json`. In that case, we're not in the context of an editor, so go-to etc is unimportant. We also should not blank `svelte/types/runtime/ambient.d.ts` in that case so the fallback ambient `*.svelte` module definition is picked up.
In the other cases it should be fine since the language server is only used within Svelte files, and the Typescript plugin doesn't override it so that described situation doesn't appear there.

#### noEmit

We need this because we're not building anything, just checking, and this avoids tsconfig warnings

#### declaration

Same as `noEmit`

#### skipLibCheck

I'm honestly not sure why we force this to `true` - should we just remove that?

@jasonlyu123 would love your thoughts on this.

jasonlyu123:
I think we can only add these when there isn't a tsconfig/jsconfig.json. And have a migration guide that points out some potential problems we can think of.

`allowNonTsExtensions`. I tested. If we don't have it, TypeScript will complain that it can't find the source file. 

[...]

Issue #1234, opened 2021-11-08
`svelte-check` ignoring `jsconfig.json`'s `exclude` option


kierangilliam:
`svelte-check` will not use a `jsconfig.json`'s `exclude` option to ignore `.svelte` files. 

Please see [this](https://github.com/kierangilliam/svelte-check-ignores-excludes) to reproduce. 

If you open this project in vscode and navigate to `jsconfig.json` you'll see that `src/routes` is excluded from type checking. This is confirmed to work for `*.js` files (see `src/routes/in-route.js`). However, as seen in `src/routes/index.svelte`, the language server will ignore the exclusion of this folder and perform type checking.

jasonlyu123:
Typescript's tsserver has a concept called inferred project. It seems like if according to the `include` and `exclude` the file doesn't belong to any config file, It would fall into this inferred project category. So it's not typescript respect the config and doesn't type check the `in-route.js` file. It's that the jsconfig.json doesn't apply to that file. If you enable the strict mode in the jsconfig.json and has a variable with `string | undefined` type in both `in-route.js` and `in-lib.js`. 

```js
/**@type {string | undefined}*/
let a;
```

Hover over it in the editor and you'll see that  `in-route.js` shows `let a: string` while `in-lib.js` shows` let a: string | undefined`.

So technically the difference is that we have a different approach to determine whether the file belongs to the `jsconfig`/`tsconfig`. 

But I don't know if we can make it consistent with the tsserver. Historically, setting svelte files in the `include` config doesn't do anything. When analysing the config file, typescript won't tell us if the file belongs to the config file. So we only check if the file is in the directory where the config is. Although including the svelte file in the `include` does include the file now. But there should be a lot of people who don't have svelte files in the `include`. This means if we change this behaviour **it'll be a breaking change**. 

And by the way, your exclude actually doesn't do anything. As the route path is not in the `include`. You can see the explanation in [the typescript doc](https://www.typescriptlang.org/tsconfig#exclude)

kierangilliam:
Thanks for your explanation. So, the best workaround to this issue would be to do the following in `.vscode/settings.json`, yeah?

```
{
	"svelte.plugin.typescript.diagnostics.enable": false
}
```

> But there should be a lot of people who don't have svelte files in the include. This means if we change this behaviour it'll be a breaking change.

[...]

Issue #2154, opened 2023-09-11
no error when running `svelte-check` on invalid `tsconfig.json`


DetachHead:
### Describe the bug

when your `tsconfig.json` is invalid, `svelte-check` does not complain
```
> tsc
tsconfig.json:12:47 - error TS6046: Argument for '--moduleResolution' option must be: 'node10', 'classic', 'node16', 'nodenext', 'bundler'.

12     "module": "NodeNext", "moduleResolution": "asdf"
                                                 ~~~~~~


Found 1 error in tsconfig.json:12

> svelte-check --tsconfig tsconfig.json

====================================
Loading svelte-check in workspace: c:\Users\user\Documents\my-app
Getting Svelte diagnostics...

====================================
svelte-check found 0 errors and 0 warnings
```

### Reproduction

1. put an invalid value in `tsconfig.json`, eg. `"moduleResolution": "asdf"`
2. run `svelte-check --tsconfig path/to/tsconfig.json`

### Expected behaviour

an error, like what `tsc` outputs:
```
tsconfig.json:12:47 - error TS6046: Argument for '--moduleResolution' option must be: 'node10', 'classic', 'node16', 'nodenext', 'bundler'.

12     "module": "NodeNext", "moduleResolution": "asdf"
                                                 ~~~~~~


Found 1 error in tsconfig.json:12
```

### System Info

- OS: windows 10

### Which package is the issue about?

svelte-check

PR #2463, opened 2024-08-12
feat: Project reference support

jasonlyu123:
#2148

It might make sense to make this part of the svelte-check v4 breaking changes. The way files are assigned to tsconfig needs to change, #1234 also needs that. And `languageServiceHost.getParsedCommandLine` also can't use the API we used to parse to tsconfig so it's also related to #1976. But we could also argue that assigning a file to a different tsconfig is already a breaking change. And we're unlikely to cherry-pick changes back to v3 so probably safer to make this v4 feature. 

There are a few main changes:

1. project management. the open document is assigned to a service containing the document as a project file. This is actually the important part. The files need to be loaded with referenced tsconfig, Otherwise, TypeScript will skip the check. One problem with this is that you'll need to run svelte-check multiple times with different tsconfig. Should we introduce an option similar to tsc's `--build`?

2. module resolution only has one step now. No try ts first and then try custom logic later because the project-reference-redirect thingy uses the language service host and therefore it always uses svelte-sys. 

3. Don't force that many tsconfig. But one question is should we still check incompatible compiler options? And should there be default a `target`? Removing the `target` default breaks a lot of our tests and might confuse a lot of users.  

4. if an open file is not a project file for any tsconfig, it'll be treated as if there isn't a tsconfig. 

5. update-import now run with `forAllService` because the default project assigns changes. Our update import doesn't block VSCode from actually renaming the file so sometimes the request is after the project files update, sometimes it is before. So using both `oldPath` and `newPath` doesn't reliably get the correct service and might unnecessarily create a default service. If it's folder rename it also won't work. 

We could probably revert 3 and 4. 4 does help with catching some problems with project references. Since only the project files for the referenced project are assigned to the project. 

While writing tests for this I also run into some problems with the existing test. Two are existing flaky tests and one just does not test anything because the assertion error is swallowed. Another change is that while updating the document, the `getSnapshot` method also unnecessarily opens a new service if it belongs to multiple services.  If we decide not to merge this, I'll separate the fixes into separate PRs.


dummdidumm:
Nice work!

> One problem with this is that you'll need to run svelte-check multiple times with different tsconfig. Should we introduce an option similar to tsc's --build?

Can you elaborate on that? I don't fully understand? In which case would you need to run svelte-check multiple times, and how would you need to run it? And how would that differ to what you can do today?

> module resolution only has one step now. No try ts first and then try custom logic later because the project-reference-redirect thingy uses the language service host and therefore it always uses svelte-sys.

Just curious: Is this is a nice side effect of this work or a necessary thing? This may even speed up things a little, right? Since we're not going to do it two times for Svelte files.

> Don't force that many tsconfig. But one question is should we still check incompatible compiler options? And should there be default a target? Removing the target default breaks a lot of our tests and might confuse a lot of users.

Looking at the changed code there isn't any change to the tsconfig options? Did you revert that to make it separate?

> Another change is that while updating the document, the getSnapshot method also unnecessarily opens a new service if it belongs to multiple services.

Can you give an example? How often will that happen? What would the new service be? How bad would that be for performance, or could that even introduce bugs?

jasonlyu123:
> In which case would you need to run svelte-check multiple times

Take our repo for example. The root tsconfig is a "solution style" tsconfig which includes nothing and is meant to load `language-server` and `svelte-vscode`. if you run tsc with this tsconfig, nothing will be built. You need to either run tsc twice in language-server and svelte-vscode, or run `tsc -b` with the root tsconfig. Maybe there are already people running it twice with the current svelte-check version since you can pass in an arbitrary tsconfig path. 

> Is this is a nice side effect of this work or a necessary thing

More like "a nice side effect". We'll need the compiler host created by TypeScript to redirect project reference `.d.ts` to the corresponding source file which means a lot of `.ts` files will go through the svelte-sys route anyway. 

> Looking at the changed code there isn't any change to the tsconfig options

If there is tsconfig the forced option won't override the tsconfig. The `parseDefaultCompilerOptions` is the old behaviour and overrides some TypeScript defaults. I should remove some unused logic to make that clearer. 

> the getSnapshot method also unnecessarily opens a new service if it belongs to multiple services

I found that with our test because the update-import test keeps timeout after the update-import change. I forgot to mention that in the PR description, adding that now. Some test files belong to multiple tsconfig because we didn't exclude them from the root tsconfig(`test/plugins/typescript/testfiles/tsconfig.json`). During the program creation of the root tsconfig, TypeScript asked for the snapshot for those files because they're project files. Our `getSnpshot` logic then looks at the files and assigns them to the nearest tsconfig. ex.
`preferences/code-action.svelte` to `preferences/tsconfig.json`. It probably doesn't happen that often in user setup but it does speed up our test a few seconds.

packages/language-server/src/plugins/typescript/service.ts
```diff
@@ -648,26 +737,31 @@ async function createLanguageService(
...
 
-    function getDefaultExclude() {
-        return ['node_modules', ...ignoredBuildDirectories];
+        const parsedConfig = ts.parseJsonConfigFileContent(configJson, tsSystem, workspacePath);
+
+        const compilerOptions: ts.CompilerOptions = {
+            ...parsedConfig.options,
+            target: ts.ScriptTarget.Latest,
+            allowNonTsExtensions: true,
+            moduleResolution: ts.ModuleResolutionKind.Node10
```
dummdidumm, line 761: Why force `"moduleResolution": "node"`? We didn't do that previously - just an accident?

dummdidumm:
>  During the program creation of the root tsconfig, TypeScript asked for the snapshot for those files because they're project files. Our getSnpshot logic then looks at the files and assigns them to the nearest tsconfig.

Ok so just be clear: You're saying you found a somewhat unrelated bug/performance issue and fixed that in this PR? I first understood it as "this PR has this drawback because it now does it", but it seems I understood wrong and it's actually the other way around?

> update-import now run with forAllService because the default project assigns changes. Our update import doesn't block VSCode from actually renaming the file so sometimes the request is after the project files update, sometimes it is before. So using both oldPath and newPath doesn't reliably get the correct service and might unnecessarily create a default service. If it's folder rename it also won't work.

This was something that could happen before, too, but now can happen more often because we're possibly doing it for multiple services? Why would it unnecessarily create a default service?

> Don't force that many tsconfig. But one question is should we still check incompatible compiler options? And should there be default a target? Removing the target default breaks a lot of our tests and might confuse a lot of users.

What about only setting target if we see it's either a very old one that likely breaks, or is not set at all?

jasonlyu123:
>  You're saying you found a somewhat unrelated bug/performance issue and fixed that in this PR

Not a drawback but just explain why the change was made. Didn't remember the exact problem when I wrote the previous comment. Now I remember it is related to project references. when: a file is loaded in `referenceA` because of an import but is a project file of  `referenceB`. This will cause the `referenceB` project to open while it may never be used. 

> Why would it unnecessarily create a default service

It is more related to the project assignment issue. Because now the client-opened file needs to match the "include" and "exclude" of the tsconfig otherwise we'll be treated as if it doesn't have any tsconfig. The project assignment change can be a separate PR but because project references also need this "include"/ "exclude" match check. So I put it in here as well but it could be separated later. Having it might catch some problem in project reference but doesn't have a dedicated test. 

> What about only setting target if we see it's either a very old one that likely breaks, or is not set at all?

The default is ES5 in typescript 5.5. Should we force it to `ES2015`?

packages/language-server/src/plugins/typescript/service.ts
jasonlyu123: This one is the default when there isn't a tsconfig/jsconfig.json. So the `moduleResolution` should be undefined here. Even if it's removed it'll be set to `Node10` in L699.

dummdidumm:
> The default is ES5 in typescript 5.5. Should we default it to ES2015? And if it's explicitly set to ES5, do we force it to be ES2015?

Let's go with this: Default to Latest if nothing is set, like currently. If something earlier than ES2015 is set, set to ES2015.

packages/language-server/src/plugins/typescript/service.ts
dummdidumm: my bad, I should've looked at this with a proper diff view

dummdidumm:
Another clarification question: You linked several issues above - which of those are actually fixed by this PR?
- #2148 - I guess this one is solved
- #1234 - it sounds like this is also solved then? Though your comment in there is somewhat valid: is there any way how we can ease the migration a bit for people having outdated tsconfigs that don't have Svelte files in their includes?
- #1976 - I guess this one is solved, too

jasonlyu123:
Yeah. These 3 are fixed by this PR. I am a bit uncertain if these should be separated but turns out to be easier to do it at once. 

> is there any way how we can ease the migration a bit 

What about we do something like that `No inputs were found in config file '{0}'. Specified 'include' paths were '{1}' and 'exclude' paths were '{2}'.` diagnostic but check only svelte files? We still create the service based on the nearest tsconfig.json and use that to check the project files. So if the `parsedCommandLine.fileNames` doesn't have any svelte file, it doesn't have `references` and `files`  (https://github.com/microsoft/TypeScript/blob/c8a7d589e647e19c94150d9892909f3aa93e48eb/src/compiler/commandLineParser.ts#L3216) we'll report a similar diagnostic on the tsconfig.

dummdidumm:
To reiterate: If
- we open a Svelte file
- we find a nearby tsconfig
- it doesn't have project references
- we read the file and it turns out there is no Svelte file at all in there
- THEN we issue a warning (or error?) (but where?) (if it's a warning do we fall back to something?) ?

sounds good to me. For svelte-check we can error, for VS Code I'm not really sure what to do. If we error for svelte-check only, that likely means you need to read the returned diagnostics property of the "parse the config" method, right? In that case, maybe we can combine that with #2154

jasonlyu123:
I set the diagnostics in the config file.
圖片

And for the svelte-check, I put it in the `configErrors`. we can probably change it to show every `configErrors` in svelte-check for #2154. But since there are more errors probably need to print it like other diagnostics and need adjustment with the writer. We probably don't need to show the code for config errors. we previously didn't just throw an error and tsc also doesn't.

packages/language-server/src/plugins/typescript/service.ts
```diff
@@ -640,6 +737,21 @@ async function createLanguageService(
             }
         }
 
+        const svelteConfigDiagnostics = checkSvelteInput(parsedConfig);
+        if (docContext.reportConfigError) {
```
dummdidumm, line 741: Shouldn't we only do this when there's actual diagnostics?
```suggestion
        if (docContext.reportConfigError && svelteConfigDiagnostics.length) {
```

PR #2468, opened 2024-08-19
typo fix in DisagnosticsProvider

ericmuckley:
A typo was fixed in DiagnosticsProvider.ts where a closing curly brace was missing after `$bindable()`.

Old:
```
 "Cannot use 'bind:' with this property. It is declared as non-bindable inside the component.\n" +
`To mark a property as bindable: 'let { ${propName} = $bindable() = $props()'`;
```

New:
```
 "Cannot use 'bind:' with this property. It is declared as non-bindable inside the component.\n" +
`To mark a property as bindable: 'let { ${propName} = $bindable() } = $props()'`;
```


jasonlyu123:
Thanks!

Issue #1056, opened 2021-06-15
Difference between how Svelte and JS imports from node_modules are handled


dummdidumm:
Through #1048, I noticed there is a difference between how JS and Svelte files imported from node_modules are handled:
- Imports from node_modules that have no accompanying type definition (`d.ts` file) are marked as `any` which errors on strict mode (no implicit any). This means TS does not attempt to aquire the type structure from the node_modules package.
- Svelte imports from node_modules are read and treated as a regular file inside the project. This means they don't need a accompanying type definition and their types are aquired, which can be good or bad depending on the quality of the Svelte file. Some libraries like `svelte-loading-spinners` take advantage of this and provide TS-versions of Svelte files as types, which is dangerous.

I think in the long run it would be best to align with the behavior of TypeScript here. This would probably a breaking change to some users which do deep imports of Svelte files within libraries which don't have a `.svelte.d.ts` next to them. I'm also not sure how to even achieve this as I thought this stuff is handled inside `ts.resolveModuleName`.
@jasonlyu123 thoughts?

cor:
Is there a possible workaround for this in the meantime? It's hard to take advantage of `svelte-check` when the error list is polluted by dependencies.

dummdidumm:
Unfortunately not

cor:
So I currently have two options:

1) Check both `.svelte` and `.ts` files, with the error list polluted by dependencies
2) Only check `.svelte` files, `.ts` files remain unchecked

Correct?

dummdidumm:
Yes. You could check `.ts` files with `tsc --noEmit` however. But that _might_ produce some false positives, depending on whether or not you have named imports from Svelte files inside TS files.

probablykasper:
Is there any workaround for this yet, or are the only two options what @cor mentioned?

dummdidumm:
Could you specify what your problem is? Files inside node_modules shouldn't get diagnosed anymore.

probablykasper:
I'm running `svelte-check --tsconfig ./tsconfig.json` and it's showing errors/hints from node_modules:

<details>
<summary>Logs</summary>

```
> @ check /Users/kasper/dev/git/kryp
> svelte-check --tsconfig ./tsconfig.json


====================================
Loading svelte-check in workspace: /Users/kasper/dev/git/kryp
Getting Svelte diagnostics...

/Users/kasper/dev/git/kryp/node_modules/@tauri-apps/api/helpers/event.ts:7:1
[...]

PR #2478, opened 2024-08-28
feat: resolve Svelte components using TS from exports map

dummdidumm:
This change allows people to write export maps using only a `svelte` condition (and no `types` condition) and still have the types for their components resolved (i.e. the import is found) as long as they use TypeScript (i.e. have lang="ts" attribute) inside it. This should help people using monorepo setups with strong typings and not wanting to provide d.ts files alongside.

This is achieved doing three adjustments:
- add `customConditions: ['svelte']` to the compiler options, so that TypeScript's resolution algorithm takes it into account
- ~ensure that Svelte files have a module kind of ESM, so that TypeScript's resolution algorithm goes into the right branches~ no longer needed since the switch to using `.d.svelte.ts` as the "virtual file check" entry point
- deal with `.d.svelte.ts` files in the context of an exports map, because that's what TypeScript will try to resolve this to in the end

This is also related to #1056 insofar that we align with TypeScript for this new capability: We don't resolve the file if it's a component not using TypeScript (i.e. not having the lang="ts" tag), similar to how TypeScript does not resolve .js files within node_modules

As a side effect, this is a major version bump for language-tools and the typescript plugin, and consequently for svelte-check/vs code extension

### DIFF
--- a/packages/typescript-plugin/src/module-loader.ts
+++ b/packages/typescript-plugin/src/module-loader.ts
@@ -146,7 +146,7 @@ export function patchModuleLoader(
         return resolved.map((tsResolvedModule, idx) => {
             const moduleName = moduleNames[idx];
             if (
-                // Only recheck relative Svelte imports or unresolved non-relative paths (which hint at node_modules,
+                // Only recheck relative Svelte imports or unresolved non-relative paths (which hint at node_modules
                 // where an exports map with "svelte" but not "types" could be present)
                 (!isSvelteFilePath(moduleName) && (moduleName[0] === '.' || tsResolvedModule)) ||
                 // corresponding .d.ts files take precedence over .svelte files


dummdidumm:
Mhm not sure why this fails - works locally for me (Windows)

jasonlyu123:
I didn't see the package.json with exportMap. Maybe it's ignored because of gitignore? About `.d.svelte.ts `, this isn't exclusive to export maps import. TypeScript would also try this with relative import. Not sure if anyone uses it to override the type definition of the svelte file. This will change the resolved module from the .d.svelte.ts to the actual svelte file.

dummdidumm:
Doh, you're right, the node modules folder isn't checked in.
Good point about the files - theoretically can happen for svelte.ts, too. I'll look into it

dummdidumm:
I realized that since 5.0, TypeScript will check for `d.svelte.ts` first before checking `.svelte.ts` etc - which we can take to our advantage to do the following:
- be less exposed to the danger of clashing with a real `.svelte.ts` file (in Svelte 5 this will be more common)
- when doing `import Foo from './foo.svelte`, `foo.svelte` will now take precedence over a sibling `foo.svelte.ts` file. You can import that sibling by doing `import { foo } from './foo.svelte.js'`. It's definitely discouraged to have files with the same name, but it's possible now. Strictly speaking this is a breaking change because the order is different now

jasonlyu123 (approved):

dummdidumm:
Having second thoughts about this - should we really do this? It could encourage bad patterns. On the other hand we could also embrace it - I'm not sure. This needs discussion with other maintainers.

I'm gonna split out the other changes to land them and put this into draft.

Issue #2479, opened 2024-08-29
vscode multi-root workspace autoImportFileExcludePatterns not being excluded from autocomplete [monorepo]


datstarkey:
### Describe the bug

When using a mono-repo its quite common to share some components and export them without building as they'll never be released as a library. using the exports and pointing svelte and types to the index.ts works, and provider correct intellisense

when using pnpm workspace for this, the imports all work fine however the quick-fixes/code completion suggests the relative path alongside of the import which is incorrect, as sub paths is not allowed and should only be imported from the index.ts. 

![image](https://github.com/user-attachments/assets/768721c8-7be4-4d8a-8c89-dd50f428dd1f)
This also seems to be a random order, as sometimes the first completion is the relative path instead of the shortest one, I've played around with all the importModuleSpecifiers and it doesn't seem to affect the order of them. This can be insanely annoying if you have multiple components with the same name like InputText,InputNumber,InputDate, resulting having to go down alot of different imports before the one you want.

Inside of the vscode settings at the root of the monorepo you can add the autoImportFileExcludePatterns to get around this.
![image](https://github.com/user-attachments/assets/c379b6e1-1bb6-4678-858c-70e7f2b8e12c)

Resulting in the auto-complete only giving the desired result.
![image](https://github.com/user-attachments/assets/699f80ba-7f3f-47e4-8db9-985bb54da267)

This is great, however if you use a vscode multi-root  workspace, this stops working. I've tried adding the settings to the .code-worspace and it works as Intended inside typescript files but not inside svelte files. so it looks like the typescript-plugin isn't detecting the code-workspace settings?

### Reproduction

https://github.com/datstarkey/svelte-workspace-import-exclude/tree/master

1.  clone the repo and pnpm i
2. open the root directory first not the workspace.
3. go to packages/src/lib/working.ts and run the quickfix on stuff, notice only the correct import is available.
4. go to packages/src/lib/notWorking.svelte and run the quickfix on stuff, notice only the correct import is available.
5. open the project from the monorepo.code-workspace inside the root.
6. repeat steps 3 & 4, notice how the typscript file works, but the svelte file still gives the wrong import.

### Expected behaviour

The typescript.preferences.autoImportFileExcludePatterns vscode setting should produce the same behaviour in workspace or folder/root level for svelte files.

### System Info

- OS: [Windows]
- IDE: [VSCode]

### Which package is the issue about?

Svelte for VS Code extension

### Additional Information, eg. Screenshots

[...]

PR #2480, opened 2024-08-30
fix: allow autoImportFileExcludePatterns to ignore files outside the root

jasonlyu123:
The first part of #2479. I will probably open a separate PR to deprioritise the component auto-import later. I did try it before and it should be possible to detect if there is a barrel import in the same list.

### DIFF
--- a/packages/language-server/src/ls-config.ts
+++ b/packages/language-server/src/ls-config.ts
@@ -478,16 +478,25 @@ export class LSConfigManager {
         };
     }
 
-    getTsUserPreferences(lang: TsUserConfigLang, workspacePath: string | null): ts.UserPreferences {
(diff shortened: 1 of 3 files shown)


packages/language-server/src/ls-config.ts
```diff
@@ -478,16 +478,25 @@ export class LSConfigManager {
         };
     }
 
-    getTsUserPreferences(lang: TsUserConfigLang, workspacePath: string | null): ts.UserPreferences {
+    getTsUserPreferences(
+        lang: TsUserConfigLang,
+        normalizedWorkspacePath: string | null
```
dummdidumm, line 483: you changed the name here - just for readability, or is there a corresponding normalization missing in another file?

PR #2488, opened 2024-09-06
warn svelte input only when program has no svelte files at all

jasonlyu123:
While implementing the delayed check. I found out that tsserver actually assigns files that aren't project files into the service that loads the files through import. And this has all sorts of inconsistent states 😅. So still need to compare the behavior to tsserver.

### DIFF
--- a/packages/language-server/src/plugins/typescript/DocumentSnapshot.ts
+++ b/packages/language-server/src/plugins/typescript/DocumentSnapshot.ts
@@ -109,7 +109,7 @@ export namespace DocumentSnapshot {
         tsSystem: ts.System
     ) {
         if (isSvelteFilePath(filePath)) {
-            return DocumentSnapshot.fromSvelteFilePath(filePath, createDocument, options);
+            return DocumentSnapshot.fromSvelteFilePath(filePath, createDocument, options, tsSystem);
         } else {
             return DocumentSnapshot.fromNonSvelteFilePath(filePath, tsSystem);
         }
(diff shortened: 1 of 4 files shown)


dummdidumm:
Honestly, if this makes for too much complicated logic, and if this doesn't uncover an unrelated inconsistency, then I would just not show the error in VS Code, and only show it in svelte-check, by checking if there are any Svelte file in the program files we iterate over to get all diagnostics.

dummdidumm:
> The svelte file would use our default config, but it would use the tsconfig in svelte-check

When would that happen? I'm not sure I follow. Can you give an example?

jasonlyu123:
You can create a create-vite project with typescript and svelte. Then update svelte-check to v4 and remove Svelte files from the include config. 

In the editor, the current logic is to assign files to a shared service without a tsconfig. But svelte-check is the other way around it loads the tsconfig and loops through the files in the program.

dummdidumm:
But isn't that correct? In the editor you can access files outside of your tsconfig definition, so they need to go _somewhere_. For `svelte-check` (and IIUC also `tsc`) that's not the case, because those only look at the included files, and then traverse from there. Anything that is not found through that traversal just doesn't exist for them.

jasonlyu123:
The change is only for files that are already in the program. They would already be checked in svelte-check. The other files are still assigned to the implicit project, i.e. as if there isn't a tsconfig.

dummdidumm (approved):
👍 I realized that this fixes exactly what I was worrying about

PR #2496, opened 2024-09-11
fix: update SvelteKit trailingSlash types

bluwy:
I noticed the types is incorrect for `trailingSlash`

image

This aligns with SvelteKit's types at https://github.com/sveltejs/kit/blob/108cb127eb0a4c3655aeff1f749bcbd98b70324e/packages/kit/types/index.d.ts#L1578

I also removed the `(missing)` from the doc text since I don't think it's specific to missing trailing slashes only? SvelteKit could also handle requests differently if it gets a URL with trailing slash but with the `never` config (which it 308 redirects to the non-trailing slash path)


dummdidumm (approved):
👍

PR #2502, opened 2024-09-13
chore: upgrade to chokidar 4

benmccann:
Removes a dozen dependencies


paulmillr:
Shouldn’t second arg be default = {} instead?

benmccann:
Chokidar does that be internally in the watcher constructor, so it shouldn't need to be done elsewhere

benmccann:
This PR is ready for review now

packages/svelte-check/src/index.ts
```diff
@@ -149,8 +149,12 @@ class DiagnosticsWatcher {
         filePathsToIgnore: string[],
         ignoreInitialAdd: boolean
     ) {
-        watch(`${workspaceUri.fsPath}/**/*.{svelte,d.ts,ts,js,jsx,tsx,mjs,cjs,mts,cts}`, {
-            ignored: ['node_modules', 'vite.config.{js,ts}.timestamp-*']
+        watch(workspaceUri.fsPath, {
+            ignored: [
+                'node_modules',
+                'vite.config.{js,ts}.timestamp-*',
```
dummdidumm, line 155: Is this correct? AFAIK `node_modules` meant that everything with node_modules in the path was ignored, right? If not, we should change that.
The vite config string also looks suspiscuous, it's still using glob syntax, does this need to turn into a regex?

dummdidumm: Ok this has a couple of more bugs, will fix

benmccann:
awesome. thanks for the cleanup!

PR #2528, opened 2024-10-09
fix: include named exports in svelte 5 type

paoloricciuti:
This should fix https://github.com/sveltejs/svelte/issues/13508

I'm missing writing and fixing the tests but i'm pushing this already so that it can be validated. I'll fix tests and write new later.

EDIT: while working on this i noticed that even if a component had `<svelte:options runes />` it was still exporting like if it was not in rune mode. A bit of digging led me to the discovery that that option was never read. I added this but i don't know if i misunderstood something.

Also i write some new test and despite the fact that it's just for runes is still requiring an `expectedv2` file which seems a bit weird...is there some configuration i need to set to prevent this?


dummdidumm:
Code looks good 👍 you can add `.v5` to the end of the test folder, then it will only be run in Svelte 5, and then only `expectedv2` will be needed

paoloricciuti:
> Code looks good 👍 you can add `.v5` to the end of the test folder, then it will only be run in Svelte 5, and then only `expectedv2` will be needed

Oh gotcha...will update then.

P.s. I know this repo is less contributor friendly to begin with but I think we should try to write those things somewhere 😌

dummdidumm:
Totally agree, a contribution guide is needed

dummdidumm (approved):

paoloricciuti:
> Totally agree, a contribution guide is needed

I'll try to write down something as soon as i can but i will probably need a lot of feedback since as you've seen i'm missing some info 😄

Issue #2529, opened 2024-10-09
Warnings in script tag get quick-fixed in the wrong place


paoloricciuti:
### Describe the bug

If i have code like this
```svelte
<script lang="ts">
    let count = $state(2);
    let x = count;
    //      ^^^^^ State referenced in its own scope will never update. Did you mean to reference it inside a closure?svelte(state_referenced_locally)
</script>

<div></div>
```
and i try to quick-fix it it get's quick-fixed like this
```svelte
<script lang="ts">
    let count = $state(2);
    let x = count;
    //      ^^^^^ State referenced in its own scope will never update. Did you mean to reference it inside a closure?svelte(state_referenced_locally)
</script>

<!-- svelte-ignore state_referenced_locally -->
<div></div>
```

### Reproduction

1. open VSCode
2. paste the above code
3. try to quick fix it

### Expected behaviour

```svelte
<script lang="ts">
    let count = $state(2);
    // svelte-ignore state_referenced_locally
    let x = count;
</script>

<div></div>
```

### System Info

VSCode

### Which package is the issue about?

Svelte for VS Code extension, svelte-language-server

PR #2531, opened 2024-10-09
fix: correctly position quick fixes in instance and module

paoloricciuti:
I took a shot at closing #2529 

some tests seems to be failing but they also seems to be unrelated to the change but maybe i'm wrong. Also, i need some guidance on where and how to write new tests here 🫨 

I also have a problem with indentation because the indent seems to be always an empty string...i tried logging the `afterStartLineStart` and it seems to consider all the content on a single lnie despite having some `\n` in it


packages/language-server/src/plugins/svelte/features/getCodeActions/getQuickfixes.ts
```diff
@@ -207,26 +225,31 @@ function getSvelteIgnoreEdit(
     const indent = getIndent(afterStartLineStart);
 
     // TODO: Make all code action's new line consistent
-    const ignore = `${indent}<!-- svelte-ignore ${code} -->${EOL}`;
+    let ignore = `${indent}// svelte-ignore wtf ${code}${EOL}${indent}`;
```
dummdidumm, line 228: 😄 
```suggestion
    let ignore = `${indent}// svelte-ignore ${code}${EOL}${indent}`;
```

dummdidumm (commented):
Looks good! Could you add a test for this? You can enhance the file `svelte-ignore-code-action.svelte` for this and add the test to the `It should provide svelte ignore code actions` test suite in `getCodeAction.test.ts`,

packages/language-server/src/plugins/svelte/features/getCodeActions/getQuickfixes.ts
paoloricciuti: Hahahaha i was getting frustrated because changes were not reflecting during debugs 😄

PR #2540, opened 2024-10-21
feat(svelte2tsx): autotype `const load = ...` declarations

GauBen:
Hi!

This PR adds autotyping to `const load = ...` declarations, opening the door to SvelteKit "plugins"

Usage might be:

```ts
export const load = myPlugin((pluginData, event) => {
  return pluginData
})

// Dummy plugin
// We write the generics this way for the types to flow from output to input
function myPlugin<T>(
  load: T extends (event: infer Event) => infer Output
    ? (data: Record<string, unknown>, event: Event) => Output
    : never,
): T {
  return ((event) => load({ hello: "World!" }, event)) as T;
}
```

It works by rewriting `const load = ...` as `const load = (...) satisfies PageLoad`, letting TypeScript do the generic instantiation. 

`satisfies` was added in TS 4.9 and peer TS is set to `^4.9.4`, so this should not be a breaking change


dummdidumm (approved):
thank you!

PR #2544, opened 2024-10-24
chore: optimize svelte-check bundle size

jasonlyu123:
Following Ben's effort to trim down the size. I was actually trying to add some glob library back because chokidar v4 removed the glob support entirely() so was trying to trim it to balance it out 😆. The difference seems relatively large. The bulk of it seems to be HTML custom data which took around 500KB.

### DIFF
--- a/packages/language-server/package.json
+++ b/packages/language-server/package.json
@@ -7,7 +7,8 @@
     "exports": {
         "./package.json": "./package.json",
         ".": "./dist/src/index.js",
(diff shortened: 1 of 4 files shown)


jasonlyu123:
It's technically a breaking change. So draft until we're planning for svelte-check v4

jasonlyu123:
Actually, no, why did I remember this being a breaking change :sweat_smile:

Issue #2541, opened 2024-10-22
Imports completion not always working since version 109.0.1


dennisvanderweide:
### Describe the bug

In all versions >= 109.0.1 of the _Svelte for VS Code_ extension, I stopped getting suggestions for some imports.
Some packages still work while others never do. Sometimes for a brief moment after I reload my VS Code, I have some correct suggestions, but after this first time I always just get a "No suggestions.".

### Reproduction

I have a mono repo project with a SvelteKit app. This project runs in a Docker container using devcontainers.
When trying to import packages in a `.svelte` file, most of the time I get correct suggestions, but for some packages, in my case `@capacitor/haptics`, I don't. Other packages always work, this package never works.
After downgrading the _Svelte for VS Code_ extension to version <= 109.0.0 and restarting VS Code, I have no problems with imports anymore.

### Expected behaviour

I would expect imports to work for every package, just like they did before version 109.0.1.

### System Info

- OS: Linux 6.11.3 with Gnome 47 on Wayland
- IDE: VSCode 1.94.2 (problem is only on VS Code >= 1.82)
- Svelte for VS Code 109.1.0

### Which package is the issue about?

Svelte for VS Code extension

jasonlyu123:
Can you provide a reproduction repository and more detail on what kind of import completion has the problem? I can't reproduce this.
![圖片](https://github.com/user-attachments/assets/1afbea73-dd2e-4f46-acf5-bdcfe1f56bfb)

sebwib:
I have the same issue. Restarting svelte language server helps, for a while.
I can't provide a repo because it doesnt happen in small repos I have created. Can this be related to project size? Our project is very large

PR #2545, opened 2024-10-25
feat: support typescript 5.6

jasonlyu123:
I bumped typescript-auto-import-cache for #2541, but there is no test for it. I am not really sure what the exact condition triggering this is, but the problem is that the caching implementation of typescript-auto-import-cache is out of sync with TypeScript 5.6. TypeScript added another parameter for some functions. 

Mark as a draft for now, I'll add the support for [Exclude Patterns for Auto-Imports](https://devblogs.microsoft.com/typescript/announcing-typescript-5-6/#exclude-patterns-for-auto-imports) and maybe [Granular Commit Characters](https://devblogs.microsoft.com/typescript/announcing-typescript-5-6/#exclude-patterns-for-auto-imports) later.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
         "lint": "prettier --check ."
     },
     "dependencies": {
-        "typescript": "^5.5.2"
+        "typescript": "^5.6.3"
     },
     "devDependencies": {
         "cross-env": "^7.0.2",
--- a/packages/language-server/package.json
+++ b/packages/language-server/package.json
@@ -61,7 +61,7 @@
         "svelte": "^4.2.19",
         "svelte2tsx": "workspace:~",
         "typescript": "^5.5.2",
-        "typescript-auto-import-cache": "^0.3.3",
+        "typescript-auto-import-cache": "^0.3.4",
         "vscode-css-languageservice": "~6.3.0",
         "vscode-html-languageservice": "~5.3.0",
         "vscode-languageserver": "9.0.1",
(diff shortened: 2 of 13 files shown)


packages/language-server/test/plugins/typescript/features/CompletionProvider.test.ts
```diff
@@ -191,6 +219,7 @@ describe('CompletionProviderImpl', function () {
 
         assert.deepStrictEqual(eventCompletions, <CompletionItem[]>[
             {
+                commitCharacters: [],
```
jasonlyu123, line 222: Empty array and undefined should behave the same but won't be overridden by `itemDefaults`

packages/language-server/test/plugins/typescript/features/CompletionProvider.test.ts
```diff
@@ -1290,7 +1330,7 @@ describe('CompletionProviderImpl', function () {
             insertText: undefined,
             insertTextFormat: undefined,
             labelDetails: undefined,
-            commitCharacters: ['.', ',', ';', '('],
+            commitCharacters: [],
```
jasonlyu123, line 1333: This one is returned by the TypeScript. The TypeScript PR mentioned it is to prevent something like this from triggering accepting completion suggestions. It was added to VSCode first and then migrated to TypeScript. 

```ts
someFunction(`somting.`)

function someFunction(str: 'something.a', 'something.b') {}
```

dummdidumm (approved):
thank you!

PR #2548, opened 2024-10-28
fix: add Sveltekit snippet for Svelte 5

nielk:
When using SvelteKit with Svelte 5, the SvelteKit snippets still generate snippets with Svelte 4 syntax.

I fixed this by detecting the current Svelte version and generating Svelte 4 or Svelte 5 snippets according to it.

It follows a discussion from `Ypsilon` on the Svelte discord:

> I know there is a massive list of things to do probably , but wouldn't it be nice to have a use svelte 5 syntax option in the vscode extension. So when you use snippets, or feature like create route that it creates the files with the new runes syntax. And not the old export let data syntax ?


YpsilonTM:
Thank you sir, You are a hero. Let's hope it gets merged.

dummdidumm (approved):
thank you!

YpsilonTM:
`Code style issues found in 10 files. Run Prettier with --write to fix.` 🥲

Issue #2493, opened 2024-09-07
Bundled Svelte 4 types loaded in a Svelte 5 project and causing conflict


jethrolarson:
### Describe the bug

See title and repro

Work around is to add the types to the callbacks manually but that skips typechecking so not ideal.

### Reproduction

```ts
  import { writable, type Writable } from 'svelte/store';
...
  let _curMessage: string;
  const currentMessage = writable<string>('');
  currentMessage.subscribe((cur) => (_curMessage = cur));
```

![image](https://github.com/user-attachments/assets/820092e7-6955-4b7a-9342-b6555cd03c57)

node_modules/svelte/src/store/shared/index.js:

![image](https://github.com/user-attachments/assets/00e0aaa0-3d1e-4c83-9678-ecddbd13df27)

### Logs

```shell
n/a vscode
```

### System Info

```shell
osx apple m1 Sonoma

VSCode
Version: 1.90.1 (Universal)
Commit: 611f9bfce64f25108829dd295f54a6894e87339d
Date: 2024-06-11T21:02:41.372Z
Electron: 29.4.0
ElectronBuildId: 9593362
Chromium: 122.0.6261.156
Node.js: 20.9.0
V8: 12.2.281.27-electron.0
OS: Darwin arm64 22.1.0
```

### Severity

annoyance

dummdidumm:
@jasonlyu123 do you have an idea why our bundled types are loaded? Is it because of our `d.ts` files in `svelte2tsx` which reference Svelte imports? I always had the impression that those are pointing back to the user's Svelte version.

jasonlyu123:
It's like `require.resolve('svelte', '/path/to/shims.d.ts')` I think. Not sure if we can move the import to the generated code and pass it into the helper function as a parameter. Or if we could create a virtual module inside the user's node_modules to re-export from svelte.

PR #2561, opened 2024-11-05
fix: copy global types into hidden folder in IDE contexts 

dummdidumm:
Since the global types now reference `svelte/...` imports, they need to be placed in the same context the users' Svelte package lives in. Else these imports would load the types of the Svelte version that comes bundled with the IDE.

https://github.com/sveltejs/language-tools/issues/2493

(side note: we had this problem before, when loading `svelte/elements`, and solved it through #9070, but that solution is not applicable here)

### DIFF
--- a/packages/language-server/.gitignore
+++ b/packages/language-server/.gitignore
@@ -1,4 +1,4 @@
 dist/
 .vscode/
 node_modules/
-!test/plugins/typescript/features/diagnostics/fixtures/exports-map-svelte/node_modules/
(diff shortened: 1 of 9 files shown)


jasonlyu123 (approved):
Looks good. This resolved the double result for definition and https://github.com/sveltejs/svelte/issues/13268. I can't get the repo of #2493 to work anymore even without the fix. Not sure if the conflict is gone or what, but anyway at least the other problems are fixed.

Issue #2555, opened 2024-10-24
Allow exporting of types from module script


tzuleger:
### Describe the problem

In Svelte <=4, it was possible to export JSDOC custom types through the `context="module"` `<script>` tag by simply declaring the type.

In Svelte 5, it seems that types defined in the `module` script are not exported. At least, there's no clear way of doing it or no documentation on it.

### Describe the proposed solution

Allow users to export types from the module script, to be imported from anywhere in the project.

`$lib/components/MyComponent.svelte`:
```html
<script module>
/** @typedef {"myType"} MyType */
</script>
<script>
</script>

<!-- stuff -->

<style>
</style>
```

`$lib/util/misc.js`:
```js
/** @import { MyType } from "$lib/components/MyComponent.svelte" */

/**
 * @param {MyType} x
 */
function stuff(x) {
  
}
```

### Importance

would make my life easier

Issue #2538, opened 2024-10-18
ScriptKind Error with library projects


datstarkey:
### Describe the bug

When working with a library project in a mono repo with the tsconfig module resolution set to : `"moduleResolution": "NodeNext"`, it eventually starts getting this error on the svelte LSP, which eventually seems to cause the TS server or Svelte server to crash at some point and loses all intellisense.

```
Error: Debug Failure. False expression: Script kind should match provided ScriptKind:3 and sourceFile.scriptKind: 1, !entry: false
    at getDocumentRegistryEntry (c:\Users\{OMITTED}\.vscode\extensions\svelte.svelte-vscode-109.1.0\node_modules\typescript\lib\typescript.js:140976:11)
    at acquireOrUpdateDocument (c:\Users\{OMITTED}\.vscode\extensions\svelte.svelte-vscode-109.1.0\node_modules\typescript\lib\typescript.js:141006:32)
    at Object.updateDocumentWithKey (c:\Users\{OMITTED}\.vscode\extensions\svelte.svelte-vscode-109.1.0\node_modules\typescript\lib\typescript.js:140961:12)
    at getOrCreateSourceFileByPath (c:\Users\{OMITTED}\.vscode\extensions\svelte.svelte-vscode-109.1.0\node_modules\typescript\lib\typescript.js:149933:37)
    at Object.getOrCreateSourceFile (c:\Users\{OMITTED}\.vscode\extensions\svelte.svelte-vscode-109.1.0\node_modules\typescript\lib\typescript.js:149919:14)
    at c:\Users\{OMITTED}\.vscode\extensions\svelte.svelte-vscode-109.1.0\node_modules\typescript\lib\typescript.js:149840:42
    at Object.getSourceFileWithCache [as getSourceFile] (c:\Users\{OMITTED}\.vscode\extensions\svelte.svelte-vscode-109.1.0\node_modules\typescript\lib\typescript.js:124075:24)
    at findSourceFileWorker (c:\Users\{OMITTED}\.vscode\extensions\svelte.svelte-vscode-109.1.0\node_modules\typescript\lib\typescript.js:126536:23)
    at findSourceFile (c:\Users\{OMITTED}\.vscode\extensions\svelte.svelte-vscode-109.1.0\node_modules\typescript\lib\typescript.js:126452:20)
[...]

PR #2565, opened 2024-11-05
fix: allow JS snapshots within TS plugin

dummdidumm:
So far we've always assumed that a Svelte file is a TS file for simplicity, due to our default language (which was removed a long time ago) and also because IIRC there were issues with TS having its snapshot scriptKind being switched. That seems to be no longer the case, and so we can get better at properly analyzing whether or not this is a TS file, to allow JSDoc to take action when it's a JS file.

#2555

@jasonlyu123 do you remember why that scriptKind switch broke previously, but apparently doesn't anymore? I vaguely remember us having the same problem within the language server, but you PRd a change at some point stating that it's no longer a problem.

Also, is this a breaking change, because if someone has no `allowJS` in their config they would get errors?

### DIFF
--- a/packages/typescript-plugin/src/svelte-snapshots.ts
+++ b/packages/typescript-plugin/src/svelte-snapshots.ts
@@ -16,13 +16,18 @@ export class SvelteSnapshot {
         private svelteCode: string,
         private mapper: SourceMapper,
         private logger: Logger,
-        public readonly isTsFile: boolean
+        public isTsFile: boolean
     ) {}
 
-    update(svelteCode: string, mapper: SourceMapper) {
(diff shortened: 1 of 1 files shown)


jasonlyu123:
Yes, there is a `Debug.assert` for ScriptKind changes. TypeScript added a fix for it some time ago. The `Debug.assert` is still there, so it might still happen in some edge cases, but it shouldn't happen in most cases. #2538 is an example. In ts-plugin, you'll need file save to trigger updates, so for #2538 to happen in ts-plugin, it is edge case + edge case. Also, The more reliable reproduction for #2538 should only happen in module resolution "node16"+ because of the impliedNodeFormat patch but we didn't patch it in ts-plugin.

PR #2570, opened 2024-11-07
fix(vscode): let `linkedEditing` work when tag name contains `.`

### LINKED ISSUE #2569: VSCode: `editor.linkedEditing` does not work when the tag contains `.`
### Describe the bug

When a tag name contains a `.` character, `linkedEditing` does not work. While it's technically invalid HTML to have non-alphanumeric characters in a tag name, the Svelte compiler accepts this as valid.

`shadcn-svelte` and `bits-ui` makes heavy use of `.` in tag names as can be seen in docs: https://www.shadcn-svelte.com/docs/components/table

### Reproduction

Steps to Reproduce:

1. Take the following sample Svelte and paste it in a blank HTML file
```svelte
<Table.Header>
	<Table.Row>
		<Table.Cell>Item</Table.Cell>
		<Table.Cell>Amount</Table.Cell>
	</Table.Row>
</Table.Header>
```
2. Enable `editor.linkedEditing`
3. Attempt to edit `Table.Row` in the start or end tag
4. See that the matching tag does not get edited

### Expected behaviour

Linked editing should allow me to edit the corresponding tag even if it contains a `.`.

### System Info

- OS: Ubuntu 20.04
- IDE: VSCode

Svelte for VS Code extension

Related: #861

I am willing to contribute a fix for this. If I could get a code pointer or two, that would be great.

dyc3:
fixes #2569

Tested it manually on my machine and it works.

### DIFF
--- a/packages/language-server/src/plugins/html/HTMLPlugin.ts
+++ b/packages/language-server/src/plugins/html/HTMLPlugin.ts
@@ -333,7 +333,12 @@ export class HTMLPlugin
             return null;
         }
 
-        return { ranges };
+        // Note that `.` is excluded from the word pattern. This is intentional to support property access in Svelte component tags.
+        return {
+            ranges,
+            wordPattern:
+                '(-?\\d*\\.\\d\\w*)|([^\\`\\~\\!\\@\\#\\^\\&\\*\\(\\)\\=\\+\\[\\{\\]\\}\\\\\\|\\;\\:\\\'\\"\\,\\<\\>\\/\\s]+)'
+        };
     }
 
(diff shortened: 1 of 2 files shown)


jasonlyu123:
What I meant is to copy this regex to HTMLPlugin, tweak it and return it with the link editing result ranges. https://github.com/sveltejs/language-tools/blob/master/packages/language-server/src/plugins/html/HTMLPlugin.ts#L320. This way only the linked editing get the override and it won't affect other features.

dyc3:
Ah, my bad, I misunderstood. Will fix.

dyc3:
Should be good now. Tested it manually again and it works.

PR #2571, opened 2024-11-07
fix: hoist types related to `$props` rune if possible

dummdidumm:
This allows TypeScript to resolve the type more easily, especialy when in dts mode. The advantage is that now the type would be preserved as written, whereas without it the type would be inlined/infered, i.e. the interface that declares the props would not be kept

@jasonlyu123 does the approach make sense to you? Anything I might have missed?


jasonlyu123 (commented):
These two also cause some errors.

```ts
    interface Props {
        hi2: C;
    } 
 
    const a = 1
    type C = typeof a | '2' | '3';
```

```ts
    interface Props {
        hi2: C;
    } 
 
    const a = 1
    type C = typeof a | '2' | '3';
```

jasonlyu123:
Oops. Copied the wrong thing and I don't even remember what is the other case either 😆. Probably just a variant of the same issue I guess.

PR #2580, opened 2024-11-10
fix: data from $props does not need type

YpsilonTM:
The data type is not need because $props() is type with a data attribute. And like it was now was even wrong because props is not of type data its an attribute.


jasonlyu123:
We discussed whether the snippet should include explicit type annotation in https://github.com/sveltejs/language-tools/pull/203. But we think we should keep it. So we should fix the type issue here instead, and the one in layout as well. If you want to make it generate without the type annotation, you can create a separate PR with a config to control it.

dummdidumm (approved):

Issue #2556, opened 2024-11-02
Why is this line marked as error?


dm-de:
### Describe the bug

```
<script>
	let { array = [] } = $props()
	let array2 = $derived(array.filter(i => !i.hidden))
</script>
```

In this short example, filter is marked red
'array' is of type 'unknown'.

For example, when I use next code in TS playground, it will not mark array or filter as error. Expected errors are $props() and i. But not arary or filter.

```
let { array = [] } = $props()
array.filter(i => !i.hidden) 
```

My configuration:
installed using Vite (fresh installation)
but not Svelte kit
just JS Svelte 5 (newest)
or even TS (no difference)
newest Svelte extension for VS code
newest VS Code update

Such error come up with Svelte 5 first (Svelte 4 show no errors).
I hoped all the time that this will be fixed with final v5.

I played with checkJS - or ts-nocheck.
But this will hide any error.

![sv3](https://github.com/user-attachments/assets/2ac1f8f7-1443-403d-8b91-2e4bd2477564)

### Reproduction

```
<script>
	let { array = [] } = $props()
	let array2 = $derived(array.filter(i => !i.hidden))
</script>
```

### Expected behaviour

no error

### System Info

- OS: Windows
- IDE: VSCode

### Which package is the issue about?

Svelte for VS Code extension

dummdidumm:
You can't both strictly type check your code and expect no error in these positions. If you opt into more type checking, you have to provide the types, too. What we can do however is detect that you're creating an array-type fallback, and type that as `unknown[]`.

dm-de:
is fixed now, but what can we do, to auto-remove such errors?

all properties of an array/object are unknown - but it is annoying
![sv5](https://github.com/user-attachments/assets/d118baad-f61b-4f3e-9557-59499581517d)

dm-de:
I found other issue. This is not fixed:

![sv7](https://github.com/user-attachments/assets/9f1832f7-007b-4f06-9092-1db8bebb2bac)


```
<script>
	let {
		children,
	} = $props()
</script>

<div>
	{@render children?.()}
</div>
```


In my eyes, this line should also be error-free:

`{@render children()}`

dm-de:
Here is another error. Possible same like previous.

![sv6](https://github.com/user-attachments/assets/337b831b-07b7-47de-af4c-abf8cd04024e)

I found one way to hide it with:
`ontoggle = ()=>{},`

Is it possible to make it error free without extra function?

[...]

PR #2582, opened 2024-11-12
fix: fall back to `any` instead of `unknown` for untyped `$props`

dummdidumm:
The idea was to be conservative about it, but this causes problems for people not wanting strict type checking but only a basic form of it. Falling back to `any` is the more pragmatic choice

#2556

@jasonlyu123 any reservations against this?


jasonlyu123:
Ok with fallback to `any`. There seem to be many more people who got confused about it since the stable release is out 😅. We'll lose the ability to warn people who forgot to type the props. But at least for me that doesn't happen very often.

PR #2585, opened 2024-11-14
feat: TypeScript 5.7 support

jasonlyu123:
Draft until 5.7 stable is out. Most of the changes with tests are because of [Checks for Never-Initialized Variables](https://devblogs.microsoft.com/typescript/announcing-typescript-5-7-rc/#checks-for-never-initialized-variables). Might have to check if there is anything that false positive specific to svelte related to this change.


jasonlyu123:
I didn't find problems with the [Never-Initialized Variables](https://devblogs.microsoft.com/typescript/announcing-typescript-5-7-rc/#checks-for-never-initialized-variables). This should be ready.

dummdidumm (approved):
👍

Issue #1190, opened 2021-10-03
Add extension support for vscode web (github.dev)


GagnDeep:
Github.dev doesn't have syntax highlighting of svelte files by default. While it does allow installing extensions, but svelte for vscode shows its not compatible with vscode-web

![image](https://user-images.githubusercontent.com/35252411/135738610-a138963d-d7eb-4551-8934-a0d6d7bdb7e6.png)

While enviroments like [codesandbox](https://codesandbox.io) does support svelte syntax highlighting and other language features but they were using the old jamesbirtles.svelte-vscode extension to support svelte [source](https://github.com/codesandbox/codesandbox-client/tree/master/standalone-packages/vscode-extensions/out/extensions/jamesbirtles.svelte-vscode-0.7.1). But this extension is also not available in github.dev as it was removed from vscode marketplace

So is it possible to support the latest [svelte for vscode](https://marketplace.visualstudio.com/items?itemName=svelte.svelte-vscode) extension to work in web enviroments like [github.dev](https://github.dev/sveltejs/template)

dummdidumm:
This is a significant undertaking which requires many different things
- Add a bundle step to ship a browser and editor version (related to #139 ; I'd like to use rollup instead of webpack, if possible). This among other things needs to shim some node globals and alias node package imports to something that runs in the browser
- Adjust the code so that things that are dependent on the platform are refactored out into separate locations to better control what runs when
- Turn off some features like loading a `svelte.config.js` which is not possible because we can't load other scripts in a web context. I guess this is similar to the reduced feature set of untrusted workspaces (#1051)

More details can be found here: https://code.visualstudio.com/api/extension-guides/web-extensions

@asafamr did a POC of porting the language server to run in the monace editor, which is related: https://github.com/asafamr/monaco-svelte . Maybe he has some more insights.

asafamr:
I'm not really familiar with VSCode web extensions but from a quick look, it does seem related. Two points I can think of
* seems like you [can't load other modules](https://code.visualstudio.com/api/extension-guides/web-extensions#web-extension-main-file) - so some subset of used modules should be bundled in (svelte compiler, prettier + svelte plugin, postcss? preprocessor? typescript pp, svelte2tsx...?)

jasonlyu123:
[...]

PR #2611, opened 2024-11-30
feat: Allow embedded languages to utilise Svelte Intellisense in VSCode

BOJIT:
Hello! This PR modifies the behaviour of the Svelte VSCode extension so that it activates for URIs that don't use the `file://` scheme.
During standard extension usage, this makes no difference, as all Svelte files on disk will use the `file://` scheme in their URI. However, enabling this support allows other extensions to use Svelte as an embedded language and still take full advantage of all of the Svelte extension's Intellisense.

This is handled in VSCode by a mechanism called [Request Forwarding](https://code.visualstudio.com/api/language-extensions/embedded-languages#request-forwarding). In summary, VSCode will preprocess the custom language and create a virtual document that contains just the "Svelte" portion of the document. It can then invoke the Svelte extension on that virtual document.
The problem is that these virtual documents must use a unique URI scheme. As Svelte currently only activates on `file://` scheme URIs, these requests will be dropped.

### Example Use-Cases
- MDsveX syntax support: I've got a [draft extension](https://github.com/BOJIT/vscode-MDsveX) that adds MDsveX syntax support to VSCode.
- Custom languages built around Svelte. I know that [Evidence BI](https://github.com/evidence-dev/evidence/tree/next/packages/extension) maintain an extension that would benefit from Svelte intellisense/language features.

### Potential Problems
Various parts of the extension (document snapshots, Typescript plugin) track the `path` of the file, not the URI. The conversion from path to URI in `language-server/utils.ts` will assume the `file://` URI scheme when mapping between paths and URIs.

I am not familiar enough with the Svelte Language Server inner workings to fully know what the impact of supporting all URI schemes is. I've tested these modifications locally and the extension _seems_ to behave identically to the original extension, but I would appreciate some advice from the maintainers on the impact of this change!

I'm also not sure if there are any URI schemes that should be explicitly disallowed by the extension. For example, `http`/`https` URIs ending in `.svelte` should probably not be recorded in document snapshots, as these might appear in webviews?

### Next Steps
- My suggestion would be to only ever pass references to URIs in the language server, and only convert them into paths where speific APIs require it.
- Some of the Svelte-Kit specific Intellisense still doesn't work in virtual documents. Not sure where this gets tracked in the extension.

I'm hoping that this small change does not break any existing behaviour, as merging it will make any downstream embedded svelte extension's experience much better!

### DIFF
--- a/packages/language-server/src/utils.ts
+++ b/packages/language-server/src/utils.ts
@@ -28,13 +28,16 @@ export function clamp(num: number, min: number, max: number): number {
 
 export function urlToPath(stringUrl: string): string | null {
     const url = URI.parse(stringUrl);
-    if (url.scheme !== 'file') {
+    if (url.scheme === 'http' || url.scheme === 'https') {
         return null;
     }
     return url.fsPath.replace(/\\/g, '/');
 }
 
-export function pathToUrl(path: string) {
+export function pathToUrl(path: string, scheme?: string) {
+    const baseUri = URI.file(path);
+    if (scheme)
+        return baseUri.with({ scheme: scheme }).toString();
     return URI.file(path).toString();
 }
 
--- a/packages/language-server/src/plugins/typescript/DocumentSnapshot.ts
+++ b/packages/language-server/src/plugins/typescript/DocumentSnapshot.ts
@@ -266,7 +266,7 @@ function preprocessSvelteFile(document: Document, options: SvelteSnapshotOptions
 export class SvelteDocumentSnapshot implements DocumentSnapshot {
     private mapper?: DocumentMapper;
     private lineOffsets?: number[];
-    private url = pathToUrl(this.filePath);
+    private url = this.fileUrl;
 
     version = this.parent.version;
     isSvelte5Plus = Number(this.svelteVersion?.split('.')[0]) >= 5;
@@ -281,7 +281,11 @@ export class SvelteDocumentSnapshot implements DocumentSnapshot {
         private readonly exportedNames: IExportedNames,
         private readonly tsxMap?: EncodedSourceMap,
         private readonly htmlAst?: TemplateNode
-    ) {}
+    ) { }
+
+    get fileUrl() {
+        return this.parent.getURL() || '';
+    }
 
     get filePath() {
         return this.parent.getFilePath() || '';
--- a/packages/svelte-vscode/src/extension.ts
+++ b/packages/svelte-vscode/src/extension.ts
@@ -151,7 +151,7 @@ export function activateSvelteLanguageServer(context: ExtensionContext) {
     }
 
     const clientOptions: LanguageClientOptions = {
-        documentSelector: [{ scheme: 'file', language: 'svelte' }],
+        documentSelector: [{ scheme: '*', language: 'svelte' }],
         revealOutputChannelOn: RevealOutputChannelOn.Never,
         synchronize: {
             // TODO deprecated, rework upon next VS Code minimum version bump


packages/language-server/src/utils.ts
```diff
@@ -28,13 +28,16 @@ export function clamp(num: number, min: number, max: number): number {
...
     }
     return url.fsPath.replace(/\\/g, '/');
 }
 
-export function pathToUrl(path: string) {
+export function pathToUrl(path: string, scheme?: string) {
+    const baseUri = URI.file(path);
+    if (scheme)
+        return baseUri.with({ scheme: scheme }).toString();
     return URI.file(path).toString();
```
dummdidumm, line 41:
```suggestion
export function pathToUrl(path: string, scheme?: string) {
    const baseUri = URI.file(path);
    if (scheme) {
        return baseUri.with({ scheme: scheme }).toString();
    }
    return baseUri.toString();
```

dummdidumm (commented):
Thanks for the detailed PR description! I'm ok to give this a try under the assumption that we're not going to jump through countless hoops to make it work 100% correctly for all use cases.

@jasonlyu123 any objections? Anything you think that might need to be adjusted / could go wrong?

BOJIT:
Ah, good point! Didn't test untitled documents... Will have another look this week and check a few additional cases.

I imagine that I'll may need to raise some future PRs to fully support every embedded language forwarding feature, but the priority in this PR is to ensure that it doesn't break any existing Svelte extension features!

I'll do some further testing :+1:

BOJIT:
Okay, so after a bit more testing, I've realised that dealing with the untitled document.

The Untitled document is handled with a URI scheme of `untitled`. This means that with this PR, the Svelte  language server is activated. The positive of this is that Untitled documents get completion/language server features, which they wouldn't otherwise. But unfortunately, any Svelte language features that require a filesystem-related call will crash. Some good notes [here](https://github.com/Microsoft/vscode-docs/issues/1565).

I couldn't find any official list in VSCode, but the URI schemes that could potentially be a problem are:
- `untitled`
- `vscode`
- `http`
- `https`
- `git`

Unfortunately, the `scheme` field in the [DocumentSelector](https://code.visualstudio.com/api/references/document-selector#document-scheme) doesn't seem to allow more complex globs/exclusion patterns.

Best as I can tell, there are three options for properly supporting non-vdoc URIs:

1. Add a specific URI scheme that extensions can use: e.g. 
```
documentSelector: [
    { scheme: 'file', language: 'svelte' },
    { scheme: 'embedded-svelte', language: 'svelte' },
],
```

Straightforward, but does limit only one extension to register a scheme provider at a time, so not ideal.

2. Modify the VSCode language client document selector to support more complex exclusion patterns for schemes.
Potentially doable, I need to look more into what would be involved for this.

3. Make the Svelte Language Server handle non-file URIs safely.
This is probably the most work, but does have some notable advantages.
- Full intellisense for "untitled" documents in VSCode (and possibly Git virtual docs too)
- Any editor can take advantage of request forwarding, not just VSCode.
- Ability to get "partial" Svelte extension support in [VSCode for web](https://code.visualstudio.com/docs/editor/vscode-web#_extensions). Currently the Svelte extension cannot be used there.

I'm happy to start exploring the last option and see how big of a change this would be. Let me know your thoughts on this.

dummdidumm:
Sure, go ahead exploring 👍 

> making the language tools available on the web would be awesome, though I'm not sure how connected it really is to this work (see #1190 for more).

jasonlyu123:
I think it isn't the problem that TypeScript only handles physical files. The problem with completion is the inconsistency of the document URI schema. What is sent back in `completionItem.data.uri` is `file:` but the document is `untitled:`.

Issue #2534, opened 2024-10-12
Deleting or pasting code sometimes causes messed-up highlighting and broken rename until reloading window


CherryDT:
### Describe the bug

Sometimes when deleting or pasting code, all the syntax highlighting gets messed up (things get highlighted with different colors in the middle of a word, etc.), the "rename" operation doesn't work correctly anymore and overwrites random other things in the file with the new name instead of changing the places where the original name was used. This state gets fixed only by reloading the window.

It started only recently, maybe a week or so ago.

### Reproduction

Don't know - it randomly happens when pasting code.

### Expected behaviour

Normal syntax highlighting, normal "rename" behavior

### System Info

- OS: Windows with remote WSL1
- IDE: VSCode 1.94.1

### Which package is the issue about?

Svelte for VS Code extension

### Additional Information, eg. Screenshots

![image](https://github.com/user-attachments/assets/e83cb51c-8050-476b-907b-028188b93deb)

jasonlyu123:
We'll need an actual reproduction to investigate. So please at least try to narrow down what the issue might be. Does this happen in a specific file? If so, does copying this file to any empty svelteKit project reproduce the problem? Is there a file in the directory with a similar name? Or maybe there is anything you did before you pasted any code? Like renaming the file?

CherryDT:
I realized it actually happens more frequently when _deleting_ code rather than pasting.

It's very hard for me reproduce reliably. It happens randomly on various files and various different areas of code, in different projects, approximately 1-2 times per hour. I was not renaming or creating or deleting files, just working on existing files.

I looked into the output pane a bit and I found two things that seemed relevant:

1: Under Svelte, there was this error:
```
Error: Debug Failure. False expression: Script kind should match provided ScriptKind:1 and sourceFile.scriptKind: 3, !entry: false
    at getDocumentRegistryEntry (/home/david/.vscode-server/extensions/svelte.svelte-vscode-109.1.0/node_modules/typescript/lib/typescript.js:140976:11)
    at releaseDocumentWithKey (/home/david/.vscode-server/extensions/svelte.svelte-vscode-109.1.0/node_modules/typescript/lib/typescript.js:141070:19)
    at registry.releaseDocumentWithKey (/home/david/.vscode-server/extensions/svelte.svelte-vscode-109.1.0/node_modules/svelte-language-server/dist/src/plugins/typescript/service.js:989:13)
[...]

PR #2635, opened 2024-12-23
fix: prevent error when the script tag is removed in nodenext projects

jasonlyu123:
Part of #2538. The problem is becuase we're patching `impliedNodeFormat` in module resolution. When we can't even parse the script tag the module resolution won't happen and the `impliedNodeFormat` won't be patched. This causes TypeScript to try to clear the old source file and trigger the Debug.assert.

I also found another situation that would trigger the same `Debug.assert` but I haven't found a way to fix it.  Might have to open an issue on TypeScript for that one. At least we can get this easier one fixed. 

I also can't figure out the exact condition that triggers https://github.com/sveltejs/language-tools/issues/2534#issuecomment-2408603785. From the typescript source code, it makes no sense this would happen. The error fixed in this PR might cause the document registry to update but not the language service. So maybe https://github.com/sveltejs/language-tools/issues/2534#issuecomment-2408603785 is caused by the weird state after the error but I still can't find the exact situation.


jasonlyu123:
Oh, that was embarrassing 😅. I thought I fixed it. Back to the drawing board.

dummdidumm (approved):

PR #2667, opened 2025-01-16
fix: disambiguate `render` in module script

paoloricciuti:
One day i'll learn how to properly generate the tests...sorry if i fumbled.

This fixes a problem that you get when you import or create a variable/function named `render` in the module...since we create a `render` function ourselves it conflicts with it giving a confusing error to the user.

It also just occured to me that maybe we can even `rename` the function to something less usual so that it has less chance of conflicting with the user (now that we have a global variable with the name could even be a random identifier so that even if by any change it conflicts with the user it should go away the next restart of language server).

I think this should work fine but I hope the name (or the length of the name) was not referenced somewhere else so please have an extra eye for this things.

### DIFF
--- a/packages/svelte2tsx/src/svelte2tsx/addComponentExport.ts
+++ b/packages/svelte2tsx/src/svelte2tsx/addComponentExport.ts
@@ -25,6 +25,7 @@ export interface AddComponentExportPara {
     usesSlots: boolean;
     isSvelte5: boolean;
     noSvelteComponentTyped?: boolean;
+    renderName: string;
 }
 
 /**
@@ -54,7 +55,8 @@ function addGenericsComponentExport({
     generics,
     usesSlots,
     isSvelte5,
-    noSvelteComponentTyped
+    noSvelteComponentTyped,
+    renderName
 }: AddComponentExportPara) {
     const genericsDef = generics.toDefinitionString();
     const genericsRef = generics.toReferencesString();
@@ -71,13 +73,13 @@ function addGenericsComponentExport({
     let statement = `
 class __sveltets_Render${genericsDef} {
     props() {
-        return ${props(true, canHaveAnyProp, exportedNames, `render${genericsRef}()`)}.props;
+        return ${props(true, canHaveAnyProp, exportedNames, `${renderName}${genericsRef}()`)}.props;
     }
     events() {
-        return ${_events(events.hasStrictEvents() || exportedNames.usesRunes(), `render${genericsRef}()`)}.events;
+        return ${_events(events.hasStrictEvents() || exportedNames.usesRunes(), `${renderName}${genericsRef}()`)}.events;
     }
     slots() {
-        return render${genericsRef}().slots;
+        return ${renderName}${genericsRef}().slots;
     }
 `;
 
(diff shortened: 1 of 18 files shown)


paoloricciuti:
For the moment i've run the tests in v5...not sure how to generate the v4 version

paoloricciuti:
Oh small note: this technically "over fires"...since it's looking for all the identifiers in the module tag it could rename the render function even for `x.render` for example...I don't think is worth it to be precise here since it's not a big deal (the user will never see this) and it's not a performance hit and better be over firing than specifying every case and maybe missing something. But i can fix it if we want.

dummdidumm:
what about just renaming it to `$$render` instead? Starts with a disallowed character, so no possibility of clashes. Will mean we gotta adjust basically every test but that can happen with one test-folder-level search&replace I think.

paoloricciuti:
> what about just renaming it to `$$render` instead? Starts with a disallowed character, so no possibility of clashes. Will mean we gotta adjust basically every test but that can happen with one test-folder-level search&replace I think.

Ohhh i like this...let me do it

paoloricciuti:
@dummdidumm what do you think of still using a variable name since i've already did the work of replacing it everywhere? this way it could be easier to do if we later want to change it for some reason. Otherwise i can just revert everything and manually change them

dummdidumm:
We can keep it, but there's no point in passing it through functions, I would just do `export const RENDER = '$$render'` and use that

paoloricciuti:
> We can keep it, but there's no point in passing it through functions, I would just do `export const RENDER = '$$render'` and use that

Uh yeah...i think there's some fixture that needs to be updated for the language server too (and also sourcemaps for svelte 4)...could you guide me on how to rerecord them?

packages/svelte2tsx/index.d.ts
```diff
@@ -195,3 +195,5 @@ export namespace InternalHelpers {
         paramsPath: string;
     }
 }
+
+export const RENDER_NAME:string;
```
dummdidumm, line 199: Can we expose this from a sub module, or via the helpers object that is not subject to semver? Wanna make sure that the "public public" API stays clean.

dummdidumm (commented):
Good idea of making it an export! Though it should not be exposed directly, rather through this helper object; also because this requires svelte2tsx to be a certain version we need to bump the svelte2tsx version within the svelte2tsx package to the version it's gonna be after the patch is released. That way the language-server etc packages will force a dependency update.

packages/svelte2tsx/index.d.ts
paoloricciuti: Oh yeah better...just did it in a new commit

packages/svelte2tsx/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "svelte2tsx",
-    "version": "0.7.25",
+    "version": "0.7.26",
```
dummdidumm, line 3: It's actually going to be this
```suggestion
    "version": "0.7.35",
```
It's confusing, because we bump those by hand only from time to time, but at deploy time it is basically ignored in favor of what the release name is.

dummdidumm:
Thank you!

PR #2668, opened 2025-01-17
fix: take other snippets into account when checking for hoistability

### LINKED ISSUE #2664: Language Server: False "Cannot find name" error in nested snippet renders
### Describe the bug

The language server incorrectly reports "Cannot find name" error when using `{@render ...}` inside snippet definitions, even though the snippet function is properly defined under specific conditions.

### Reproduction

```html
<script module lang="ts">
</script>

<script lang="ts">
  function foo() { console.log("hello") }
</script>

{#snippet test()}
  {@render renderButton()}
{/snippet}

{#snippet renderButton()}
  <button type="button" onclick={foo}>test button</button>
{/snippet}
```

Image

### Expected behaviour

No "Cannot find name" error should be reported for properly defined snippet functions.

Image

### System Info

- OS: Windows 11
- IDE: VSCode (Ver. 1.96.3)
- Svelte for VS Code: 109.5.2

svelte-language-server

I have verified this issue across multiple versions:
- The error is present in versions 109.5.2 and 109.5.0
- The error does not exist in version 109.4.0. The Expected behaviour screenshot was taken using this version.

The error can be eliminated by either:
- Removing the `onclick={foo}` handler
- Removing the `<script module lang="ts"></script>` tag

Additionally, `{@render renderButton()}` functions correctly when used outside of snippet definitions.

paoloricciuti:
Closes #2664

Dunno if there's a better way...this basically includes the snippets in the disallowed values but it has to do it until is stable to prevent situations where the snippet reference a snippet which is found as non hoistable after the check.

### DIFF
--- a/packages/svelte2tsx/src/htmlxtojsx_v2/index.ts
+++ b/packages/svelte2tsx/src/htmlxtojsx_v2/index.ts
@@ -54,7 +54,7 @@ export interface TemplateProcessResult {
     scriptTag: BaseNode;
     moduleScriptTag: BaseNode;
     /** Start/end positions of snippets that should be moved to the instance script or possibly even module script */
-    rootSnippets: Array<[start: number, end: number, globals: Map<string, any>]>;
+    rootSnippets: Array<[start: number, end: number, globals: Map<string, any>, string]>;
     /** To be added later as a comment on the default class export */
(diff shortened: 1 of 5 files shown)


dummdidumm (approved):
👍

Issue #2675, opened 2025-01-31
unknown css-property dominant-baseline


nstringham:
### Describe the bug

Svelte Check doesn't know that `dominant-baseline` is a CSS property

### Reproduction

```svelte
<style>
	text {
		dominant-baseline: central;
	}
</style>
```

### Expected behaviour

no warning should be given

### System Info

- OS: Windows
- IDE: VSCode

### Which package is the issue about?

svelte-check

### Additional Information, eg. Screenshots

https://developer.mozilla.org/en-US/docs/Web/CSS/dominant-baseline

I believe this bug is very similar to #1661

Issue #2388, opened 2024-06-03
`svelte-check`: Warn: Unknown property: 'r' (scss)


threema-gian:
### Describe the bug

When I use the `r` property (SVG circle radius) in CSS, `svelte-check` displays the following warning: `Warn: Unknown property: 'r' (scss)`.

For reference:

- [https://svgwg.org/svg2-draft/geometry.html#R](https://svgwg.org/svg2-draft/geometry.html#R).
- [https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/r](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/r) (note the following info box: "**Note:** Starting with SVG2, `r` is a _Geometry Property_ meaning this attribute can also be used as a CSS property for circles.").

Possibly related issues:

- [https://github.com/sveltejs/language-tools/issues/977](https://github.com/sveltejs/language-tools/issues/977).

### Reproduction

Reproducer here (built with the latest Vite + Svelte starter at the time of writing): [https://github.com/threema-gian/svelte-check-unknown-property-reproducer](https://github.com/threema-gian/svelte-check-unknown-property-reproducer).

### Expected behaviour

Using the following CSS block should not produce any `svelte-check` warnings:

```css
svg circle {
  r: 5px;
}
```

### System Info

- OS: Fedora 40
- IDE: VSCode
- Package:
  - `svelte`: 4.2.12
  - `svelte-check`: 3.6.6
  - `typescript`: 5.3.3
  - ...others, probably not relevant

Relevant `npm` script:
- `"lint:svelte": "svelte-check --workspace src/app/ --no-tsconfig --compiler-warnings \"a11y-click-events-have-key-events:ignore\""`

### Which package is the issue about?

svelte-check

jasonlyu123:
This would need support from upstream vscode-css-languageservice which pulls data from MDN. And MDN also doesn't have a dedicated page for these geometry properties.

phobetron:
@jasonlyu123 It seems these properties should be registered in current versions of `svelte-check`, based on the version numbers/dates of `vscode-css-languageservice` and `vscode-custom-data`,  but they're still being reported as unknown. We can see this specific property registered [here](https://github.com/microsoft/vscode-custom-data/blob/c008a80baa3c6ea9d6757d2640eaab215b28f9a6/web-data/css/css-schema.json#L21302). 

[...]

PR #2677, opened 2025-02-03
chore: bump vscode-html/css-language-service

jasonlyu123:
#2675 #2388. For svelte-check. The extension should already be shipping with these versions.


AlbertMarashi (approved):

AlbertMarashi:
Thanks. Would be nice to get these `rx`, `ry` and other SVG warnings going away soon

PR #2694, opened 2025-02-24
fix: properly transform `$props.id` when `$props` is assigned to `props`

paoloricciuti:
This is a bit of weird one and I'm not sure there's a better way to handle this. The problem is that when you are doing this

```svelte
<script>
    let props = $props();
    let id = $props.id();
</script>
```

the second `$props` identifier was not considered a rune because it's not an immediate function call. There's a catch tho: even if we check that the parent is a call expression and the text is `$props.id` we can't rule that out immediately because it depends if the `props` variable comes from the assignment of `$props()` or not. But the scope we have here is far less powerful than the one in the svelte compiler. So i had to do this weird dance where i still add the possible store in the array but i also save the fact that comes from `$props.id`...then while walking if i encounter an identifier `props` which is in a variable declaration i check if the variable declaration initializer is `$props()`.

It should cover all the edge cases but it's a bit convoluted. Also i would've loved to add some comments here and there but everytime i open that file my vscode crashes 😳

### DIFF
--- a/packages/svelte2tsx/src/svelte2tsx/processInstanceScriptContent.ts
+++ b/packages/svelte2tsx/src/svelte2tsx/processInstanceScriptContent.ts
@@ -93,6 +93,7 @@ export function processInstanceScriptContent(
     let scope = new Scope();
     const rootScope = scope;
 
+    //track is the variable declared as `props` comes from `$props()`
     let isPropsDeclarationRune = false;
 
     const pushScope = () => (scope = new Scope(scope));
@@ -130,6 +131,7 @@ export function processInstanceScriptContent(
             return;
         }
 
+        //if we are in a variable declaration and the identifier is `props` we check the initializer
         if (
             ident.text === 'props' &&
             variableDeclarationNode &&
@@ -327,6 +329,7 @@ export function processInstanceScriptContent(
 
     //resolve stores
     if (isPropsDeclarationRune) {
+        //we filter out every pendingStore resolution that `isPropsId` if the variable names `props` comes from `$props()`
         pendingStoreResolutions = pendingStoreResolutions.filter(({ isPropsId }) => !isPropsId);
     }
     pendingStoreResolutions.map(resolveStore);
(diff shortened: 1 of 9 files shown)


packages/svelte2tsx/src/svelte2tsx/processInstanceScriptContent.ts
```diff
@@ -82,13 +83,18 @@ export function processInstanceScriptContent(
...
+
     //track $store variables since we are only supposed to give top level scopes special treatment, and users can declare $blah variables at higher scopes
     //which prevents us just changing all instances of Identity that start with $
-    const pendingStoreResolutions: PendingStoreResolution[] = [];
+    let pendingStoreResolutions: PendingStoreResolution[] = [];
 
     let scope = new Scope();
     const rootScope = scope;
 
+    let isPropsDeclarationRune = false;
```
paoloricciuti, line 96:
```suggestion
    //track is the variable declared as `props` comes from `$props()`
    let isPropsDeclarationRune = false;
```

packages/svelte2tsx/src/svelte2tsx/processInstanceScriptContent.ts
```diff
@@ -124,6 +130,16 @@ export function processInstanceScriptContent(
             return;
         }
 
+        if (
```
paoloricciuti, line 133:
```suggestion
        //if we are in a variable declaration and the identifier is `props` we check the initializer
        if (
```

packages/svelte2tsx/src/svelte2tsx/processInstanceScriptContent.ts
```diff
@@ -295,6 +326,9 @@ export function processInstanceScriptContent(
     tsAst.forEachChild((n) => walk(n, tsAst));
 
     //resolve stores
+    if (isPropsDeclarationRune) {
+        pendingStoreResolutions = pendingStoreResolutions.filter(({ isPropsId }) => !isPropsId);
```
paoloricciuti, line 330:
```suggestion
        //we filter out every pendingStore resolution that `isPropsId` if the variable names `props` comes from `$props()`
        pendingStoreResolutions = pendingStoreResolutions.filter(({ isPropsId }) => !isPropsId);
```

dummdidumm (approved):
ah yeah the store autosubscription syntax really bites us there 😅

Issue #2031, opened 2023-05-26
Replace mocha with Vitest


dummdidumm:
- adjust with Svelte(Kit) where that's used, too
- Vitest can deal with ESM better, which will become more and more relevant (there are some workarounds currently in the tests to not having to import ESM Svelte)

ghostdevv:
Happy to help with this one if you like

dummdidumm:
Sure!

PR #2706, opened 2025-03-08
(deps) Replace `pascal-case` with `change-case` (Maintained & Lighter)

jycouet:
This PR replaces the unmaintained `pascal-case` package with its recommended successor, `change-case`.

#### **Why?**
- `pascal-case@3.1.1` is outdated, and `4.0.0` is deprecated.
- The author suggests migrating to `change-case`.
- `change-case` has **0 dependencies** and a smaller total footprint.

#### **Comparison:**
| Package       | Size  | Dependencies | Total Size |
|--------------|-------|-------------|------------|
| pascal-case  | 14KB  | 2 (`no-case` 25KB, `lower-case` 17KB) | **56KB** |
| change-case  | 35KB  | 0 | **35KB** |

---

✅ **Net win**: Fewer dependencies, lower size, and a more future-proof solution.

### DIFF
--- a/packages/svelte2tsx/package.json
+++ b/packages/svelte2tsx/package.json
@@ -66,6 +66,6 @@
     ],
     "dependencies": {
         "dedent-js": "^1.0.1",
-        "pascal-case": "^3.1.1"
+        "change-case": "^5.4.4"
     }
 }
--- a/packages/svelte2tsx/rollup.config.mjs
+++ b/packages/svelte2tsx/rollup.config.mjs
@@ -107,7 +107,7 @@ export default [
             'svelte',
             'svelte/compiler',
             'dedent-js',
-            'pascal-case'
+            'change-case'
         ]
     }
 ];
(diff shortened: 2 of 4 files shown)


jasonlyu123:
The CI is probably broken because the default environment has changed. It probably failed in master, too. I'll try to fix it later.

jycouet:
Hooo okay.
Thank you for the info 👍

jasonlyu123:
Weird... The problem is because change-case is an ESM libary. For some odd reason, the test breaks as long as you import one ESM libary.  It probably has something to do with the loader API mocha depends on to run TypeScript tests.

jycouet:
I'm willing to work on this if you are ok with it :)

jasonlyu123:
Sure. I haven't figure out a solution or found a upstream issue yet.

jycouet:
I'm turning and turning and turning, but can't find a solution!
Locally, I don't get the error, you manage to have it ? 
You think it's a `mocha` issue only ? Or the esm/cjs issue will happen also for consumer ? 
There is a reason for `svelte2tsx` to be `"type": commonjs` ?

dummdidumm:
FWIW we have a "would be nice" ticket to migrate from `mocha` to `vitest` (#2031), maybe that solves that issue (but it obviously is quite the undertaking)

jycouet:
Let's see tomorrow if I can start something. 🫠

jycouet:
Close in favor of https://github.com/sveltejs/language-tools/pull/2842

PR #2711, opened 2025-03-11
Tentative Vitest

jycouet:
In order to have https://github.com/sveltejs/language-tools/pull/2706 we probably need to go to `vitest` and manage well `ESM`.

Also linked with this issue: https://github.com/sveltejs/language-tools/issues/2031

---

Just started small to see if it's the good direction.
svelte2tsx test script is calling `mocha` & `vitest` and when tests are ok in vitest, I `skip` them in mocha.
So that the diff is approachable.

What do you think ?

### DIFF
--- a/packages/svelte2tsx/package.json
+++ b/packages/svelte2tsx/package.json
@@ -40,7 +40,8 @@
         "svelte": "~4.2.19",
         "tiny-glob": "^0.2.6",
         "tslib": "^2.4.0",
-        "typescript": "^5.8.2"
+        "typescript": "^5.8.2",
+        "vitest": "^3.0.8"
     },
     "peerDependencies": {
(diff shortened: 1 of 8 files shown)


dummdidumm:
Don't fully understand the skip thing - I don't have any opinion on how you wanna approach this, the end result of using Vitest instead of Mocha to run the tests (without adjusting any of the tests, ideally) is what counts.

jycouet:
_Forget about the skip thing 😅_

Here is what I have locally atm: 
image

The `346` are green! 🎉

dummdidumm:
Seems like this for some reason made the language server tests fail

jycouet:
Going to close this one 👍 
The other one looks in a better shape 🚀

Issue #2727, opened 2025-03-30
Generic type doesn't work with `extends`


rChaoz:
### Describe the bug

When used in the `extends` clause of the props interface, Svelte typechecking cannot find the generic type.

### Reproduction

Paste this code in an IDE with the Svelte plugin, or just run `sv check`:

```svelte
<script lang="ts" generics="Data">
    interface Props extends Data {
        id: string
    }

    let {}: Props = $props()
</script>
```

### Expected behaviour

This should work without errors, but instead:

![Image](https://github.com/user-attachments/assets/8afa8354-a31f-4ba6-87ad-ab67269e9f2e)

`sv check` output:

```svelte
Error: Cannot find name 'Data'. (ts)
<script lang="ts" generics="Data">
    interface Props extends Data {                                                                                                                                                                                               
        id: string
```

Note that actually using the `Props` interface is required, the error does not appear if the interface is unused (weird?). Might be some acorn-typescript shenanigans.

### Workaround

Use `type` instead of `interface`, this works correctly:

```ts
type Props = Data & {
    id: string
}
```

### Which package is the issue about?

No idea

Issue #2671, opened 2025-01-21
Error referencing properties of destructured variables


jhwz:
### Describe the bug

Follow up to https://github.com/sveltejs/language-tools/issues/2659

When referencing properties of destructured variables in a type, it will show the error `Cannot find name '<variable>'` in both svelte-check and the language server.

Workaround is to do `(typeof <variable>)["<property>"]` instead of `typeof <variable>.<property>`.

### Reproduction

```svelte
<script lang="ts">
	import type { PageData } from './$types.js';

	type Props = {
		data: PageData;
	};
	let { data }: Props = $props();

	type Fails = typeof data.cfg; // error here: Cannot find name 'data'
	type Okay = (typeof data)['cfg'];
</script>
```

### Expected behaviour

Shouldn't throw a type error.

### System Info

- OS: MacOS
- IDE: VSCode

### Which package is the issue about?

svelte-language-server, svelte-check

PR #2740, opened 2025-04-15
fix: various hoistability check

jasonlyu123:
#2671 #2727

The namespace fixes are mostly a drive-by when I was checking what exactly is allowed to be `ts.QualifiedName` in a type. I don't think many people are using it 🤣.


dummdidumm:
Can you also add a test where `extends ...` _will_ be hoisted? otherwise good to go 👍

dummdidumm (approved):
👍

Issue #2742, opened 2025-04-16
Component suggestions working only at certain lines


IcyFoxe:
### Describe the bug

Hello, I found this strange bug, where Svelte component suggestions work only at certain lines.
Please look at this recording:

https://github.com/user-attachments/assets/cba02f5b-7307-4b22-a7b6-5f5374c330ea

At first I thought I had some configuration issues, so I went literally file to file comparing to the configuration of freshly initialized Svelte project, but considering I have almost exactly the same configuration settings, nothing did help. Only after that I've realized, that suggestion still work, just not on every line...

### Reproduction

After trying to replicate it, I found a hint to what might be causing the issue:

- initialize new Svelte app using `npx sv create my-app`
- select these settings: minimal, typescript, prettier + eslint + adapter
- in `src/components` create `Button.svelte` component
- in `.prettierrc` add `"svelteSortOrder": "markup-options-scripts-styles",` <-- this seems to be causing the problem
- in `src/routes/+page.svelte` add `script` UNDER the html, add `div` with `span` in it, and start importing button several times, just like in the video, to see the issue appear

### Expected behaviour

Import suggestion should appear appear on any line.

### System Info

- OS: Windows 10
- IDE: VSCode

jasonlyu123:
I can't reproduce this. It probably requires other conditions that aren't shown in your screenshot. Can you try removing some code from the file to narrow down a minimal reproducible example?

IcyFoxe:
Windows 10, VSCode, extensions: ESLint, GitLens, Prettier, Svelte for VS Code
Here's how I can reproduce it easily:
- `npx sv create`
- SvelteKit minimal
- TypeScript
- prettier
- npm

Next:
- add `"svelteSortOrder": "markup-options-scripts-styles",` to `.prettierrc`
- create `src/components/Button.svelte` (no need to add anything to it
- in `src/routes/+page.svelte` copy this:

```svelte
<Button />
<h1>Welcome to SvelteKit</h1>
<p>Visit <a href="https://svelte.dev/docs/kit">svelte.dev/docs/kit</a> to read the documentation</p>
<div>
  <span>test</span>
</div>

<script lang="ts">
  import Button from '../components/Button.svelte';
</script>
```

And finally, now try typing `<Button...` after the `<span>` element. It will not autocomplete the button component, or even some other component.

What seems to be causing the issue is the Svelte sort order along with having at least one component already imported.

@jasonlyu123 Let me know if you can replicate it now.

[...]

PR #2750, opened 2025-04-30
fix: prevent error with unclosed tag followed by LF or end of file

jasonlyu123:
#2742 

The reason that it happens to LF but not CRLF is that CRLF has two characters, so it doesn't touch the range of the start tag name, and therefore won't trigger the magic string error. This should also improve auto-import performance a bit because TypeScript only needs to compute auto-imports for Button and not a global auto-import.


dummdidumm (approved):
👍

Issue #2738, opened 2025-04-14
Svelte LSP crashes due to out of memory


MrWaip:
### Describe the bug

Hi!

At my work in our monolithic project the svelte extension started to crash on startup. This doesn't happen on version 108.5.2, so we locked it. In fact, VS Code turns into a primitive text editor.

I just checked, the problem is reproduced on version 109.5.4.
The project uses Svelte 4.2.18

```
Debugger listening on ws://127.0.0.1:9229/cb55dc39-8085-4658-803d-d836a3c4b77a
For help, see: https://nodejs.org/en/docs/inspector
Initialize language server at  file:///Users/mrwaip/code/project
Initialize new ts service at  /Users/mrwaip/code/project/tsconfig.json
Trying to load configs for /Users/mrwaip/code/project
Loaded config at  /Users/mrwaip/code/project/svelte.config.js
SnapshotManager File Statistics:
Project files: 13363
Svelte files: 3729
From node_modules: 0
Total: 13363
SnapshotManager File Statistics:
Project files: 13363
Svelte files: 3729
From node_modules: 2938
Total: 16608

<--- Last few GCs --->

[64316:0x108002e0000]    91020 ms: Mark-Compact 4065.1 (4074.8) -> 4063.5 (4074.8) MB, pooled: 2 MB, 617.17 / 0.00 ms  (average mu = 0.129, current mu = 0.051) allocation failure; scavenge might not succeed
[64316:0x108002e0000]    91884 ms: Mark-Compact 4065.5 (4074.8) -> 4063.7 (4075.1) MB, pooled: 2 MB, 857.50 / 0.00 ms  (average mu = 0.061, current mu = 0.007) allocation failure; scavenge might not succeed


<--- JS stacktrace --->

FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
----- Native stack trace -----

 1: 0x11503c0bc node::OnFatalError(char const*, char const*) [/Applications/Visual Studio Code.app/Contents/Frameworks/Electron Framework.framework/Versions/A/Electron Framework]
 2: 0x10f496af4 node::GetEnvironmentIsolateData(node::Environment*) [/Applications/Visual Studio Code.app/Contents/Frameworks/Electron Framework.framework/Versions/A/Electron Framework]
[...]

PR #2754, opened 2025-05-02
fix: invalidate module resolution cache when higher priority file is added

jasonlyu123:
While debugging #2738, I found that the memory usage of svelte-language-server is much higher than that of TSServer when there are many files. It turns out that the way we store the mapping of "failed lookup path" to "containing file" uses a lot of memory. I guess path normalisation might have prevented V8 from optimising the string memory usage, so the same path has multiple instances in the `failedPathToContainingFile` map.

In an empty SvelteKit project that imports `googleapis`, which has 800+ `.d.ts` files and takes around 80 MB, the memory usage went down from around 1.4GB to 1 GB. It's still higher than TServer, but it's a lot closer. 

圖片
圖片

Also, a bit embarrassing that the old way I wrote also doesn't completely invalidate the cache, and the test is wrong 😅.

### DIFF
--- a/packages/language-server/src/plugins/typescript/module-loader.ts
+++ b/packages/language-server/src/plugins/typescript/module-loader.ts
@@ -164,8 +164,13 @@ export function createSvelteModuleLoader(
     >();
 
     const impliedNodeFormatResolver = new ImpliedNodeFormatResolver(tsSystem);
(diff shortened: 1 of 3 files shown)


dummdidumm (approved):
good investigation!

Issue #2697, opened 2025-02-27
Bug: Duplicate "Go to Definition" Results for Components in Svelte 5 in Neovim


urboifox:
### Describe the bug

When using "Go to Definition" (gd in Neovim) on components in the markup section of a Svelte 5 file, the LSP returns duplicate definition locations. This issue does not occur in Svelte 3 or 4 projects, and is specific to component definitions in the markup (not in script tags).
I'm using the language server through [mason](https://github.com/williamboman/mason.nvim)

### Example Code
```svelte
<script lang="ts">
    import MetricCard from "$lib/features/charts/components/metric-card.svelte";
</script>

<MetricCard /> <!-- Triggering "Go to Definition" here shows duplicate results -->
```
### Reproduction

1. Open a Svelte 5 file containing component imports and usage in neovim
2. Position cursor on a component name in the markup
3. Trigger "Go to Definition"
4. Observe that Telescope (or any LSP client) shows duplicate definition locations

### Expected behaviour

Single definition result pointing to the component's source file

### System Info

- OS: [Linux]
- IDE: [Neovim]

### Environment

- Svelte Language Server version: [0.17.10]
- LSP Client: built-in Neovim LSP

### Which package is the issue about?

svelte-language-server

### Additional Information, eg. Screenshots

![Image](https://github.com/user-attachments/assets/2d506bfe-1543-489b-9a09-63a5014186f7)

- This issue is specific to Svelte 5; the same setup works correctly with Svelte 3 and 4 projects
- The duplicate definitions appear only for components in markup, not for variables or imports in script tags
- The LSP response contains duplicate entries with identical targetUri and range information

dummdidumm:
I'm not able to reproduce this using VS Code + triggering "go to definition" and then looking at what our LSP returns. The relevant request triggered is [here](https://github.com/sveltejs/language-tools/blob/master/packages/language-server/src/plugins/typescript/TypeScriptPlugin.ts#L394). My guess is that for whatever reason the ["see that this is within generated code and therefore filter it out" logic](https://github.com/sveltejs/language-tools/blob/master/packages/language-server/src/plugins/typescript/TypeScriptPlugin.ts#L418) somehow isn't triggered though I have no idea why that would be.

urboifox:
Thanks for responding
I tried VS Code, it works perfectly, no issues.
[...]

PR #2759, opened 2025-05-14
fix: deduplicate definition for rune-mode components

jasonlyu123:
#2697


packages/svelte2tsx/src/svelte2tsx/addComponentExport.ts
```diff
@@ -260,7 +260,9 @@ declare function $$__sveltets_2_isomorphic_component<
             if (exportedNames.usesRunes() && !usesSlots && !events.hasEvents()) {
                 statement =
                     `\n${doc}const ${componentName} = __sveltets_2_fn_component(${internalHelpers.renderName}());\n` +
-                    `type ${componentName} = ReturnType<typeof ${componentName}>;\n` +
+                    surroundWithIgnoreComments(
```
dummdidumm, line 263:
```suggestion
                    // Surround the type with ignore comments so it is filtered out from go-to-definition etc,
                    // which for some editors can cause duplicates
                    surroundWithIgnoreComments(
```

dummdidumm (approved):
Nice solution!

PR #2761, opened 2025-05-14
feat: support generics on snippets

paoloricciuti:
`language-tools` part of https://github.com/sveltejs/svelte/pull/15915

I've added a test and manually verified that works but maybe an extra pair of eyes could benefit, i hope i didn't miss anything.

There's still a problem with the textmate grammar that I have no idea how to fix...i tried vibe coding it with awful results...specifically when a generic uses the `extends` keyword then the code looks like this

image

I noticed we don't have tests for snippets in the `svelte-vscode` package, am i mistaken?


dummdidumm (approved):
thank you!

Issue #2375, opened 2024-05-16
Can't find symbols unless *.ts/*.js file is open


shadow-identity:
### Describe the bug

There is a known [bug / inconvenience in VSCode](https://github.com/microsoft/vscode/issues/11173): if you don't open any *ts/*js files in your project, you can not find symbols defined there using "Go to symbol in workspace..." command.

But when working on Svelte project, having any *.svelte file open is not enough to make the command to work: you always have "no matching symbols" message. You still need to open *.ts file to make it work.

### Reproduction

1. Open VSCode with Svelte plugin enabled
2. Open any workspace containing svelte files
3. Open any *.svelte file in editor
4. Execute "Go to symbol in workspace..." command from the command palette and try to find any existing symbol from your workspace
5. See that there is no results

### Expected behaviour

Open *.svelte file should be treated by VScode as a ts/js file, and should activate JS/TS extension to perform symbol search.

### System Info

- OS: macos 14.4.1 (23E224)
- IDE: vscode Version: 1.89.0 (Universal)

### Which package is the issue about?

Svelte for VS Code extension

### Additional Information, eg. Screenshots

<img width="407" alt="image" src="https://github.com/sveltejs/language-tools/assets/1223112/39bda883-1c2b-4114-b1bd-7d8b4427a1d5">

jasonlyu123:
Marking this as a feature request because we didn't actually have workspace symbol support. Most of the current TypeScript features for svelte files are also provided by the Svelte extension while The TypeScript extension provides the ts/js features. So it won't automatically activate the TypeScript extension.

shadow-identity:
Could be convenient to have at least TypeScript extension activated if the Svelte file with `script lang=ts` is open.

PR #2769, opened 2025-05-23
feat: workspace symbols support

jasonlyu123:
#2375. 
Adding support in the Svelte language server so that you don't need to open ts/js files to use this feature. You can trigger this feature with `Ctrl + T` and search for symbol names. Some of the handling is also synced to the typescript-plugin so that it's more consistent, which also means vscode can better deduplicate the results from both tsserver and Svelte language server. 

If users are using an lsp client that doesn't deduplicate results from different language servers, you can also disable this feature so that it doesn't show duplicate entries. Or if your client allows you to write your own middleware, you can skip the calculation if the ts extension is running.


dummdidumm (approved):
nice work!

Issue #2773, opened 2025-05-30
svelte-check --watch crashes when the bundler removes files from output directory


odinhb:
#### Description

We use esbuild to build/bundle our javascript and it may add/remove files in the public/dist directory of our project. When it does so svelte-check sometimes crashes with this stacktrace:

<details>
<summary>stacktrace</summary>

```
Watching for file changes...node:fs:600
  handleErrorFromBinding(ctx);
  ^

Error: ENOENT: no such file or directory, open '/home/odin/dev/project/public/dist/chunk-56ZZUHUH.js'
    at Object.openSync (node:fs:600:3)
    at Object.readFileSync (node:fs:468:35)
    at DiagnosticsWatcher.updateDocument (/home/odin/dev/project/node_modules/.pnpm/svelte-check@3.5.2_postcss@8.4.31_svelte@4.2.2/node_modules/svelte-check/dist/src/index.js:103683:44)
    at FSWatcher.<anonymous> (/home/odin/dev/project/node_modules/.pnpm/svelte-check@3.5.2_postcss@8.4.31_svelte@4.2.2/node_modules/svelte-check/dist/src/index.js:103675:39)
    at FSWatcher.emit (node:events:513:28)
    at FSWatcher.emitWithAll (/home/odin/dev/project/node_modules/.pnpm/chokidar@3.5.3/node_modules/chokidar/index.js:540:8)
    at FSWatcher._emit (/home/odin/dev/project/node_modules/.pnpm/chokidar@3.5.3/node_modules/chokidar/index.js:632:8)
    at NodeFsHandler._handleFile (/home/odin/dev/project/node_modules/.pnpm/chokidar@3.5.3/node_modules/chokidar/lib/nodefs-handler.js:400:14)
    at NodeFsHandler._addToNodeFs (/home/odin/dev/project/node_modules/.pnpm/chokidar@3.5.3/node_modules/chokidar/lib/nodefs-handler.js:637:21) {
  errno: -2,
  syscall: 'open',
  code: 'ENOENT',
  path: '/home/odin/dev/project/public/dist/chunk-56ZZUHUH.js'
}

Node.js v18.14.0
 ELIFECYCLE  Command failed with exit code 1.
```

</details>

I also encountered a very similiar crash on a unix socket created by [overmind](https://github.com/DarthSim/overmind):

<details>
<summary>stacktrace</summary>

```
Watching for file changes...node:fs:600
  handleErrorFromBinding(ctx);
  ^

[...]

PR #2779, opened 2025-06-11
fix: invalidate project file cache and handle watcher race condition

jasonlyu123:
#2773

This delays the file reading in `svelte-check --watch` until we're sure the new file is a project file. I still added a file existence check before reading, but it might be more of a fail-safe because the chance of it being deleted at this point is much smaller. 

While checking this, I also found that the new file watch is broken because it now caches the tsconfig parsing result. So I changed the "read the config again" logic to the one language server used to invalidate the project files cache.

### DIFF
--- a/packages/svelte-check/src/index.ts
+++ b/packages/svelte-check/src/index.ts
@@ -194,6 +194,7 @@ class DiagnosticsWatcher {
         await this.svelteCheck.upsertDocument(
             {
                 // delay reading until we actually need the text
+                // prevents race conditions from crashing svelte-check when something is created and deleted immediately afterwards
                 get text() {
                     return fs.existsSync(path) ? fs.readFileSync(path, 'utf-8') : '';
                 },


packages/svelte-check/src/index.ts
```diff
@@ -191,8 +191,16 @@ class DiagnosticsWatcher {
     }
 
     private async updateDocument(path: string, isNew: boolean) {
-        const text = fs.readFileSync(path, 'utf-8');
-        await this.svelteCheck.upsertDocument({ text, uri: URI.file(path).toString() }, isNew);
+        await this.svelteCheck.upsertDocument(
+            {
+                // delay reading until we actually need the text
```
dummdidumm, line 196:
```suggestion
                // delay reading until we actually need the text
                // prevents race conditions from crashing svelte-check when something is created and deleted immediately afterwards
```

dummdidumm (approved):
👍

PR #2804, opened 2025-07-23
feat: add support for loading svelte config with .ts or .mts

dominikg:
extension in runtimes that support it.

Should .cjs be removed? The `--experimental-modules` flag is probably not required anymore either.

TODO: tests

Question: why is the config loader using find-up? As far as i am aware svelte config isn't discovered that way by other tools like sveltekit or vite-plugin-svelte, leading to potential mismatches if there is a stray svelte.config.js file deeper down in the source tree

### DIFF
--- a/packages/language-server/src/lib/documents/configLoader.ts
+++ b/packages/language-server/src/lib/documents/configLoader.ts
@@ -47,10 +47,10 @@ const _dynamicImport = new Function('modulePath', 'return import(modulePath)') a
     modulePath: URL
 ) => Promise<any>;
 
-const configRegex = /\/svelte\.config\.(js|cjs|mjs)$/;
+const configRegex = /\/svelte\.config\.(js|ts|cjs|mjs|mts)$/;
 
 /**
- * Loads svelte.config.{js,cjs,mjs} files. Provides both a synchronous and asynchronous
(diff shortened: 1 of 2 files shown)


jasonlyu123:
The reason why it's a find-up is that the original version, before ESM support, is handled by cosmiconfig. Not sure if there is anyone actually using it.

dummdidumm:
What do you mean by find-up?
The reason we go up/down the tree is that we need to discover the config file before a potential typescript language service is instantiated, whose setting might be influenced by what's in the Svelte config.

dominikg:
in these situations, language-tools might use a svelte.config.js file that sveltekit or vite-plugin-svelte would ignore:

no svelte.config.js inside `app/` but outside there is one
```sh
svelte.config.js
app/package.json
app/vite.config.js
app/src/App.svelte
```

extra svelte config inside src is closer than the one in `app/`
```sh
app/package.json
app/vite.config.js
app/svelte.config.js
app/src/foo/App.svelte 
app/src/svelte.config.js 
```

ideally it would be able to discover svelte app root by some kind of heuristic, eg .git or package manager lock, package.json and svelte.config.js in same dir and package.json has svelte in deps?

Issue #2797, opened 2025-07-16
Unable to exclude directories from being watched with --tsconfig


Corrl:
### Describe the bug

When running `svelte-check` in watch mode I run into an issue
```
Getting Svelte diagnostics...

failed to load config from /Users/.../vite.config.ts
Preprocessing failed
Error: spawn EBADF
    at ChildProcess.spawn (node:internal/child_process:421:11)
    at Object.spawn (node:child_process:762:9)
    at ensureServiceIsRunning (/Users/.../node_modules/esbuild/lib/main.js:1982:29)
```
which I tracked down to be related to the `chokidar` from version 4 with projects including a lot of files on macOS https://github.com/paulmillr/chokidar/issues/1385  

Both downgrading to `svelte-check@4.0.2` with as `chokidar@3.6.0` dependency or removing some directories containing a lot of files from the project solve the issue. So I would like to exclude these directories from being watched. The docs say
> --ignore \<paths\>
> Only has an effect when used in conjunction with --no-tsconfig. **When used in conjunction with --tsconfig, this will only have effect on the files watched**, not on the files that are diagnosed, which is then determined by the tsconfig.json.  

Which is contradictory. Adding the `--ignore` flag gives the error
```
> svelte-kit sync && svelte-check --tsconfig ./tsconfig.json --watch --ignore="build"

Error: Invalid svelte-check CLI args: `--ignore` only has an effect when using `--no-tsconfig`
```
that is caused by this line https://github.com/sveltejs/language-tools/blob/14dc5841442eb972bc57b724067979ce5b4df24a/packages/svelte-check/src/options.ts#L76 that was added with https://github.com/sveltejs/language-tools/pull/2453 to solve https://github.com/sveltejs/language-tools/issues/1074 so the docs don't seem to be up to date.

If the `--ignore` flag can't be used in conjunction with `--tsconfig`, how can I prevent the directories from being watched without disabling the checking of `.js` / `.ts` files? Excluding the directories in `tsconfig.json` doesn't seem to have an effect on that.

### Reproduction

Use `--ignore` flag in cunjuntion with `--tsconfig`

### Expected behaviour

Exclude ignored directories from being watched

### System Info

macOS 13.7.2
svelte-check 4.2.2

### Which package is the issue about?

svelte-check

jasonlyu123:
[...]

PR #2807, opened 2025-07-27
fix: respect tsconfig/jsconfig exclude patterns in file watcher

aewing:
- Add support for tsconfig/jsconfig exclude patterns in watch mode
- Support extended configs with proper inheritance chain
- Dynamic config reloading with 5-second debounce
- Watch svelte.config.js changes for generated aliases
- Use async TypeScript import to avoid hard dependency
- Add --ignore-tsconfig-excludes CLI flag for backwards compatibility

### DIFF
--- a/packages/svelte-check/src/index.ts
+++ b/packages/svelte-check/src/index.ts
@@ -145,18 +145,42 @@ async function getDiagnostics(
 
 class DiagnosticsWatcher {
     private updateDiagnostics: any;
+    private tsconfigCache: Map<string, { excludes: string[], mtime: number }> = new Map();
+    private tsconfigExcludes: string[] = [];
+    private svelteConfigPath?: string;
+    private svelteConfigMtime?: number;
+    private lastConfigCheck: number = 0;
+    private configCheckInterval: number = 5000; // Check every 5 seconds max
 
     constructor(
         private workspaceUri: URI,
         private svelteCheck: SvelteCheck,
         private writer: Writer,
         filePathsToIgnore: string[],
-        ignoreInitialAdd: boolean
+        ignoreInitialAdd: boolean,
+        private tsconfigPath?: string,
+        private respectTsconfigExcludes: boolean = true
     ) {
         const fileEnding = /\.(svelte|d\.ts|ts|js|jsx|tsx|mjs|cjs|mts|cts)$/;
         const viteConfigRegex = /vite\.config\.(js|ts)\.timestamp-/;
         const userIgnored = createIgnored(filePathsToIgnore);
         const offset = workspaceUri.fsPath.length + 1;
+        
+        // Look for svelte config
+        const possibleSvelteConfigs = ['svelte.config.js', 'svelte.config.mjs', 'svelte.config.cjs'];
+        for (const configName of possibleSvelteConfigs) {
+            const configPath = path.join(workspaceUri.fsPath, configName);
+            if (fs.existsSync(configPath)) {
+                this.svelteConfigPath = configPath;
+                this.svelteConfigMtime = fs.statSync(configPath).mtimeMs;
+                break;
+            }
+        }
+        
+        // Load initial tsconfig excludes
+        if (respectTsconfigExcludes && tsconfigPath) {
+            this.loadTsconfigExcludes();
+        }
 
         watch(workspaceUri.fsPath, {
             ignored: (path, stats) => {
@@ -168,12 +192,35 @@ class DiagnosticsWatcher {
                     return true;
                 }
 
+                const relativePath = path.slice(offset);
+
+                // Check user-specified ignores
                 if (userIgnored.length !== 0) {
-                    path = path.slice(offset);
                     for (const i of userIgnored) {
-                        if (i(path)) {
+                        if (i(relativePath)) {
                             return true;
                         }
+                    }
+                }
+
+                
+                for (const exclude of this.tsconfigExcludes) {
+                    if (exclude.endsWith('/**')) {
+                        const dir = exclude.slice(0, -3);
+                        if (relativePath.startsWith(dir + '/')) {
+                            return true;
+                        }
+                    } else if (exclude.includes('*')) {
+                        const regexPattern = exclude
+                            .replace(/\*\*/g, '.*')
+                            .replace(/\*/g, '[^/]*')
+                            .replace(/\./g, '\\.');
+                        const regex = new RegExp('^' + regexPattern + '$');
+                        if (regex.test(relativePath)) {
+                            return true;
+                        }
+                    } else if (relativePath.startsWith(exclude)) {
+                        return true;
                     }
                 }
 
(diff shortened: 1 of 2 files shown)


jasonlyu123:
Thank you for the contribution, but this isn't the right approach. We already have `tsconfig.json` parsing logic in the `language-server` package. https://github.com/sveltejs/language-tools/blob/master/packages/language-server/src/plugins/typescript/service.ts. This file also has logic for watching directories in the `include` patterns that are outside the workspace root. It also watches the tsconfig.json, so you also don't need to watch svelte.config.js for changes that might change tsconfig.json. 

Also, as I mentioned in #2797. I think the right approach is to watch only directories that match the `include` pattern. Not the other way around.

aewing:
> Thank you for the contribution, but this isn't the right approach. We already have `tsconfig.json` parsing logic in the `language-server` package. https://github.com/sveltejs/language-tools/blob/master/packages/language-server/src/plugins/typescript/service.ts. This file also has logic for watching directories in the `include` patterns that are outside the workspace root. It also watches the tsconfig.json, so you also don't need to watch svelte.config.js for changes that might change tsconfig.json.

Fair point. I will address this.
 
> Also, as I mentioned in #2797. I think the right approach is to watch only directories that match the `include` pattern. Not the other way around.

Understood. I had already prepared this as a stop gap for the EMFILE issue locally before that discussion. I will address this as well.

aewing:
@jasonlyu123 I have made changes to address your feedback, let me know if you want me to address anything else or take another approach.

aewing:
@jasonlyu123 I have cleaned this up. 
Added a method to get the project config from the typescript service. 
Added an async getWatchDirectories to the svelte-check service in language-server. 
Saved the watcher as an object property of the DiagnosticsWatcher.
On init/project reload we traverse watched directories and watch/unwatch from a heuristic diff.

packages/svelte-check/src/index.ts
```diff
@@ -210,10 +242,15 @@ class DiagnosticsWatcher {
         this.scheduleDiagnostics();
     }
 
-    scheduleDiagnostics() {
+    scheduleDiagnostics(updateWatchers = false) {
```
jasonlyu123, line 245: I would prefer this to be a separate method like `onProjectReloaded` since updateWatcher is not related to `scheduleDiagnotics`

jasonlyu123 (changes requested):
Thank you, I left some comments.

aewing:
> Thank you, I left some comments.

addressed feedback, thank you!

jasonlyu123:
There is another problem with this change. This will no longer watch the files that are loaded through module resolution and under the workspace root, but not under the "include" pattern. Of course, the old behaviour isn't ideal either. It should also watch files that aren't under the workspace root. 

Having thought about this more, I am wondering if we should add the parent directory of these files to the result of `SvelteCheck.getWatchDirectories`. But this would make the check a lot more frequent. Or maybe it should be done in a callback option. The service.ts or the snapshotManager calls the callback function when a new snapshot is created. @dummdidumm, what do you think?

aewing:
> There is another problem with this change. This will no longer watch the files that are loaded through module resolution and under the workspace root, but not under the "include" pattern. Of course, the old behaviour isn't ideal either. It should also watch files that aren't under the workspace root.
> 
> Having thought about this more, I am wondering if we should add the parent directory of these files to the result of `SvelteCheck.getWatchDirectories`. But this would make the check a lot more frequent. Or maybe it should be done in a callback option. The service.ts or the snapshotManager calls the callback function when a new snapshot is created. @dummdidumm, what do you think?

Thoughts on this solution?
https://github.com/sveltejs/language-tools/pull/2807/commits/8b53cb8ccf8b66643edc7b10952f348e848ae015

packages/svelte-check/src/index.ts
```diff
@@ -143,49 +143,93 @@ async function getDiagnostics(
...
+        }
+        this.watcher.add(dir);
+        this.currentWatchedDirs.add(dir);
+        // New files might now be visible; schedule a run
+        this.scheduleDiagnostics();
+    }
+
+    private async updateWatchedDirectories() {
+        const watchDirs = await this.svelteCheck.getWatchDirectories();
+        const dirsToWatch = watchDirs || [{ path: this.workspaceUri.fsPath, recursive: true }];
```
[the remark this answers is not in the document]
jasonlyu123: Ah.  Chokidar doesn't support specifying recursive here. I'm fine with it always being recursive. So we don't need to keep the recursive info here, we just need to check if it's under a directory we already watched.

packages/language-server/src/plugins/typescript/LSAndTSDocResolver.ts
```diff
@@ -46,6 +46,11 @@ interface LSAndTSDocResolverOptions {
     tsSystem?: ts.System;
     watchDirectory?: (patterns: RelativePattern[]) => void;
     nonRecursiveWatchPattern?: string;
+    /**
+     * Optional callback invoked when a new snapshot is created.
+     * Used by svelte-check to dynamically add parent directories to file watchers.
+     */
+    onSnapshotCreated?: (dirPath: string) => void;
```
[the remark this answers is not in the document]
aewing: Refactored to `onFileSnapshotCreated`

packages/svelte-check/src/index.ts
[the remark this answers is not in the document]
aewing: Addressed this feedback, thank you!

packages/svelte-check/src/index.ts
```diff
@@ -143,51 +143,118 @@ async function getDiagnostics(
...
-            ignoreInitial: ignoreInitialAdd
+            ignoreInitial: this.ignoreInitialAdd
         })
             .on('add', (path) => this.updateDocument(path, true))
             .on('unlink', (path) => this.removeDocument(path))
             .on('change', (path) => this.updateDocument(path, false));
 
-        if (ignoreInitialAdd) {
-            this.scheduleDiagnostics();
+        this.updateWatchedDirectories();
```
jasonlyu123, line 197: You probably removed the `scheduleDiagnotics` because it previously got triggered by a new directory file watch. But after that is removed, it now just hangs there until a file changes. I think we can change this to run a diagnostic immediately and run `updateWatchDirectories` afterwards.

```suggestion
        if (this.ignoreInitialAdd) {
            getDiagnostics(this.workspaceUri, this.writer, this.svelteCheck).then(() => {
                this.updateWatchedDirectories();
            });
        }
```

This way, the initial diagnostics don't have to wait an extra second, and the loading message "Getting Svelte diagnostics..." shows up before TypeScript builds the program.

jasonlyu123: Actually no. This would be updateWatchedDirectories run after all the `addWatchDirectory`.

jasonlyu123: Sorry. I changed my mind. Let's still let `updateWatchedDirectories` run before `getDiagnostics`. Running `updateWatchedDirectories` won't let TypeScript build the program. It'll only parse the config file. I get confused because most other features in the language server always trigger TypeScript to build the program first.

jasonlyu123 (approved):

mustafa0x:
Would be awesome if we can get a new release. I currently can't use LSP-Svelte in many projects due to this issue. https://github.com/sublimelsp/LSP-svelte/issues/155

Thanks a lot for this @aewing.

edit: I fixed for now by manually adding .venv to `const gitOrNodeModules = /\.git|node_modules|.venv/;` in `~/Library/Caches/Sublime Text 3/Package Storage/LSP-svelte/server/node_modules/svelte-language-server/dist/src/lib/FallbackWatcher.js`

update: that wasn't enough. i had to add another folder. to quickly test, i used this script

```js
const { FallbackWatcher } = require('./node_modules/svelte-language-server/dist/src/lib/FallbackWatcher.js')
const { FileChangeType } = require('./node_modules/vscode-languageserver/node.js')

const watch_extensions = ['.ts', '.js', '.mts', '.mjs', '.cjs', '.cts', '.json', '.svelte']
const workspace_paths = ['/my/app/dir']

const watcher = new FallbackWatcher(watch_extensions, workspace_paths)

watcher.onDidChangeWatchedFiles(({ changes }) => {
  console.log('fs events:', changes.length)
})

console.log('fallback watcher running...')
setInterval(() => {}, 1000)
```

PR #2808, opened 2025-08-01
chore: bump to typescript 5.9

jasonlyu123:
This adds support for the new [hover length config](https://devblogs.microsoft.com/typescript/announcing-typescript-5-9/#configurable-maximum-hover-length). One thing that is caught by the test is that the "click" event is now "PointerEvent" instead of `MouseEvent`. This might need to be synced to `svelte/elements` later.


jasonlyu123:
In case anyone is wondering, [expandable hover](https://devblogs.microsoft.com/typescript/announcing-typescript-5-9/#expandable-hovers-(preview)) won't be supported in the Svelte files any time soon. The VSCode API is experimental, it's also not supported by LSP, and even the TypeScript API is considered internal.

varungandhi-src:
This PR causes a regression; it can cause a stack overflow, so svelte-check no longer works for our work codebase. Potentially related to this upstream issue: https://github.com/microsoft/TypeScript/issues/62218

I'll see if we can provide a minimal reproducer upstream.

jasonlyu123:
@varungandhi-src, it should be https://github.com/sveltejs/language-tools/issues/2819. Can you check if your tsconfig.json has both `references` and `include` config? If so, check if the referenced project has the same include config as the main tsconfig. 

Edit: Another situation is if you only have `references` and not `include` and `files` in your tsconfig.json, becuase that means include everything.

PR #2811, opened 2025-08-03
Update +error.svelte creation template in vscode extension

### LINKED ISSUE #2810: +error.svelte page creation in vscode from template uses outdated syntax
### Describe the bug

+error.svelte page creation in vscode from template uses `'$app/stores'` instead of `'$app/state'`

### Reproduction

Create +error.svelte file using Svelte VSCode extension

### Expected behaviour

Uses latest syntax

### System Info

All latest OS
Latest VSCode
Latest SvelteKit VSCode extension

Svelte for VS Code extension

_No response_

olegpolin:
Fixes #2810

### DIFF
--- a/packages/svelte-vscode/src/sveltekit/generateFiles/templates/error.ts
+++ b/packages/svelte-vscode/src/sveltekit/generateFiles/templates/error.ts
@@ -2,6 +2,14 @@ import { GenerateConfig, Resource } from '../types';
 
 const defaultScriptTemplate = `
 <script>
+    import { page } from '$app/stores';
+</script>
+
+<h1>{$page.status}: {$page.error.message}</h1>
+`;
+
+const jsSv5ScriptTemplate = `
+<script>
     import { page } from '$app/state';
 </script>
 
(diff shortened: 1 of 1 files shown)


dummdidumm:
We have to check which version of Svelte/SvelteKit people use since this will only work when you're using a sufficiently new version of both

olegpolin:
> We have to check which version of Svelte/SvelteKit people use since this will only work when you're using a sufficiently new version of both

Thanks, added the different variants

olegpolin:
@dummdidumm is it fine now?

jasonlyu123:
Can you add a flag similar to the `withProps` used in `layout.ts` and `page.ts`? So that it only shows in SvelteKit 2.12 and above.

olegpolin:
Oops misclick, sorry

olegpolin:
@jasonlyu123 So should withRunes be changed to withProps? Since withProps without withRunes is impossible and the +error.svelte template doesn't use props?

jasonlyu123:
I meant to add a new `withAppState` option. The `withProps` option is for the `PageData`  => `PageProps` migration, similar to the `$app/store` to `$app/state` migration.

olegpolin:
Got it, thanks

olegpolin:
@jasonlyu123 added

dummdidumm (approved):
thank you!

PR #2812, opened 2025-08-04
chore: specify non-beta release as peer dependency for svelte@5

martme:
svelte-check does not currently accept the latest stable version of svelte as a peer dependency, but will fail gracefully with a warning. This PR should resolve that issue.

### DIFF
--- a/packages/svelte-check/package.json
+++ b/packages/svelte-check/package.json
@@ -33,7 +33,7 @@
         "sade": "^1.7.4"
     },
     "peerDependencies": {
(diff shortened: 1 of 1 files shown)


jasonlyu123:
The current version specifier should accept the latest stable version. What is the error you encountered?

martme:
You're totally correct. Coderabbit has been reporting this issue in our pipelines, but turns out it's entirely a non-issue and the robots are not to be trusted in this instance 🤖  Sorry for waisting your time!

Issue #2843, opened 2025-09-11
svelte-check fails to infer props when component has `<svelte:options runes={false} />`


axel7083:
### Describe the bug

To test our UI library, we have some backward compatibility tests, where we force a component to be renderer in svelte4, here is an example

```svelte
<svelte:options runes={false} />
<script lang="ts">
  import { Button } from '/@/lib';

  export let onclick: () => void;
  export let content: string | undefined;
</script>

  <Button on:click={onclick} >
    {content}
</Button>
```

When using this component inside our `test.spec.ts` as followed

```ts
import '@testing-library/jest-dom/vitest';

import { fireEvent, render } from '@testing-library/svelte';
import { expect, test, vi } from 'vitest';

import ButtonSvelte4Test from './ButtonSvelte4Test.svelte';

test('should render content', () => {
  const { getByText } = render(ButtonSvelte4Test, {
    onclick: vi.fn(),
    content: 'Dummy Button',
  });

  const btn = getByText('Dummy Button');
  expect(btn).toBeInTheDocument();
});
```

We get the following error when running `svelte-check`

```
Error: Type 'Mock<Procedure>' is not assignable to type 'never'. 
  const { getByText } = render(ButtonSvelte4Test, {
    onclick: vi.fn(),                                                                                                                                                                                                         
    content: 'Dummy Button',

c:\Users\axels\Documents\github\redhat\podman-desktop\packages\ui\src\lib\button\ButtonSvelte4Test.spec.ts:30:5
Error: Type 'string' is not assignable to type 'never'.
    onclick: vi.fn(),
    content: 'Dummy Button',                                                                                                                                                                                                  
  })
```

Removing the `<svelte:options runes={false} />` files the error, but not what we want...

### Reproduction

See above

### Expected behaviour

The props should be properly typed

### System Info

- OS: Windows
- IDE: Webstorm


### Which package is the issue about?

svelte-check

### Additional Information, eg. Screenshots

_No response_

PR #2847, opened 2025-09-11
fix(svelte2tsx): support for runes={false} in svelte:options

axel7083:
## Description

The `<svelte:options runes/>` were not checking for the value, so it was always true. When using `<svelte:options runes={false}/>` the props exported were `Record<string, never>` which is incorrect.

## Related issues

Fixes https://github.com/sveltejs/language-tools/issues/2843

## Testing

- [x] appropriate sample has been added

### DIFF
--- a/packages/svelte2tsx/test/svelte2tsx/samples/ts-export-list-runes-false/expected-svelte5.ts
+++ b/packages/svelte2tsx/test/svelte2tsx/samples/ts-export-list-runes-false/expected-svelte5.ts
@@ -1,0 +1,26 @@
+///<reference types="svelte" />
+;function $$render() {
+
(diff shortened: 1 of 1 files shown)


axel7083:
> This PR includes changesets to release 3 packages
> Name	Type
> svelte2tsx	Patch
> svelte-language-server	Patch
> typescript-svelte-plugin	Patch

Pretty weird that `svelte-check` is not bumped? it depends on `svelte-language-server` :/

axel7083:
@dummdidumm I updated the failing test 👍

axel7083:
@dummdidumm I added the `expected-svelte5.ts` after seeing the test-svelte5 failing. (I was not aware it was required first)

axel7083:
@dummdidumm yeahh all tests are passing :p

PR #2853, opened 2025-09-18
perf: use failed path to invalidate module cache

jasonlyu123:
Alternative to the module cache handling of #2852. Still have to check why the "adding all missing imports" test failed in local.

### DIFF
--- a/.changeset/dark-falcons-pump.md
+++ b/.changeset/dark-falcons-pump.md
@@ -1,0 +1,6 @@
+---
+'svelte-language-server': patch
+'svelte-check': patch
+---
+
+perf: more precise module cache invalidation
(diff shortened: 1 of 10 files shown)


43081j:
this looks great

locally i can see a huge speed up in createSnapshot too (now <10ms self vs 600 before)

43081j (approved):
all looks good to me 🙏 

it runs a _lot_ faster 😄

jasonlyu123:
For svelte-check and editor startup, the change is mainly skipping the cache invalidation check.  `getSnapshot` and `getSnapshotIfExist` both get called because it's resolved through module resolution or the file is a project file. So we should already know that the file exists. The credit should still belong to @43081j for finding out that the module cache is immediately invalidated after it was created. 

And for the cache invalidation check, it was scheduled to run before TypeScript builds the program. The check is done by checking the failed file path that TypeScript recorded during module resolution. So it should only run after the first load since we won't have the info before that.

dummdidumm (approved):
Awesome!

Issue #2858, opened 2025-10-03
Svelte-check --watch doesn't watch any longer in version 4.3.2 (works in 4.3.1 and earlier)


Al38andr0:
### Describe the bug

As for the title, I unfortunately don't have any more details available; it just doesn't re-run after any file has changed.

Since I'm here, does anyone know why it runs twice (most often, but not always) when it starts?

Thank you.

### Reproduction

Execute the command: `npx svelte-check --watch` or `npm run svelte-check --watch` (passing the proper configuration and variables, if needed).

Modify any TS or Svelte file.

The console doesn't refresh the results.

### Expected behaviour

For versions 4.3.1 and earlier, I expect to see the Svelte-check result updated every time any of my files change.

### System Info

- MacBook Pro - Chip: Apple M1 Pro - OS: Sequoia 15.6.1
- Webstorm 2025.2.2
- Node 24.9.0

### Which package is the issue about?

svelte-check

jasonlyu123:
The main issue is a duplicate of #2855. As for why it runs twice when initialised, it's usually because vite-plugin-svelte's vite-preprocess triggers SvelteKit to regenerate the ./svelte-kit/tsconfig.json. I'll see what we can do to prevent it. (I tried it before, but haven't figured out an efficient way to do it. )

Issue #2855, opened 2025-09-22
`svelte-check`'s `--watch` mode doesn't work because of regression from #2807


abrahamguo:
### Describe the bug

in #2807 (@jasonlyu123) (specifically, in [refactor(check): clarify snapshot callback and simplify directory watching; ignore .crush](https://github.com/sveltejs/language-tools/pull/2807/commits/5ca5f125ffc40133a530c40295beac1f32723316)), @aewing introduced a regression that can cause watch mode to not run.

In that commit, this code was added:

```
        // If new dir is a parent of existing ones, unwatch children
        const toRemove: string[] = [];
        for (const existing of this.currentWatchedDirs) {
            if (this.isSubdir(existing, dir)) {
                toRemove.push(existing);
            }
        }
        if (toRemove.length) {
            this.watcher.unwatch(toRemove);
```

`this.watcher` refers to a `chokidar` watcher instance.

It's not explicitly documented in Chokidar's documentation, but `unwatch`ing a child absolute path, followed by `add`ing a parent absolute path, causes the child paths to remain ignored.

### Reproduction

```
import { watch } from 'chokidar';

const { dirname } = import.meta;
const a = `${dirname}/a`;
const b = `${a}/b`;
watch(b).unwatch(b).add(a).on(`change`, () => console.log(`change!`));
```

Note that the `change` handler runs for everything in the `a` directory, *except for* changes that happen in the `b` subdirectory.

### Expected behaviour

`svelte-check` should properly handle parent and child directories when passing to `chokidar`.

### System Info

- OS: [e.g. macOS]

### Which package is the issue about?

svelte-check

Al38andr0:
@jasonlyu123 , I've just pulled v.4.3.3, but the --watch is still not working.

aewing:
I hate that I didn't see this sooner and own this problem. I am sorry for the time bomb, I tried to shake issues like this out. Let me know if I can help mitigate fallout.

aewing:
> @jasonlyu123 , I've just pulled v.4.3.3, but the --watch is still not working.

Can you provide a simple reproduction or more details about what you hope to see but aren't? I will look into it.

Al38andr0:
I'd love to be more helpful, but the truth is, it just doesn't refresh the check when changes occur. No log, no messages, it just doesn't check again.

[...]

PR #2859, opened 2025-10-04
fix: prevent file watcher issue 

jasonlyu123:
For #2855 and #2858

The issue happens when there is a file loaded in the parent directory of a tsconfig.json include directory. For example, vite.config.ts. It also depends on the order in which the file is loaded.


dummdidumm (approved):
👍

PR #2867, opened 2025-10-17
fix: support for @nativescript-community/svelte-native

farfromrefug:
We moved the svelte-native plugin to @nativescript-community. This supports it while still supporting the old package


dummdidumm (approved):
thank you!

PR #2879, opened 2025-11-02
✨ Use custom element's JSDoc as doc for completion/hover

ouvreboite:
Currently, custom elements added to `SvelteHTMLElements` supports autocompletion, but no description is displayed (from the JSDoc), while their attributes support both autocompletion and JSDoc description.

This PR's goal is to enable JSDoc documentation during autocompletion and hover for custom element's tags.

For example, this declaration would yield this autocomplete:
```typescript
declare module 'svelte/elements' {
  export interface SvelteHTMLElements {
    /** 
     * Custom doc for custom element 
     * 
     * ### Details
     * Some interesting stuff about this component (in markdown)
    */
    'custom-element': any;
  }
}
```

### DIFF
--- a/.changeset/few-moons-crash.md
+++ b/.changeset/few-moons-crash.md
@@ -1,0 +1,5 @@
+---
+'svelte-language-server': patch
+---
+
+feat: custom element JSDoc documentation for completion/hover


packages/language-server/src/plugins/typescript/features/CompletionProvider.ts
```diff
@@ -544,87 +546,32 @@ export class CompletionsProviderImpl implements CompletionsProvider<CompletionRe
...
-
-        const typingsNamespace = lsContainer.getTsConfigSvelteOptions().namespace;
-
-        const typingsNamespaceSymbol = this.findTypingsNamespaceSymbol(
-            typingsNamespace,
-            typeChecker,
-            sourceFile
+        let tags = getCustomElementsSymbols(lang, lsContainer, tsDoc).filter((t) =>
+            t.name.startsWith(tag.tag!)
         );
```
ouvreboite: Before, only the CompletionProvider was checking the custom elements, and not the HoverProvider. So I moved the logic that extracts the custom "types" to utils.ts to share across both

jasonlyu123:
I think it would be better to resolve the documentation in `resolveCompletion`. Could you check if it's possible? This way, it'll only calculate the documentation when it's needed. In hover, the reused method also only needs to calculate the documentation needed instead of all the customer elements.

ouvreboite:
I'm away from home the next few days, but I'll look into that shortly.

PR #2898, opened 2025-12-06
Remove outdated "experimental-modules" runtime flag

MathiasWP:
Based on the comments in the code it seems like this is ready to be removed


aewing (approved):
Makes sense to me

dummdidumm (approved):
thank you!

PR #2900, opened 2025-12-06
[perf]: only create objects for completion items for svelte syntax once

MathiasWP:
Re-creating these objects every time `getCompletionsWithRegardToTriggerCharacter` is called is unnecessary. Only creating them once should free up a small bit of memory.


aewing:
I don't see any reason why we would need to do this dynamically, so this change is moving in a sensible direction. Not much gained from the change though. Plus it feels odd to have these features just inlined here and not imported from some centralized language feature metadata location or something, but that's a larger and separate discussion.

PR #2905, opened 2025-12-15
fix(syntax): sveltejs ts highlighting when trailing type alias misses semicolon

### LINKED ISSUE #1110: Export type breaks syntax highlighting
**Describe the bug**
When module context has only `export type`, syntax highlighting breaks
```svelte
<script lang="ts" context="module">
export type T = 't'
</script>
```
Adding a semicolon at the end fixes the issue

**Screenshots**
Broken:
image

Normal (with semicolon):
image

### LINKED ISSUE #1627: weird syntax hilight

### LINKED ISSUE #1776: Syntax highlighting breaks inside script `context="module"` when the last line is `export type`
### Describe the bug

Syntax highlighting breaks inside script `context="module"` when the last line is `export type`

### Reproduction

Take following component and observe incorrect color highlighting. Adding semicolon or new line fixes it.
```svelte
<script lang="ts" context="module">
	import { ICON_PATHS } from './icon-paths'

	export type IconKind = keyof typeof ICON_PATHS // This line breaks
</script>

<script lang="ts">
	export let kind: IconKind
</script>

<svg viewBox="0 0 24 24" class="fill-current">
	<path d={ICON_PATHS[kind]} />
</svg>

```

### Expected behaviour

Correct syntax highlighting when exporting type 

### System Info

- OS: MacOS
- IDE: VSCode

Svelte for VS Code extension

image

With semicolon 
image

### LINKED ISSUE #1779: incorrect syntax highlighting on `</script>` tag when previous line ends with `>`
### Describe the bug

if the last line of a `<script>` block ends with a `>`, the `</script>` has the wrong colour in vscode

### Reproduction

```svelte
<script lang="ts">
   type Foo = Array<string>
</script>
```

### Expected behaviour

both `<script>` and `</script>` are blue

### System Info

- OS: windows 10
- IDE: vscode

Svelte for VS Code extension

image

### LINKED ISSUE #1970: `export let x: Type` and type declaration without ending semicolon last in script block breaks syntax highlighting
### Describe the bug

When running Typescript mode and you have either an `export let varName: Type`, or `type Type = type`  without a line-ending semicolon last in a script block, the syntax highlighting breaks.

The problem with type declarations I've seen for ages, but the problem with component properties arrived yesterday after updating VS Code.

I think the screenshots are quite explanatory.

**Syntax highlight breaks after ZeRow**

export-let

**Syntax highlighting breaks after the ending curly brace for type declaration `Test`**

type

**Syntax highlight works when ending type declaration with a semicolon**

type-with-semi

**Syntax highlighting works when ending property declaration with semicolon**

export-let-with-semi

**Syntax highlighting works when ending script block with some other statement or expression**

other-expr

### Reproduction

This simple snippet is sufficient to to reproduce the bug.

```
<script lang="ts">
export let test: any
</script>
```

### Expected behaviour

I expect the ending `</script>` tag to be correctly highlighted

### System Info

- IDE: VS Code

Svelte for VS Code extension

_No response_

### LINKED ISSUE #1989: Syntax highlighting fails in TS script tags without a semicolon
### Describe the bug

If a TypeScript script content ends with a type without a semicolon, then the syntax highlighting breaks down:

CleanShot 2023-04-18 at 06 03 34@2x

I don't know if this is the right repo to report the issue, but the only place where the issue is apparent is in Svelte files.

### Reproduction

```svelte
<script lang="ts">
export let name: string
</script>

<div>
  {name}
</div>
```
NB: same issue with syntax highlighting here

### Expected behaviour

Expected the code to be highlighted like this, but without the use of a semicolon:

CleanShot 2023-04-18 at 06 14 37@2x

### System Info

- OS: Mac OS
- IDE: VSCode

Svelte for VS Code extension

I have reported a similar issue (https://github.com/sveltejs/language-tools/issues/1110), but it looks like the problem is less specific than what was described there.

### Workaround

Adding a comment (`//`) at the very end resolves the issue with syntax highlighting:
```svelte
<script lang="ts">
export let name: string //
</script>

<div>
  {name}
</div>
```

### LINKED ISSUE #2458: type declaration without ending semicolon last in script block breaks syntax highlighting (again)
### Describe the bug

I am experiencing the exact same symptoms from https://github.com/sveltejs/language-tools/issues/1970

Inside my script tags, if the last line is a type declaration, it breaks syntax highlighting for the rest of the file if there is no semicolon.

Prior bug claimed this was fixed in vscode 1.78
I'm using vscode 1.92.0

This is my first svelte project, so I have no prior history to reference.

### Reproduction

_component.svelte_
```
<script lang="ts">
  
  type $$Props = {}
</script>

<div>My component</div>
```

image

### Expected behaviour

(but without the semicolon)
image

### System Info

- OS: M1 Max Macbook Pro OS 14.3
- IDE: VSCode 1.92

Svelte for VS Code extension

_No response_

### LINKED ISSUE #2739: Syntax highlighting of closing `</script>` broken when omitting semicolons
### Describe the bug

When not using end of line semicolons in ts, and the last statement in the script is a type definition, the closing script tag is highlighted incorrectly, as well as the next tag in the file.

### Reproduction

Example of a constant being the last statement

Image

(the textmate scopes in this case are identical to the next example)

```svelte
<script module lang="ts">
  export const Foo = 'bar'
</script>
```

---

Example of a type being the last statement *and* ending with a semicolon

Image

Textmate scopes

Image

```svelte
<script module lang="ts">
  export type Foo = 'bar';
</script>
```

---

Example of a type being the last statement *and not* ending with a semicolon

Image

Textmate scopes

Image

```svelte
<script module lang="ts">
  export type Foo = 'bar'
</script>
```

### Expected behaviour

The closing script tag should be highlighted as markup, not as typescript

### System Info

- OS: Linux 6.12.21
- IDE: VSCode 1.99.2
- Svelte SVCode: 109.5.4

Svelte for VS Code extension

This is purely a highlighting issue, the language server parses the code correctly. Could be something broken in the typescript highlighter I suppose - I'm unfamiliar with how the svelte extension deals with the embedded language.

The issue only occurs if the characters between the end of the type declaration and the start of the closing script tag are either `/\n\s*/`, or contain line/block comments on each line after. If there is a blank line between, or any other statement, the issue doesn't occur.

Importance: almost doesn't matter

serkodev:
Fix #1110,  Fix #2739, Fix #2458, Fix #1989, Fix #1779, Fix #1776, Fix #1627, Fix #1970

### Before
b

### After
a

### Notice
- If `</script>` is intended as a closing tag but does not appear at the beginning of a line, syntax highlighting will also break (though single-line `<script>` tags are not affected but semicolon is required):
  ```svelte
  <!-- bad -->
  <script lang="ts" context="module">
    const a = 1;</script>

  <!-- good -->
  <script lang="ts" context="module">
    const a = 1
  </script>
  <script lang="ts" context="module">const a = 1;</script>
  ```
- If `</script>` appears as a `string` and at the beginning of a line, syntax highlighting will break:
  ```svelte
  <script lang="ts" context="module">
    const a = `
  </script> <- break
  `;
  </script>
  ```
  However, since `</script>` as a string currently is also not supported by Svelte, so this issue has no impact.

References:
- https://github.com/vuejs/language-tools/pull/5853

### DIFF
--- a/packages/svelte-vscode/syntaxes/svelte.tmLanguage.src.yaml
+++ b/packages/svelte-vscode/syntaxes/svelte.tmLanguage.src.yaml
@@ -23,10 +23,10 @@ injections:
   'L:(meta.script.svelte | meta.style.svelte) (meta.lang.ts | meta.lang.typescript) - (meta source)':
     patterns: [
       # single line script tag
-      {begin: '(?<=>)(?=[^\n]+</script[\s>])', end: '(?=</script[\s>])', name: meta.embedded.block.svelte,
+      {begin: '(?<=>)(?=[^\n]+</(script|style)[\s>])', end: '(?=</(script|style)[\s>])', name: meta.embedded.block.svelte,
       contentName: source.ts, patterns: [{ include: source.ts }]},
       # fix #1110
-      {begin: '(?<=>)(?!</)', while: '^(?!\s*</script[\s>])', name: meta.embedded.block.svelte,
+      {begin: '(?<=>)(?!</)', while: '^(?!\s*</(script|style)[\s>])', name: meta.embedded.block.svelte,
       contentName: source.ts, patterns: [{ include: source.ts }]}
     ]
 


jasonlyu123:
Thank you for the fix! This is indeed a long-standing problem. 

Overall all it looks good, but I found a problem with an empty script tag

```svelte
<script lang="ts"></script>
```
圖片

serkodev:
Thanks for pointing that out, fixed it again.

PR #2914, opened 2026-01-12
fix: TypeScript detection fails when <script lang="ts"> is at the bottom

### LINKED ISSUE #2854: TypeScript is not parsed correctly for snippet arguments
### Describe the bug

There seems to be an issue with parsing object types for snippet parameters. This only happens when the <script> is below the snippet definition.
```svelte
<!-- everything works when the script tag is at the top... -->
<!-- <script lang="ts"></script> -->

{#snippet foo(props: { props?: Record<string, unknown> })}{/snippet}

<!-- this won't error if you uncomment the snippet above -->
{#snippet bar(props: Record<string, unknown>)}{/snippet}

<!-- ...but not when it's at the bottom -->
<script lang="ts"></script>
```

EDIT: seems to happen in other cases too, not just snippets.
```svelte
<CalendarPrimitive.Day
    bind:ref
    class={cn(...)}
    {...restProps}
/>

<script lang="ts">
import {Calendar as CalendarPrimitive} from 'bits-ui'

import {buttonVariants} from '$lib/components/ui/button/index.js'
import {cn} from '$lib/utils.js'

let {ref = $bindable(null), class: className, ...restProps}: CalendarPrimitive.DayProps = $props()
</script>
```

### Reproduction

https://github.com/abdel-17/snippet-ts-bug

### Expected behaviour

It should not error

### System Info

- OS: [e.g. Linux]
- IDE: [e.g. Neovim]

_No response_

Image

Image

abdel-17:
fixes #2854 

In the following example, the svelte language server would fail to detect TS when `<script lang="ts">` is at the bottom.
```svelte
{#snippet foo({props}: {props?: Record<string, unknown>})}{/snippet}

<script lang="ts"></script>
```

I asked Codex to fix the issue and it did. The LSP no longer fails. I made sure to run the tests and nothing broke. I admit I don't understand the code very well, but it seems to work.

Before:
A screenshot of the example with an error indicating that TS is not detected

After:
A screenshot of the example without the error

### DIFF
--- a/packages/language-server/test/lib/documents/parseHtml.test.ts
+++ b/packages/language-server/test/lib/documents/parseHtml.test.ts
@@ -122,6 +122,6 @@ describe('parseHtml', () => {
         );
         assert.strictEqual(document.roots.length, 1);
         assert.strictEqual(document.roots[0].tag, 'script');
-        assert.strictEqual(document.roots[0].attributes?.lang, 'ts');
+        assert.strictEqual(document.roots[0].attributes?.lang, '"ts"');
     });
 });


packages/language-server/src/lib/documents/utils.ts
```diff
@@ -442,27 +442,34 @@ export function getLangAttribute(...tags: Array<TagInformation | null>): string
...
+            return false;
+        }
+    }
+
+    let depth = 0;
+    for (let i = tagStart; i < position; i++) {
+        const char = html[i];
+        if (char === '{') {
+            depth++;
+        } else if (char === '}' && depth > 0) {
```
abdel-17: I'm not sure if the `depth > 0` check in the `else if` here is correct. It looks wrong.

jasonlyu123:
Thank you for the contribution. I am not really sure if this is the proper way to fix this. Checking `{` count could work in a simple expression, but you'll have a problem if there is a `{` or `}` in the string. 

To be honest,  I'll suggest you write your script tag at the top. You might not have a good experience by putting the script tag at the bottom.  It'll be affected by a syntax error during editing way more than if you put it at the top.

abdel-17:
Yeah I wasn't sure if the code was correct either.

The decision to write script tags at the top isn't mine unfortunately. We have a prettier config that reorders script tags to the bottom. The "solution" so far has been to manually override this for files that cause a problem, but it isn't ideal. Is there a better way to deal with this?

abdel-17:
I reverted the AI's code. I tried implementing a fix by detecting type definition patterns when the parser hits `<` or `>`, but it was too complicated. Not sure how else to solve this. Would it break source maps if the `parseHtml` function placed the script at the top as a preprocessing step?

paoloricciuti:
I wonder how costly would be to do a single pass parsing of the whole component to find script lang ts beforehand.

jasonlyu123:
The reason we use vscode-html-languageservice to parse it here instead of the Svelte compiler is mostly that it is more forgiving of syntax errors, at least in the Svelte 3/4 days when this is written. The result AST is also used in various other places. 

One solution I can think of is to use the Svelte compiler as a fallback when no script tag is found.

paoloricciuti:
From what I understand the problem is not that it doesn't find the script tag but that it start parsing without knowing it has to use TS and that leads to not finding the script tag. So my idea was to do a single pass parsing specifically looking for the script tag only to figure out if we need TS so that we can properly parse everything.

jasonlyu123:
> From what I understand the problem is not that it doesn't find the script tag but that it start parsing without knowing it has to use TS and that leads to not finding the script tag.

It shouldn't be. The problem is that the `vscode-html-languageservice` parses the `<string` as an HTML tag. So the script tag is considered a child of it, not a top-level script tag. 

Ideally, we should use a js/ts parse when the html parser encounters `<` and `{`. I did try it before, but it is kind of expensive.

dummdidumm:
Closing in favor of #2921 - thank you!

PR #2921, opened 2026-01-19
fix: extract style/script tag followed by destructuring in the template

jasonlyu123:
#2854 #2155

This builds upon the original logic of #2914. Replacing the `isInsideMoustacheTag` logic with a simple bracket match with a simplified string check. Using the Svelte compiler in this preprocessor isn't really viable because even the loose parse mode still throws a syntax error with an expression while editing. 

This also rewrites the parseHTML to use a custom parse logic instead of preprocess. Previously, we scanned the HTML to remove problematic "<" and ">". Now, it does one scan/parse. This should improve performance and might reduce memory usage since V8 won't store the preprocessed string in memory for the sliced string.  The parse code is derived from vscode-html-languageservice, andthe  main difference is 

 1. Parse expression tag in Whitespace state
 2. parse attribute with interpolation in AttributeValue state
 3. Detect Svelte blocks/tags in Content state

@abdel-17 Are there other problems you encountered like this? 

@paoloricciuti, since you also commented #2914, could you check if there are other situations that aren't covered by this? This function is used to blank `<` and `>` in HTML that confuses the vscode-html-languageservice's HTML parser.

### DIFF
--- a/packages/language-server/test/lib/documents/parseHtml.test.ts
+++ b/packages/language-server/test/lib/documents/parseHtml.test.ts
@@ -126,16 +126,6 @@ describe('parseHtml', () => {
         );
     });
 
-    it('can parse html with destructured snippet and type annotation', () => {
-        testRootElements(
-            parseHtml(
-                `{#snippet foo({ props }: { props?: Record<string, unknown> })}{/snippet}
-                <Foo checked={a} />
-                <style></style>`
-            )
-        );
-    });
-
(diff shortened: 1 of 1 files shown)


abdel-17:
none that I can reproduce no

abdel-17:
Hey I just wanted to say I really appreciate you working on this issue. Looking forward to seeing it fixed. :)

Issue #2733, opened 2025-04-08
TypeScript Go support (at least in CLI)


notramo:
### Description

JSX support has been recently implemented in the [Go implementation of TypeScript](https://github.com/microsoft/typescript-go).

The new Go implementation is a lot faster (10 times faster in some projects), and uses ~half the memory as the JS implementation.

It seems that this new version has enough features for a fast CLI `svelte-check` command. LSP support is lacking at the moment, but an optional `svelte-check` config option would be really useful.

### Proposed solution

Add a command line option to `svelte-check` that enables the use of `tsgo tsc …` instead of `tsc`.

### Alternatives

None. This seems a quite simple change, with the most obvious solution being a CLI option. Other solutions seem a lot more complicated.

jasonlyu123:
No. It's not a simple change. svelte-check does not call tsc by the command. It used low-level typescript API that is about to be rewritten in typescript go. This is definitely on the radar but not anytime soon.

KieranP:
The first Typescript Native Previews just released, and with it a very early version of the APIs: https://devblogs.microsoft.com/typescript/announcing-typescript-native-previews/#api-progress

I understand that it won't be anywhere ready for use by Svelte, but it might be a good time for the Svelte language-tools team/contributors to give the TS team feedback so that APIs support what Svelte needs from day one.

jasonlyu123:
To keep the expectation in check, TypeScript [mentioned a while ago](https://github.com/microsoft/typescript-go/discussions/455#discussioncomment-13411961) that the TypeScript API library is expected to be released after TypeScript 7 is released. So this most likely won't be available on day one.

Kae7in:
Really looking forward to this, especially in our SvelteKit app where we have ~370K LoC in `.svelte` files and `svelte-check` takes about 54s :(

pheuter:
Some significant recent movement: https://devblogs.microsoft.com/typescript/progress-on-typescript-7-december-2025/

jasonlyu123:
Yeah. The TypeScript team is very committed to it and plans to stop active development on the old codebase. The API we need, however, is still pretty much the same since my last update. The main problem is that there is no way for us to alter the module resolution to resolve Svelte files.

astralhpi:
For those looking for a faster alternative until svelte-check gets tsgo support, I made a tool using svelte2tsx + tsgo: svelte-fast-check (https://github.com/astralhpi/svelte-fast-check)
[...]

PR #2932, opened 2026-02-06
feat: svelte-check `--incremental` / `--tsgo`

### LINKED ISSUE #2131: Expecting `svelte-check` to work with TS `--incremental` flag
### Describe the bug

Trying to speed up our pre-commit hook and added `"incremental": true` flag to `tsconfig.json` compiler options.

Running `pnpm tsc --noEmit` produces a `.tsbuildinfo` file while `pnpm svelte-check  --tsconfig ./tsconfig.json` doesn't.

My understanding was that `svelte-check` executes `tsc` under the hood, but I could be wrong here! 

### Reproduction

- use the following tsconfig.json:
```
{
  "extends": "./.svelte-kit/tsconfig.json",
  "compilerOptions": {
   	...
    "incremental": true
  }
}

```

- run `pnpm svelte-check  --tsconfig ./tsconfig.json`
- no `.tsbuildinfo` file shows up

### Expected behaviour

`.tsbuildinfo` file output for `pnpm svelte-check  --tsconfig ./tsconfig.json`

### System Info

- OS: Ubuntu 23.04
- IDE: VSCode

svelte-check

_No response_

dummdidumm:
Adds two new flags:

- `--incremental`: Opts into TypeScript's incremental build cache, which speeds up subsequent runs. Saved within `.svelte-kit` or if not available within `.svelte-check`. This might result in slightly different type check outcomes, and certain patterns are not supported. Specifically, anything that is not in the root dir of your tsconfig.json and is a Svelte file will not be properly loaded and type-checked.
- `--tsgo`: Use TypeScript's Go implementation. Needs to have `@typescript/native-preview` installed. Subject to the same limitations as `--incremental`

Part of #2733
closes #2131

Huge kudos to @astralhpi for building https://github.com/astralhpi/svelte-fast-check and @pheuter for building https://svelte-check-rs.vercel.app/ , both of which heavily inspired this PR

### DIFF
--- a/packages/svelte2tsx/test/svelte2tsx/samples/jsdoc-various.v5/expectedv2.ts
+++ b/packages/svelte2tsx/test/svelte2tsx/samples/jsdoc-various.v5/expectedv2.ts
@@ -9,7 +9,7 @@ function $$render() {
   let { b } = $props();
   let rect;
 
-/*Ωignore_startΩ*//** @type {ReturnType<typeof __sveltets_2_createCreateSlot>} */ const __sveltets_createSlot = __sveltets_2_createCreateSlot();/*Ωignore_endΩ*/;
+/*Ωignore_startΩ*/;const __sveltets_createSlot = __sveltets_2_createCreateSlot();/*Ωignore_endΩ*/;
 async () => {
 
 


pheuter:
Love it ❤️

jasonlyu123:
The jsdoc changes look good. 

>There is also a problem with relative imports from svelte files

About this. I meant the relative import from svelte files to js/ts files. js/ts files aren't copied to the .svelte-check directory, so relative import failed.

dummdidumm:
relative import issue should be fixed

jasonlyu123:
Is --tsgo intended to be only used alongside --incremental? Right now, if you only use --tsgo, the flag seems to be ignored.

dummdidumm:
fixed

pheuter:
Noticing a lot of `implicit any` errors when passing `--incremental --tsgo`, does it currently support Zero-effort type safety? https://svelte.dev/blog/zero-config-type-safety

/cc @dummdidumm

ceifa:
> Noticing a lot of `implicit any` errors when passing `--incremental --tsgo`, does it currently support Zero-effort type safety? https://svelte.dev/blog/zero-config-type-safety

Yep, same here. And other errors like: Error: Cannot find name '__sveltets_2_any'. (ts)

PR #2960, opened 2026-02-26
perf: micro-optimizations in hot paths

MathiasWP:
## Summary

- **Replace O(n²) `flatten()` with `Array.prototype.flat()`** — the existing reduce+spread implementation copies the accumulator on every iteration. Called from DiagnosticsProvider, CompletionProvider, FindReferencesProvider, CodeActionsProvider, and more
- **Remove lodash `flatten` from PluginHost** — replaces ~10 `flatten(...)` call sites with `.flat()`. PluginHost is the central dispatch for all LSP operations
- **Use `Object.create(node)` instead of `{ ...node }` in `svelteNodeAt`** — avoids copying all AST node properties on every matching ancestor while walking the tree. Called from completions, diagnostics, rename, and semantic tokens
- **Cache `getExistingImports` per document URI + version** — avoids re-parsing all imports via regex on every completion request
- **Early return + `RegExp.test()` in `isExistingSvelteComponentImport`** — avoids allocating a match array for every Svelte component completion item
- **Replace `possiblyComponent` regex with charCode comparison** — regex is overkill for checking a single ASCII uppercase character
- **Fix `lineCount` to use cached line offsets** — was splitting the entire document text on every call when `getLineOffsets()` is already cached
- **Optimize `getByPrefix` to iterate without intermediate array** — push loop instead of `Array.from` + filter + map chain
- **Optimize `logStatistics` to single-pass counting** — replaces Set→Array conversion and double `.filter()` with a single loop

## Test plan

- [x] All 664 existing tests pass (8 pending, 0 failing)
- [x] Clean build with `pnpm build`
- [ ] Manual smoke test: open a Svelte project, verify completions, hover, diagnostics, and rename work correctly

### DIFF
--- a/packages/language-server/src/utils.ts
+++ b/packages/language-server/src/utils.ts
@@ -294,7 +294,9 @@ export function getIndent(text: string) {
 export function possiblyComponent(node: Node): boolean;
 export function possiblyComponent(tagName: string): boolean;
 export function possiblyComponent(nodeOrTagName: Node | string): boolean {
-    const charCode = (typeof nodeOrTagName === 'object' ? nodeOrTagName.tag : nodeOrTagName)?.charCodeAt(0);
+    const charCode = (
+        typeof nodeOrTagName === 'object' ? nodeOrTagName.tag : nodeOrTagName
+    )?.charCodeAt(0);
     return charCode !== undefined && charCode >= /* A */ 65 && charCode <= /* Z */ 90;
 }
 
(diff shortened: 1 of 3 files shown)


packages/language-server/src/utils.ts
```diff
@@ -81,10 +81,7 @@ export function getLastPartOfPath(path: string): string {
 }
 
 export function flatten<T>(arr: Array<T | T[]>): T[] {
-    return arr.reduce(
-        (all: T[], item) => (Array.isArray(item) ? [...all, ...item] : [...all, item]),
-        []
-    );
+    return (arr as T[][]).flat();
```
jasonlyu123: It'll be better to replace all usage of this function with `flat` or `flatMap`.

jasonlyu123 (approved):

PR #2962, opened 2026-02-27
perf: defer source map generation until first access

MathiasWP:
## Summary

- `generateMap({ hires: true })` accounts for ~10% of `svelte2tsx` wall-clock time
- Most consumers never need the source map on the hot path:
  - **Language server**: `getText()`, `getLength()`, `getChangeRange()` don't use the map. The mapper is already lazily initialized, but the map object itself was eagerly computed.
  - **svelte-check incremental mode**: only uses `.code`, never `.map`
- The `map` property is now a lazy getter that calls `generateMap()` on first access and caches the result via `Object.defineProperty`
- Consumers that do access `.map` (TypeScript plugin, sourcemap tests) see no regression — the map is generated once on first access

## Benchmarks

Measured with realistic generated Svelte components of various sizes:

| Component size | Without `.map` access | With `.map` access | Savings |
|---|---|---|---|
| Small (30 lines) | 0.61ms | 0.60ms | ~0% (map gen is trivial) |
| Medium (100 lines) | 1.50ms | 1.82ms | **17%** |
| Large (300 lines) | 4.50ms | 5.02ms | **10%** |
| XL (600 lines) | 8.93ms | 9.98ms | **11%** |

## Test plan

- [x] All 365 svelte2tsx tests pass (including 16 sourcemap tests)
- [x] All 665 language-server tests pass

### DIFF
--- a/.changeset/seven-jokes-whisper.md
+++ b/.changeset/seven-jokes-whisper.md
@@ -1,5 +1,5 @@
 ---
-"svelte2tsx": patch
+'svelte2tsx': patch
 ---
 
 perf: defer source map generation until first access


jasonlyu123:
This won't make any difference. The map property is accessed immediately.

PR #2965, opened 2026-03-02
fix(emitDts): drop declarations emitted outside declarationDir

### LINKED ISSUE #2182: Emit dts not respecting abs path
### Describe the bug

When I use `emitDts` with an absolute path, and I am in a subfolder from my project it weirdly saves it to a path like this: `/workspace/extractinator/workspace/extractinator/.temp`

```ts
await emitDts({
	svelteShimsPath: require.resolve('svelte2tsx/svelte-shims-v4.d.ts'),
	declarationDir: '/workspace/extractinator/.temp',
	libRoot: '/workspace/extractinator/playground',
})
```		

### Reproduction

I can provide a repo if this isn't just user error or expected - which I worry it is

### Expected behaviour

It to save it at the given path

### System Info

- OS: ubuntu docker
- IDE: vscode

svelte2tsx

_No response_

joeinnes:
Fixes #2182

## Problem

When a Svelte component imports a TypeScript module from outside the configured `libRoot`, TypeScript follows the import and attempts to emit a declaration file for that module. Because `emitDts` forces `rootDir: config.libRoot`, the imported file is outside `rootDir`, and TypeScript falls back to emitting its declaration using the file's absolute path as the output location — rather than placing it inside `declarationDir`. This pollutes the source tree with unwanted `.d.ts` files.

A second related bug compounds this: the `writeFile` hook unconditionally prepended `pathPrefix` (a relative path) to TypeScript's output filename using `path.join`. When TypeScript generated an absolute output path for the out-of-`rootDir` file, `path.join` appended the relative prefix to it, doubling the path and writing files deep inside a fake directory tree within the source tree (as reported in #2182).

## Reproduce

A project where:
- `libRoot` (or `-i`) points to a subdirectory (e.g. `src/`)
- A component in that directory imports a `.ts` file from outside it (e.g. `../utils/helper.ts`)
- `emitDts` (or `svelte-package`) is run

**Result:** a declaration file for the imported module appears outside `declarationDir`, either next to the source or at a doubled absolute path inside the source tree.

## Fix

Two changes to the `writeFile` hook in `createTsCompilerHost`:

1. **Guard `pathPrefix` with `!path.isAbsolute(fileName)`** — only prepend the relative prefix when TypeScript has produced a relative output path. Fixes the doubled-path write (#2182).

2. **Drop `.d.ts`/`.d.ts.map` writes outside `declarationDir`** — after resolving the final output path, silently skip any declaration file whose destination is not within `declarationDir`. This prevents out-of-`rootDir` imports from emitting declarations into the source tree.

## Tests

Added a new `emitDts` sample (`typescript-imports-outside-lib-root`) where a Svelte component imports from outside `libRoot`. The test runner was extended to assert that no `.d.ts` files appear outside `declarationDir` after `emitDts` completes, and to clean up any such files (including newly-created directory trees) if the assertion fails.


jasonlyu123:
Can you provide a reproduction repository? I don't quite understand the problem you're trying to fix.

joeinnes:
Sure, so for context, I'm a maintainer over at [garden-co/jazz](https://github.com/garden-co/jazz), where we're building a database product with a bunch of different capabilities. We offer Svelte bindings to allow users to build their Jazz apps with Svelte. We're running into an annoying issue where we need to run a clean-up step every time we build our repo, because the `.d.ts` files emitted during the Svelte bindings build get spat out into the source directory.

This repro should demo it: https://github.com/joeinnes/emitdts-repro

Clone, install deps, and run the build script (`pnpm` or `npm run build`). You'll see the `.d.ts` file appear at `src/runtime/db.d.ts`, not in `dist/lib` where you'd expect/want it to be.

dummdidumm:
I think this needs some path normalization before comparing them due to windows but otherwise looks good

packages/svelte2tsx/src/emitDts.ts
```diff
@@ -193,7 +193,10 @@ async function createTsCompilerHost(options: any, svelteMap: SvelteMap, absDecla
             // This happens when TypeScript follows imports to source files outside
             // rootDir and falls back to emitting declarations next to the source.
             if (fileName.endsWith('.d.ts') || fileName.endsWith('.d.ts.map')) {
-                if (!path.resolve(fileName).startsWith(absDeclarationDir + path.sep)) {
+                const resolved = path.resolve(fileName);
+                const normalise = (p: string) =>
+                    ts.sys.useCaseSensitiveFileNames ? p : p.toLowerCase();
+                if (!normalise(resolved).startsWith(normalise(absDeclarationDir + path.sep))) {
                     return;
                 }
```
dummdidumm, line 199:
```suggestion
                const resolved = path.resolve(fileName).replace(/\\/g, '/');
                if (!resolved.startsWith(absDeclarationDir.replace(/\\/g, '/') + '/'))) {
                    return;
                }
```

dummdidumm (approved):
thank you!

packages/svelte2tsx/src/emitDts.ts
```diff
@@ -179,7 +184,26 @@ async function createTsCompilerHost(options: any, svelteMap: SvelteMap) {
...
+            // Drop declaration files emitted outside declarationDir.
+            // This happens when TypeScript follows imports to source files outside
+            // rootDir and falls back to emitting declarations next to the source.
+            if (fileName.endsWith('.d.ts') || fileName.endsWith('.d.ts.map')) {
+                const resolved = path.resolve(fileName);
+                const normalise = (p: string) =>
+                    ts.sys.useCaseSensitiveFileNames ? p : p.toLowerCase();
+                if (!normalise(resolved).startsWith(normalise(absDeclarationDir + path.sep))) {
+                    return;
+                const resolved = path.resolve(fileName).replace(/\\/g, '/');
```
dummdidumm, line 196:
```suggestion
                const resolved = path.resolve(fileName).replace(/\\/g, '/');
                if (!resolved.startsWith(absDeclarationDir.replace(/\\/g, '/') + '/'))) {
                    return;
                }
```

packages/svelte2tsx/src/emitDts.ts
```diff
@@ -179,7 +184,21 @@ async function createTsCompilerHost(options: any, svelteMap: SvelteMap) {
...
+                pathPrefix && !path.isAbsolute(fileName)
+                    ? path.join(pathPrefix, fileName)
+                    : fileName;
+
+            // Drop declaration files emitted outside declarationDir.
+            // This happens when TypeScript follows imports to source files outside
+            // rootDir and falls back to emitting declarations next to the source.
+            if (fileName.endsWith('.d.ts') || fileName.endsWith('.d.ts.map')) {
+                const resolved = path.resolve(fileName).replace(/\\/g, '/');
+                if (!resolved.startsWith(absDeclarationDir.replace(/\\/g, '/') + '/'))) {
```
dummdidumm:
```suggestion
                if (!resolved.startsWith(absDeclarationDir.replace(/\\/g, '/') + '/')) {
```

PR #2966, opened 2026-03-02
fix: use Promise<Response> for async kit handler return types

MathiasWP:
## Summary

When svelte-check adds type annotations to SvelteKit route handlers (GET, POST, PUT, PATCH, DELETE, etc.), it uses `Response | Promise<Response>` as the return type for all functions. However, TypeScript requires async functions to return `Promise<T>`, not `T | Promise<T>` (TS1064).

This caused **false positive errors in `--incremental` mode**, where tsc runs directly on the generated overlay files. In non-incremental mode these errors were silently filtered out by the TypeScript plugin's `inGenerated` diagnostic filtering, but in incremental mode tsc reports them directly.

### The bug

```typescript
// Original code
export async function POST({ request }) { ... }

// Generated overlay (before fix) - triggers TS1064
export async function POST({ request }: RequestEvent) : Response | Promise<Response> { ... }

// Generated overlay (after fix) - correct
export async function POST({ request }: RequestEvent) : Promise<Response> { ... }
```

### Changes

- **`packages/svelte2tsx/src/helpers/sveltekit.ts`**: In `addTypeToFunction`, check if the function has an `async` modifier. If so, extract just the `Promise<T>` part from the return type union instead of using the full `T | Promise<T>`.
- **`packages/svelte2tsx/test/helpers/index.ts`**: Updated the async GET test to expect `Promise<Response>` and added a new test for non-async GET to verify it still gets `Response | Promise<Response>`.
- **`packages/svelte-check/test-success/`**: Added an async `+server.ts` fixture with GET and POST handlers plus the necessary `RequestEvent` type to the sanity test suite.

## Test plan

- [x] Updated unit test `upserts GET async function` fails without fix, passes with fix
- [x] New unit test `upserts GET non-async function` passes (non-async still uses `Response | Promise<Response>`)
- [x] All 366 svelte2tsx tests pass
- [x] All 6 svelte-check sanity tests pass (including incremental cold/warm cache)

### DIFF
--- a/packages/svelte2tsx/src/helpers/sveltekit.ts
+++ b/packages/svelte2tsx/src/helpers/sveltekit.ts
@@ -283,7 +283,8 @@ function upsertKitRouteFile(
             isTsFile,
             name,
             `import('./$types.js').RequestEvent`,
-            `Response | Promise<Response>`
+            `Response | Promise<Response>`,
+            `Promise<Response>`
         );
     };
     insertApiMethod('GET');
@@ -476,7 +477,8 @@ function addTypeToFunction(
     isTsFile: boolean,
     name: string,
     type: string,
-    returnType?: string
+    returnType?: string,
+    asyncReturnType?: string
 ) {
     const fn = exports.get(name);
     if (fn?.type === 'function' && fn.node.parameters.length === 1 && !fn.hasTypeDefinition) {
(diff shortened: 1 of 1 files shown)


MathiasWP:
We need this fix to be able to use the `--incremental` flag with svelte-check

PR #2978, opened 2026-03-14
feat: pull diagnostics support

jasonlyu123:
This switches the diagnostics from push model (server to client) to the [pull model](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_pullDiagnostics) when the client supports it. The advantage is that the client can prioritise sending the diagnostic request for the visible tab. With this, we can check the document split to the side immediately and don't need to check all open files after a file save. The pull mode has a way to tell the client that the diagnostics haven't changed. I implement this mostly because VSCode sends another diagnostic request for the same files when switching files in the editor, even if the file isn't changed. 

I also lowered the debounce delay in the push-mode diagnostics. I've been wondering if some of the performance complaints are actually just because the artificial delay is too long for some people. The pull mode also has a much lower artificial delay. I think we can give it a try. If people are complaining about it, we can switch to an approach similar to the "smart" mode.

### DIFF
--- a/packages/language-server/test/plugins/PluginHost.test.ts
+++ b/packages/language-server/test/plugins/PluginHost.test.ts
@@ -100,7 +100,7 @@ describe('PluginHost', () => {
         );
     });
 
-    describe.only('pull mode diagnostics', () => {
+    describe('pull mode diagnostics', () => {
         it('merge pull diagnostics results', async () => {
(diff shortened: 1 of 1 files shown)


dummdidumm (approved):
great work!

PR #2985, opened 2026-03-24
feat: typescript 6.0 support

jasonlyu123:
### TypeScript is now a peerDependency
For users manually installing the language server, this includes VSCode users using the "svelte.language-server.ls-path" config. You also need to install TypeScript for the language server if it's not automatically installed.

### TypeScript 6
See https://devblogs.microsoft.com/typescript/announcing-typescript-6-0/#breaking-changes-and-deprecations-in-typescript-6.0

Most notably:
- [types now defaults to []](https://devblogs.microsoft.com/typescript/announcing-typescript-6-0/#types-now-defaults-to-[])
- [strict is on by default](https://devblogs.microsoft.com/typescript/announcing-typescript-6-0/#simple-default-changes). If you want to disable it. Add `strict: false` to your tsconfig.json/jsconfig.json.

## PR details
This changes our moduleResolution config to node16 or bundler. It's mostly just some config changes. Only problem is `estree-walker`. The types for `estree-walker` v2 can't be resolved under `node16`, but v3 is ESM only, so copying the type for now. 

As for bundling TypeScript 6, this is definitely a breaking change for the extension. Is there any other breaking change we want to ship with this?  ~~And what about language-server? Is this considered a breaking change if both are allowed? Actually don't know if `||` works in `dependencies`.~~ This will also move typescript to `peerDependency`, so it'll be a breaking change.


packages/language-server/tsconfig.json
```diff
@@ -12,6 +12,8 @@
         "sourceMap": true,
         "composite": true,
         "skipLibCheck": true,
-        "forceConsistentCasingInFileNames": true
+        "rootDir": ".",
+        "forceConsistentCasingInFileNames": true,
+        "types": ["@types/mocha"]
```
teemingc: Do we need to add “node” here too?

packages/svelte2tsx/test/tsconfig.json
```diff
@@ -2,9 +2,10 @@
     "compilerOptions": {
         "types": ["@types/node", "@types/mocha"],
         "target": "es6",
-        "module": "commonjs",
-        "moduleResolution": "node",
-        "esModuleInterop": true
+        "module": "nodenext",
+        "moduleResolution": "nodenext",
+        "esModuleInterop": true,
+        "rootDir": "."
```
teemingc: We may not need this as it’s the default in ts6

teemingc: Oh, weird. Good catch

jasonlyu123:
About the failed test. I narrowed down the problem to `svelteHTML.mapElementTag`. For some reason, TypeScript try to create a union for every property of HTMLElement to its derived types. For example, `HTMLElement.hidden` and `HTMLDivElement.hidden`. This adds around 10ms (about 30%) on my machine for each iteration of the performance test.  But it only runs once during the type-checker's lifetime, i.e. every keystroke. 10ms every keystroke is not small, but it's not really that noticeable. I'll still try to create an issue in TypeScript, but I don't think this is a major blocker. 

Edit: Just noticed TypeScript team has updated the README with a merge criteria for the old codebase. Guess we have missed the beta/RC phase to get this fixed 😅.

jasonlyu123:
Draft until the next release. Wanted to release the current patches before we update the extension to use TypeScript 6.

packages/language-server/package.json
```diff
@@ -62,7 +62,7 @@
         "prettier-plugin-svelte": "^3.5.0",
         "svelte": "^4.2.19",
         "svelte2tsx": "workspace:~",
-        "typescript": "^5.9.2",
+        "typescript": "^5.9.2 || ^6.0.2",
```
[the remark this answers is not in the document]
dummdidumm: peer dep sounds good, though we gotta do language-server package as a minor then

packages/svelte2tsx/tsconfig.json
```diff
@@ -1,5 +1,6 @@
 {
     "compilerOptions": {
+        "strict": false,
```
dummdidumm, line 3: one day we will mark this strict and fight all the errors that come from it 😆

dummdidumm:
> I narrowed down the problem to svelteHTML.mapElementTag.

Mhhhm can we adjust the type in a way that it's not that costly anymore? Feels kinda bad that it will regress this much per keystroke.

jasonlyu123:
> Mhhhm can we adjust the type in a way that it's not that costly anymore

Maybe we could still try our luck to open an issue. Not entirely sure why it does that. `document.createElement` doesn't, on the other hand. And it also doesn't appear in typescript-go, although the hot path has a corresponding code in typescript-go. I tried debugging with the debugger, but the same `svelteHTML.mapElementTag` doesn't trigger the problem during type-check.

jasonlyu123:
Pretty weird that the performance problem also happened in typescript go on another PC. I created 
https://github.com/microsoft/TypeScript/issues/63347. Hope at least it can be improved in typescript go.

jasonlyu123:
Hmm. It actually isn't caused by runtime code changes in TypeScript 6, but the DOM definition update. Might be an existing problem that gets triggered by the DOM update. 

TypeScript 6 is also noticeably faster in some areas. Besides the performance test, our language server test is reliably faster with TypeScript 6. I also just noticed that the benchmark in the performance test is faster, so part of the reason it failed is that the target time has also decreased.

jasonlyu123:
I'll merge this next week if there's no objection. The upstream issue for the `mapElement` problem mentioned is closed as "Not a defect", so it's probably not possible to fix the root cause of the problem in TypeScript 6.

PR #2987, opened 2026-03-24
chore: replace `cross-env` with pnpm shell emulator

teemingc:
This PR removes `cross-env` in favour of pnpm's shellEmulator which "just works". We're currently using this in the svelte, kit, and svelte.dev repositories too.

https://pnpm.io/settings#shellemulator


package.json
```diff
@@ -7,7 +7,7 @@
     "scripts": {
         "bootstrap": "cd ./packages/svelte2tsx && pnpm build && cd ../svelte-vscode && pnpm build:grammar",
         "build": "tsc -b",
-        "test": "cross-env CI=true pnpm test -r",
+        "test": "CI=true pnpm run -r test",
```
teemingc, line 10: Specifying the `pnpm run` command fixes an endless recursion because it didn't exclude the root workspace test command.
See https://pnpm.io/cli/recursive

jasonlyu123:
The test failure should be due to the magic-string update in svelte2tsx.

teemingc:
> The test failure should be due to the magic-string update in svelte2tsx.

Thanks for sharing. I've been wondering why removing cross-env would cause those tests to fail 😅

PR #2990, opened 2026-03-25
fix: resolve svelte files under NodeNext

### LINKED ISSUE #2984: Missing component types on `--tsgo`
### Describe the bug

Both `svelte-check` and `svelte-fast-check` works, but `svelte-check --tsgo` generates errors like:

```
/workspaces/slowreader/web/ui/navbar/index.svelte:165:16
Error: Parameter 'e' implicitly has an 'any' type. (ts)
      icon={mdiMenu}
      onclick={e => {
        if (!openMenu('other')) e.preventDefault()
```

Also, `svelte-check --tsgo` doesn't see any missed component’s props.

I assume it means that components has no types and this is why callback argument is `any`.

### Reproduction

```bash
git clone git@github.com:hplush/slowreader.git
cd slowreader
pnpm install
cd web
svelte-check --tsgo
```

### Expected behaviour

No errors like in `svelte-check` and `svelte-fast-check`

### System Info

- OS: Fedora 43 inside Dev Container on Fedora 43
- IDE: Zed, but we have this error in normal terminal

```bash
$ pnpm svelte-check --version
svelte-check, 4.4.5
```

svelte-check

`tsconfig.json`:

```json
{
  "compilerOptions": {
    "allowImportingTsExtensions": true,
    "noUncheckedIndexedAccess": true,
    "experimentalDecorators": true,
    "noImplicitOverride": true,
    "verbatimModuleSyntax": true,
    "moduleResolution": "NodeNext",
    "erasableSyntaxOnly": true,
    "esModuleInterop": true,
    "skipLibCheck": true,
    "allowJs": true,
    "noEmit": true,
    "target": "ES2024",
    "module": "NodeNext",
    "strict": true,
    "lib": ["ES2024", "DOM", "DOM.Iterable"],
    "types": ["trusted-types"]
  },
  "include": ["**/*.ts", "**/*.cts", "**/*.svelte", ".storybook/*.ts"],
  "exclude": ["dist", "storybook-static"]
}
```

An example of component usage with error:

```svelte
<script lang="ts">
  import NavbarButton from '../navbar/button.svelte'
</script>

<NavbarButton
  hasSubmenu="navbar_submenu"
  onclick={e => {
    e.preventDefault()
  }}
  size="icon"
/>
```

The component at `web//ui/navbar/button.svelte`

```svelte
<script lang="ts">
  import type { MouseEventHandler } from 'svelte/elements'

  let {
    href,
    name,
    onclick
  }: {
    href?: string
    name: string
    onclick?: MouseEventHandler<HTMLAnchorElement | HTMLButtonElement>
  } = $props()
</script>

<a {href} {onclick}>
  {name}
</a>
```

Compiled version of component from `web/.svelte-check/svelte/ui/navbar/++button.svelte.ts`:

```ts
;
import type { MouseEventHandler } from 'svelte/elements';

;type $$ComponentProps =  {
    href?: string
    name: string
    onclick?: MouseEventHandler<HTMLAnchorElement | HTMLButtonElement>
  };function $$render() {

  

  let {
    href,
    name,
    onclick
  }:/*Ωignore_startΩ*/$$ComponentProps/*Ωignore_endΩ*/ = $props()
;
async () => {

 { svelteHTML.createElement("a", { href,onclick,});
  name;
 }
};
return { props: {} as any as $$ComponentProps, exports: {}, bindings: __sveltets_$$bindings(''), slots: {}, events: {} }}
const Button__SvelteComponent_ = __sveltets_2_fn_component($$render());
/*Ωignore_startΩ*/type Button__SvelteComponent_ = ReturnType<typeof Button__SvelteComponent_>;
/*Ωignore_endΩ*/export default Button__SvelteComponent_;
```

Also `web/.svelte-check/svelte/ui/navbar/button.svelte.d.ts`

```ts
export { default } from "./++button.svelte.ts";
export * from "./++button.svelte.ts";
```

jasonlyu123:
fix #2984. The `.svelte.d.ts` approach doesn't work in NodeNext because NodeNext doesn't allow extensionless import, so TypeScript never tries to append the `.d.ts` extension.

### DIFF
--- a/packages/svelte-check/src/incremental.ts
+++ b/packages/svelte-check/src/incremental.ts
@@ -55,7 +55,7 @@ export type ParsedDiagnostic = {
     message: string;
 };
 
-const MANIFEST_VERSION = 2;
+const MANIFEST_VERSION = 3;
 const SVELTE_KIT_DIR = '.svelte-kit';
 const CACHE_DIR_NAME = '.svelte-check';
 const EMIT_SUBDIR = 'svelte';


dummdidumm (commented):
We need to update the manifest version because this changes the info in it, otherwise lgtm 👍

PR #2998, opened 2026-04-02
fix: hoist self-referenced props interface

### LINKED ISSUE #2935: Adding a `<script module>`, even without any content, causes snippet's argument typing to fail
### Describe the bug

I have a component where I am assigning a type like this:

```svelte
{#snippet positionSnippet({ title, positions }: Position)}
    ...
{/snippet}
```

This is for a component that renders org charts, so there's self-referencing types.

Previously this was working fine without errors. Recently I add a `<script lang="ts" module>` block to the component to export an unrelated object. Doing so somehow broke the types in the snippet underneath it, giving the following error:

```
Cannot find name 'Position'.
```

These were some of the ways to get the error to go away:

1. Move the type definition of `Position` into the `<script module>` block
2. Remove the self-referential type definition in `Position` which otherwise worked
3. Remove the destructuring of the argument inside the snippet

What I find really strange is that the use of a `<script module>` block caused the types in my component to fail, which were otherwise working fine before this.

### Reproduction

I made a simple repro [here on StackBlitz](https://stackblitz.com/edit/sveltejs-kit-template-default-mdbyxv8e?file=src%2Flib%2FOrg.svelte).

To reproduce:

1. `npm run check` to verify types work as expected in `Org.svelte`
2. Uncomment lines 1-3 in `Org.svelte` to enable the `<script lang="ts" module>` block
3. `npm run check` to see the type error

### System Info

From the Stackblitz repro:

```
  System:
    OS: Linux 5.0 undefined
    CPU: (8) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 20.19.1 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 10.8.2 - /usr/local/bin/npm
    pnpm: 8.15.6 - /usr/local/bin/pnpm
  npmPackages:
    svelte: ^5.49.2 => 5.50.0
```

### Severity

annoyance

jasonlyu123:
fix #2935

### DIFF
--- a/.changeset/fruity-plants-try.md
+++ b/.changeset/fruity-plants-try.md
@@ -1,0 +1,5 @@
+---
+'svelte2tsx': patch
+---
+
+fix: hoist self-referenced props interface
--- a/packages/svelte2tsx/src/svelte2tsx/nodes/HoistableInterfaces.ts
+++ b/packages/svelte2tsx/src/svelte2tsx/nodes/HoistableInterfaces.ts
@@ -146,21 +146,24 @@ export class HoistableInterfaces {
                         member.type,
                         type_dependencies,
                         value_dependencies,
-                        generics
+                        generics,
+                        interface_name
                     );
                 } else if (ts.isIndexSignatureDeclaration(member)) {
                     this.collectTypeDependencies(
                         member.type,
                         type_dependencies,
                         value_dependencies,
-                        generics
+                        generics,
+                        interface_name
                     );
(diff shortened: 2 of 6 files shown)


dummdidumm (approved):
👍

PR #3009, opened 2026-04-23
feat: add support for svelte config ts/mts files

### LINKED ISSUE #2834: Allow usage of `svelte.config.ts`
### Description

Per https://github.com/sveltejs/kit/pull/13935 loading `svelte.config.ts` is supported. However, when I use such a file in my project it seems to be undetected by both my IDE tooling and when running `svelte-check`. 

### Proposed solution

A quick look lead me to this:
https://github.com/sveltejs/language-tools/blob/2c99c58a62b9f1d0d6c4736d171692f2ba226c73/packages/language-server/src/lib/documents/configLoader.ts#L222-L225

I suspect altering this is a requirement for compatibility, but may not be sufficient. I have not yet had any time to dive into this further, so any assistance is welcome.

### Alternatives

_No response_

*Edit*: progress on resolving this can be found at https://github.com/sveltejs/language-tools/pull/2804.

craig-jennings:
Fixes #2834 

This is a follow-up to #2804 that was closed for some reason. It'd be great to get this in because I think it's the last blocker to being able to fully adopt `svelte.config.ts` files in a project. The only adjustment I made was adding @jrmajor's recommendation to remove the flag for Node 22.18 because it was [enabled by default](https://github.com/nodejs/node/blob/00a42d82053e2f59ef4130cd0fe9a85b623a64be/doc/changelogs/CHANGELOG_V22.md#2025-07-31-version-22180-jod-lts-aduh95) on that release

### DIFF
--- a/packages/language-server/src/lib/documents/configLoader.ts
+++ b/packages/language-server/src/lib/documents/configLoader.ts
@@ -143,7 +143,12 @@ export class ConfigLoader {
                 return this.fs.existsSync(path) ? path : undefined;
             };
             const configPath =
-                tryFindConfigPath('js') || tryFindConfigPath('cjs') || tryFindConfigPath('mjs');
+                tryFindConfigPath('js') ||
+                tryFindConfigPath('ts') ||
+                tryFindConfigPath('cjs') ||
+                tryFindConfigPath('mjs') ||
+                tryFindConfigPath('mts');
+
             if (configPath) {
                 return configPath;
             }
--- a/packages/language-server/src/lib/documents/configLoader.ts
+++ b/packages/language-server/src/lib/documents/configLoader.ts
@@ -47,10 +47,10 @@ const _dynamicImport = new Function('modulePath', 'return import(modulePath)') a
     modulePath: URL
 ) => Promise<any>;
 
-const configRegex = /\/svelte\.config\.(js|cjs|mjs)$/;
+const configRegex = /\/svelte\.config\.(js|ts|cjs|mjs|mts)$/;
 
 /**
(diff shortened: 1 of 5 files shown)


jasonlyu123:
~~The flag is still needed. The language server is using Node.js version bundled by VSCode by default. And the current supported VSCode versions still include some that need the flag.~~

jasonlyu123:
Sorry, never mind, I misunderstood what you meant. 

I think we also don't need to add the flag when the `svelte.language-server.runtime` config is set. We won't know the version of the custom node.js version, and users can still use the `svelte.language-server.runtime-args` config to add it.

craig-jennings:
@jasonlyu123 I believe I addressed your comments in the latest commit. When you have a sec, could you take another look?

packages/language-server/src/lib/documents/configLoader.ts
jasonlyu123: This also needs to be updated.

packages/svelte-vscode/src/extension.ts
jasonlyu123: Another one here

jasonlyu123 (changes requested):
Found that Node.js has an API for checking if `--experimental-strip-types` is enabled. Can add a check to the configLoader to load ts files if it's enabled.
```ts
const loadSvelteConfigTs = 'typescript' in process.features && !!process.features.typescript
```

craig-jennings:
@jasonlyu123 Not quite sure I'm following where you're wanting `loadSvelteConfigTs` check to go, so feel free to update this PR as you see fit.

jasonlyu123 (approved):

PR #3010, opened 2026-04-24
fix: ./$types imports were getting improperly replaced

ottomated:
svelte2tsx compiles `const { params } = $props();` into an import reference like 
`/*Ωignore_startΩ*/;type $$ComponentProps = { params: import('./$types.js').PageProps['params'] };/*Ωignore_endΩ*/`.

When renaming a file, the `import('./$types.js')` would be replaced with the path to the old file location (i.e. `'./routes/something/$types.js'`, and because the svelte2tsx code is generated, that new string would be injected in the wrong place, causing issues like this:

```svelte
<script>
  import './foo';./routes/something/$types.js

  const { params } = $props();
</script>
```

The root cause is the `isKitTypePath` function no longer working as expected. I replaced it with a check that the string ends with `$types`, but I'm not 100% confident in that fix. This PR does fix the error, though.

### DIFF
--- a/packages/language-server/src/plugins/typescript/features/UpdateImportsProvider.ts
+++ b/packages/language-server/src/plugins/typescript/features/UpdateImportsProvider.ts
@@ -118,13 +118,13 @@ export class UpdateImportsProviderImpl implements UpdateImportsProvider {
                     // If there is a better solution for this, please file a PR :)
                     change.fileName = change.fileName.replace(oldPathTsProgramCasing, newPath);
                 }
-                change.textChanges = change.textChanges.filter(
-                    (textChange) =>
+                if (path.basename(change.fileName).startsWith('+')) {
+                    change.textChanges = change.textChanges.filter((textChange) => {
                         // Filter out changes to './$type' imports for Kit route files,
                         // you'll likely want these to stay as-is
-                        !isKitTypePath(textChange.newText) ||
-                        !path.basename(change.fileName).startsWith('+')
-                );
+                        return !isKitTypePath(textChange.newText);
+                    });
+                }
                 return change;
             });
 
--- a/packages/language-server/src/plugins/typescript/features/UpdateImportsProvider.ts
+++ b/packages/language-server/src/plugins/typescript/features/UpdateImportsProvider.ts
@@ -77,7 +77,6 @@ export class UpdateImportsProviderImpl implements UpdateImportsProvider {
             return;
         }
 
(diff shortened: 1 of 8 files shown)


packages/language-server/src/plugins/typescript/features/utils.ts
```diff
@@ -356,7 +356,8 @@ export function gatherDescendants<T extends ts.Node>(
 export const gatherIdentifiers = (node: ts.Node) => gatherDescendants(node, ts.isIdentifier);
 
 export function isKitTypePath(path?: string): boolean {
-    return !!path?.includes('.svelte-kit/types');
+    if (!path) return false;
+    return /\$types(?:\.[cm]?[jt]s)?$/.test(path);
```
jasonlyu123: This regex is too complex for what it actually does. It would be more readable and faster to just use multiple `endsWith` checks. Besides, there should only be `$types.js`, `$types.d.ts` or no extension. Why would .cts and .mts exist?

packages/language-server/src/plugins/typescript/features/UpdateImportsProvider.ts
```diff
@@ -77,6 +77,7 @@ export class UpdateImportsProviderImpl implements UpdateImportsProvider {
             return;
         }
 
+        console.log(ls.getProgram()?.getSourceFile(oldPath)?.getText());
```
jasonlyu123: This should be removed

jasonlyu123 (changes requested):

packages/language-server/src/plugins/typescript/features/utils.ts
ottomated: I wrote this to futureproof, but I managed to write a test that should cover it instead.

packages/language-server/test/plugins/typescript/features/UpdateImportsProvider.test.ts
```diff
@@ -51,7 +51,7 @@ describe('UpdateImportsProviderImpl', function () {
 
     it('updates imports', async () => {
         const { updateImportsProvider, fileUri } = await setup(
-            'updateimports.svelte',
```
jasonlyu123, line 54: Instead of renaming the test file, I think it's better to add separate files.

jasonlyu123:
For the record, this will make it detect only in the routes files, not other files in the routes directory. But it's not clear what the intention of the rename is. It's also possible that someone wanted to access the `./$types` before the rename. Plus, we also check it like this in the typescript-plugin.

PR #3014, opened 2026-04-30
fix(svelte-check): flush stdout/stderr before exit

### LINKED ISSUE #3013: `svelte-check` exits before writing the full output to pipe when there are a large number of diagnostics
### Describe the bug

When stdout is a non-TTY pipe (i.e. CI), the output will sometimes be cut off before the end.

### Reproduction

1. Run `svelte-check` against a project that emits **≥1000** `ERROR`/`WARNING` diagnostics.                                                                                                                                                                                                      
2. Invoke as `svelte-check --output machine-verbose ` with stdout piped (not a TTY).
3. Inspect the captured stdout. Sometimes the output ends mid-stream.

### Expected behaviour

The `COMPLETED` summary line should always be outputted.

### System Info

- OS: Linux/Ubuntu (CI)
- IDE: CLI

svelte-check

_No response_

jackgdll:
Resolves: #3013 

`process.stdout.write` is asynchronous on non-TTY pipes, and per Node's docs `process.exit` does not wait for queued I/O. Under heavy diagnostic load (1000+ ERROR/WARNING lines, enough to saturate the kernel pipe buffer) the trailing `COMPLETED` summary of `--output machine-verbose` was dropped when the CLI exited, even though the run was clean.

Replace the bare `process.exit(N)` calls in both CLI completion paths (default and `useVirtualFiles`) with the canonical flush-then-exit pattern.

Add a regression test `test-flush.js` (`pnpm run test:flush`) that runs the CLI against a fixture emitting 1500 diagnostics, throttles the consumer to keep the pipe full, and asserts the last non-empty line matches `^\d+ COMPLETED` across 10 runs. This is useful for testing this fix but I don't know if it's worth actually commiting it to the repo or not. 

`test-flush.js` is AI generated and manually reviewed.

### DIFF
--- a/.changeset/chubby-drinks-occur.md
+++ b/.changeset/chubby-drinks-occur.md
@@ -2,4 +2,4 @@
 'svelte-check': patch
 ---
 
-flush stdout/stderr before exit
+fix: flush stdout/stderr before exit
--- a/.changeset/chubby-drinks-occur.md
+++ b/.changeset/chubby-drinks-occur.md
@@ -1,0 +1,5 @@
+---
+'svelte-check': patch
+---
+
+flush stdout/stderr before exit
--- a/packages/svelte-check/package.json
+++ b/packages/svelte-check/package.json
@@ -37,10 +37,11 @@
         "typescript": ">=5.0.0"
     },
     "scripts": {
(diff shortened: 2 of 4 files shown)


jasonlyu123:
Thanks you. I removed the test from the build script since it already serves the verification purpose, and it's unlikely to break in the future. Still keeping the test file so we can run it manually.

Issue #3001, opened 2026-04-06
Preview Icon contrast issues


sacrosanctic:
### Describe the bug

> https://github.com/sveltejs/language-tools/pull/2993#issuecomment-4149372496

It doesn't make sense to swap them around. It creates another contrast issue. And the icons between JS/CSS are not cohesive.

### Which package is the issue about?

Svelte for VS Code extension

### Additional Information, eg. Screenshots

Proposed changes:

<img width="338" height="107" alt="Image" src="https://github.com/user-attachments/assets/c1148fdd-6f2b-4931-961c-500ff0dfbe3e" />

<img width="338" height="107" alt="Image" src="https://github.com/user-attachments/assets/b726ca7b-9336-450b-b94f-3e93573198f9" />

teemingc:
For reference as to how it looks like now: 

<img width="290" height="80" alt="svelte compiled css button in vs code" src="https://github.com/user-attachments/assets/4a0159a0-d366-453c-86b4-565579fca57f" />

jasonlyu123:
> It creates another contrast issue. 

Could you elaborate more on what exactly is the problem here? Don't quiet get this one.

Agreeing with the icon is not really cohesive. I actually created a version of this feature a while ago and Icon is one of the main reason I didn't create a PR. The original icon is from vscode's icon pack and I also can't find a suitable one in the icon pack. About the proposal, I think the css/js text is a bit too small so it take more effort to differentiate between the two icon. Not sure if/how it can be tweak to be more clear.

sacrosanctic:
> Could you elaborate more on what exactly is the problem here? Don't quiet get this one.

Because they're swapped and they're filled designs unlike the rest of the set which are outlines. The dark mode one shines like a lighthouse, and the light mode one looks like a black hole. It also doesn't help that only #FFF and #000 values were used.

Admittedly, those designs are not that good either. Just wanted to start the conversation about it.

WhyAsh5114:
Hey there, when I was working on this, I also was concerned about the lack of consistency from these icons. But what I thought was it'd look much better if the fill and bg were simply flipped. Here's a quick figma mockup:

<img width="700" height="625" alt="Image" src="https://github.com/user-attachments/assets/787b5133-861c-45b6-8106-cacf725dcd0c" />

Lmk if this works, will provide the SVGs for both light and dark.

dummdidumm:
[...]

PR #3019, opened 2026-05-08
fix: tweak "show compiled" icon

jasonlyu123:
#3001. Alternative to #3002

See https://github.com/sveltejs/language-tools/issues/3001#issuecomment-4402842086 for the screenshot.


dummdidumm (approved):
let's go with this one 👍

PR #3021, opened 2026-05-10
fix: use pug syntax for emmet in template lang pug

### LINKED ISSUE #1597: Emmet not using pug syntax in template with lang="pug"
### Describe the bug

When using an Emmet abbreviation in a pug template Emmet does not use pug syntax.

### Reproduction

1. Create template with `lang="pug"`
2. Use any Emmet abbreviation within the template.

### Expected behaviour

When using an Emmet abbreviation in a pug template Emmet should use pug syntax.

### System Info

- VSCode 1.70.2
- Svelte for VS Code v105.21.0

_No response_

_No response_

tkhs0813:
fix #1597

This PR updates Emmet completion handling so that Svelte templates with `lang="pug"` use Pug syntax instead of HTML syntax.

Changes:
- Add a regression test for Emmet completions inside `<template lang="pug">`
- Use `pug` as the Emmet syntax when the template language attribute is `pug`
- Keep the default behavior as `html` for other templates

### DIFF
--- a/.changeset/rotten-gifts-develop.md
+++ b/.changeset/rotten-gifts-develop.md
@@ -1,0 +1,5 @@
+---
+"svelte-language-server": patch
+---
+
+fix: use pug syntax for emmet in template lang pug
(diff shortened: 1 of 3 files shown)


dummdidumm (approved):
thank you!

PR #3022, opened 2026-05-11
fix: support CSS completions in nested style tags

### LINKED ISSUE #2175: Intelisense Is Not Working in ``<svelte:head>`` Tag

tkhs0813:
fix #2175

Fixes CSS completions and hover for nested `<style>` tags.

Previously, the CSS plugin used `document.styleInfo`, which only represents the top-level style tag. As a result, positions inside nested style tags were not treated as being inside a generated CSS document, so completions and hover did not work.

This PR resolves the CSS document based on the current cursor position. If the current position is inside a nested `<style>` tag, the CSS plugin creates a `CSSDocument` using that tag's `styleInfo`.

## Changes
- Add `extractStyleTagAtOffset` to resolve the style tag at the current position using the HTML AST
- Allow `CSSDocument` to be created with a specific `styleInfo`
- Use a position-based CSS document for completions and hover
- Add regression tests for:
  - `<svelte:head><style></style></svelte:head>`
  - a nested `<style>` tag

### DIFF
--- a/packages/language-server/src/plugins/css/CSSPlugin.ts
+++ b/packages/language-server/src/plugins/css/CSSPlugin.ts
@@ -480,8 +480,8 @@ export class CSSPlugin
     }
 
     findDocumentHighlight(document: Document, position: Position): DocumentHighlight[] | null {
-        const cssDocument = this.getCSSDoc(document);
-        if (cssDocument.isInGenerated(position)) {
+        const cssDocument = this.getCSSDocAtPosition(document, position);
+        if (cssDocument) {
             if (shouldExcludeDocumentHighlights(cssDocument)) {
                 return wordHighlightForTag(document, position, document.styleInfo, wordPattern);
             }
@@ -537,7 +537,8 @@ export class CSSPlugin
 
         if (
             isInTag(position, document.scriptInfo) ||
-            isInTag(position, document.moduleScriptInfo)
+            isInTag(position, document.moduleScriptInfo) ||
+            isInTag(position, document.templateInfo)
         ) {
             return null;
         }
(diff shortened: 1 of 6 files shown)


packages/language-server/src/plugins/css/CSSPlugin.ts
```diff
@@ -528,6 +529,21 @@ export class CSSPlugin
...
 
+    private getCSSDocAtPosition(document: Document, position: Position) {
+        const cssDoc = this.getCSSDoc(document);
+
+        if (cssDoc.isInGenerated(position)) {
+            return cssDoc;
+        }
+
+        const offset = document.offsetAt(position);
+        const styleInfo = extractStyleTags(document.getText(), document.html).find(
```
jasonlyu123: It'll be more efficient to use `document.html.findNodeAt` to get the container node, then check if it's a style tag. And we can check if the position is inside script tags using `isInTag` before calling `findNodeAt`. Also, this function should return `null` or `undefined` if no style tag is found, that way we don't need to check `isInGenerated` again later.

tkhs0813: Thanks! I updated this to use `document.html.findNodeAt` and return `null` when the position is not inside a style tag.

[e6ae679](https://github.com/sveltejs/language-tools/pull/3022/commits/e6ae679e7a1393d176f6eb249e62a78087d46bba)

Issue #1538, opened 2022-06-27
Add `--cache` CLI option to svelte-check


oekazuma:
### Description

I use githooks to run ESLint, Prettier, and svelte-check at the same time as commit. 
ESLint and Prettier are fast because they keep a cache, but svelte-check is slow if run every time.

### Proposed solution

That when provided with an option such as `--cache`, svelte-check only process new or changed files.
With reference to [ESLint](https://eslint.org/docs/latest/user-guide/command-line-interface#--cache) and [Prettier](https://prettier.io/docs/en/cli.html#--cache).

dummdidumm:
We can't check changed files only because a change in file A could introduce errors in unchanged file B (if you change a prop name for example). Only thing we can do is to check if we can leverage some of the incremental build features of tsc to at least save time in the program creating time

oekazuma:
I think you are right. I understand that this is not simply an ESLint or Prettier approach.

If you think that faster execution can be addressed in another issue, you are free to close this issue.

Tal500:
> I think you are right. I understand that this is not simply an ESLint or Prettier approach.
> 
> If you think that faster execution can be addressed in another issue, you are free to close this issue.

It's also related to the issue I've just posted now #1584 .

Tal500:
> We can't check changed files only because a change in file A could introduce errors in unchanged file B (if you change a prop name for example). Only thing we can do is to check if we can leverage some of the incremental build features of tsc to at least save time in the program creating time

You don't have to rely on incremental build features of tsc.
(I hope) You can also ask tsc to map you the dependency graph of the files, and then rescan only the affected files.

Tal500:
@dummdidumm , I have found a library that calculate dependency graph for your code: [sverweij/dependency-cruiser](https://github.com/sverweij/dependency-cruiser).

What do you think about the idea (of I already mentioned) - instead of having a full caching compatibility, let just cache:
* hashes of input files
* generate a dependency graph (e.g. by the library I mentioned)

Then, on the next time Svelte Preprocess runs, you can check the file hashes, and for the files that were changed, using the cached dependency graph, check only the relevant files.

Did I miss something? What do you think?

FredrikAugust:
Any updates on this?

PR #3023, opened 2026-05-11
perf: cache resolved kit files settings in incremental manifest

inq:
`loadKitFilesSettings` does a dynamic `import()` of `svelte.config.js` on every `--incremental` run (~137 ms in my profile) just to read 4 paths under `kit.files`. This caches the resolved settings on the manifest, keyed by mtime+size. Manifest version bumped 3→4 to invalidate older caches.

Measured on a SvelteKit app, warm median: 2.13 s → 2.05 s (3 samples each). Loosely related to #1538.

### DIFF
--- a/.changeset/cache-kit-files-settings.md
+++ b/.changeset/cache-kit-files-settings.md
@@ -1,0 +1,5 @@
+---
+'svelte-check': patch
+---
+
+perf: cache `svelte.config.js` settings in incremental manifest, avoiding a dynamic import on every warm `--incremental` run
--- a/packages/svelte-check/src/incremental.ts
+++ b/packages/svelte-check/src/incremental.ts
@@ -29,9 +29,16 @@ type ManifestEntry = {
     addedCode?: InternalHelpers.AddedCode[];
 };
 
+type KitFilesSettingsCache = {
+    mtimeMs: number;
+    size: number;
+    settings: InternalHelpers.KitFilesSettings;
+};
+
 type Manifest = {
     version: number;
     entries: Record<string, ManifestEntry>;
+    kitFilesSettings?: KitFilesSettingsCache;
 };
 
 export type EmitResult = {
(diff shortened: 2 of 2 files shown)


jasonlyu123:
I am not sure this makes sense. As long as any svelte files are updated, the config still needs to be loaded. The svelte.config.js also has configs used by the Svelte compiler. Caching it isn't as useful as you might think.

inq:
Hello! Thank you very much and Sorry for my bad English!

But In these cases, I think it can be meaningful enhancements:

1. **--incremental + no change** -> [`index.ts#L370`](https://github.com/sveltejs/language-tools/blob/master/packages/svelte-check/src/index.ts#L370) only instantiate `SvelteCheck` (and `configLoader`) when `filesNeedingDiagnostics.length > 0`. So when nothing change, my `loadKitFilesSettings` is only code path that load `svelte.config.js`. This case happen a lot in monorepo `pnpm typecheck` — many packages have no change since last run.

2. **`.svelte` changed** -> surprisingly it is slightly faster also! (median of 7 samples each)

| version | warm, no change | warm, 1 file touched |
| --- | ---: | ---: |
| `svelte-check@4.4.6` | 2.15 s | 2.11 s |
| this PR | 2.05 s | 1.99 s |

3. **About compiler config** — `configLoader` and `loadKitFilesSettings` consume different parts of the config. We only cache `kit.files` paths, not `preprocess`/`compilerOptions`, so compiler always get fresh config.

4. **Savings scale with config size** — `svelte-check@4.4.6` re-import full dep graph every run. With heavier config (added `typescript`, `crypto`, `zlib`, `stream`):

| version | warm, no change (heavy config) |
| --- | ---: |
| `svelte-check@4.4.6` | 2.21 s |
| this PR | 2.04 s |

So I think it would be positive in every case! But if you think the scope is too small, you can close this!

Thank you very much!

jasonlyu123:
One problem is that if the `files` config comes from an environment variable, the cache invalidation won't work. This can happen if people use the `files` config to have two sets of routes.

PR #3029, opened 2026-05-21
fix: prevent > in expression affects HTML folding

### LINKED ISSUE #2537: vsc collapse/folding gets confused if < is used in condition
### Describe the bug

If "less than" is used in a svelte conditional statement and the block contains a wrapping <div>, vsc does not display the "v" to allow collapse of the block. Use of anything else works. E.G. {#if step > 1} {#if step == 1}

### Reproduction

    {#if step < 1}
        <div>
            Not collapsible (Neither )
        </div>
    {/if}

### Expected behaviour

    {#if step < 1}
        Collapsible (No <div> present)
    {/if}

    {#if step > 1}
        <div>
            Collapsible (Both )
        </div>
    {/if}

### System Info

Windows
VSCode

Svelte for VS Code extension

Issue - Collapse

jasonlyu123:
fix #2537

### DIFF
--- a/.changeset/petite-actors-spend.md
+++ b/.changeset/petite-actors-spend.md
@@ -1,0 +1,5 @@
+---
+'svelte-language-server': patch
+---
+
+fix: prevent > in expression affects HTML folding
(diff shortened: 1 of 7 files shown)


dummdidumm (approved):
nice! makes foldering much more robust

PR #3030, opened 2026-05-22
Fix slot prop diagnostics inside Svelte 5 snippets

### LINKED ISSUE #2992: "slot" HTML attribute on Svelte 5 component wrappers triggers type error
### Use case

I'm building a Svelte 5 wrapper library for SAP UI5 Web Components (`@ui5/webcomponents`).
Each custom element (e.g. `<ui5-button>`) gets a generated Svelte wrapper so users get autocompletion and type safety.

`<ui5-button>` has a named slot `badge` that expects a `<ui5-button-badge>` element.
I also wrap `<ui5-button-badge>` in a Svelte component `ButtonBadge.svelte`.

### The problem

When using `<ButtonBadge slot="badge">` inside a `{#snippet badge()}` block, `svelte-check` reports a type error:

```
Error: '$$_nottuB0.$$slot_def' is of type 'unknown'. (js)
    {#snippet badge()}
        <ButtonBadge slot="badge"></ButtonBadge>
    {/snippet}

====================================
svelte-check found 1 error and 0 warnings in 1 file
```

The code works correctly at **runtime** — this seems to be purely a type-checking issue.
Or maybe I am doing something wrong?

### Analysis

svelte-check generates $$ComponentProps = Props & Record<string, unknown> for every component.
This causes $$ComponentProps['$$slot_def'] to resolve as unknown (because Record<string, unknown> maps all unknown keys to unknown).

When the LS sees <ButtonBadge slot="badge">, it checks $$slot_def on the component type and gets unknown → type error.

Note: slot is explicitly declared in Props as slot?: string — the error persists regardless.

### Question

In this Web Component integration pattern, slot="badge" is a native HTML attribute for browser slot assignment — not a Svelte slot directive.
The intended DOM output is:

```
  <ui5-button>
    <ui5-button-badge slot="badge" text="3"></ui5-button-badge>
  </ui5-button>
```

Is it expected that slot="..." on a Svelte 5 component triggers $$slot_def checking?
Or should slot declared as a typed prop (slot?: string) be treated as a regular prop — without triggering slot-directive type checking?

### Reproduction

 **Minimal reproduction:** https://github.com/ybbus/my-app

The .d.ts files in src/lib/ represent what svelte-package generates when building a Svelte component library for distribution.

### Expected behaviour

Should work without any type errors when running svelte-check.

###  Environment

  - Windows 11
  - Svelte: 5.54.0
  - svelte-check: 4.4.2
  - @sveltejs/kit: 2.50.2
  - Svelte VS Code extension: 109.14.2
  - TypeScript: 5.9.3

svelte-check

_No response_

Abdulmumin1:
Fixes #2992.

`slot="..."` on a component directly inside a Svelte 5 snippet was being treated as legacy Svelte slot placement. That made the generated code read `$$slot_def` from the child component, which breaks for Svelte 5 wrapper components whose generated props include `Record<string, unknown>`.

In this case, the snippet itself is already the slot API:

```svelte
<Button>
	{#snippet badge()}
		<ButtonBadge slot="badge" />
	{/snippet}
</Button>
```

Here `slot="badge"` should remain a normal prop/attribute on `ButtonBadge`, so wrappers can forward it to a web component.

This breaks the virtual element/component parent chain at the snippet boundary. Nested components inside the snippet still keep their own parent chain, so this continues to work as a Svelte slot:

```svelte
{#snippet badge()}
	<X>
		<Y slot="foo" />
	</X>
{/snippet}
```

Tested with the repro from #2992:

```text
node packages/svelte-check/bin/svelte-check --tsconfig ./tsconfig.json
svelte-check found 0 errors and 0 warnings
```

Also ran:

```text
pnpm --filter svelte2tsx test
pnpm --filter svelte-check build
```

I also temporarily ran `svelte2tsx` tests with `svelte@5.54.0` locally:

```text
804 passing
```


dummdidumm (commented):
Thank you! This isn't quite right yet - if you have this inside a snippet

```svelte
{#snippet asd()}
<X>
  <Y slot="foo" />
</X>
{/snippet}
```

then it also counts as a slot. So we gotta reset the depth in that case.

We don't need this to be an integer btw, I think it's enough to say "you are inside a snippet". I also do wonder if the better change is to break the `parent` chain for elements/components. That change would be in `htmllxtojsx_v2`

Abdulmumin1:
> Thank you! This isn't quite right yet - if you have this inside a snippet
> 
> ```svelte
> {#snippet asd()}
> <X>
>   <Y slot="foo" />
> </X>
> {/snippet}
> ```
> 
> then it also counts as a slot. So we gotta reset the depth in that case.
> 
> We don't need this to be an integer btw, I think it's enough to say "you are inside a snippet". I also do wonder if the better change is to break the `parent` chain for elements/components. That change would be in `htmllxtojsx_v2`

that makes sense. I updated to the pr accordingly

so the direct child case from the issue stays a normal prop:

```svelte
<Button>
	{#snippet badge()}
		<ButtonBadge slot="badge" />
	{/snippet}
</Button>
```

but nested component slot placement inside the snippet still works:

```svelte
{#snippet asd()}
	<X>
		<Y slot="foo" />
	</X>
{/snippet}
```

dummdidumm (approved):
thank you!

PR #3031, opened 2026-05-22
feat: support reading Svelte config from vite.config.js/ts

dummdidumm:
We're thinking about getting rid of `svelte.config.js` (or rather, soft-deprecating it): `vite.config.js/ts` is now pretty much the standard for new Svelte apps and generally a better place to put your config.

This is an important stepping stone towards that future. `svelte.config.js` was originally introduced for language-tools to read config options, now we can do the same by reading the Vite config.

With this PR, `vite.config.js` is read first / preferred if both exist, but if it does not exist or no options are retrieved through it then `svelte.config.js/ts` is used.


dummdidumm:
Pete brought up the good point that reading the Svelte config could be moved into its own pkg.

gregroyal:
Just wanted to point out that this feature caused a major headache for our team. Our top level monorepo vite.config.ts file has extra logic that lets us do special things. This read that file every time when it was previously only run under special top-level conditions (like run the dev servers for all the apps in the monorepo, and proxy connections). We needed a special case so our apps can share a domain to talk between them.

By loading this, it stealth ran our mega dev server behind the scenes taking up our resources eventually crashing the box.

We've adjusted by renaming the file so your plugin doesnt load it and manually pointing to that file, but FYI you might want to change something here...

PR #3036, opened 2026-05-29
feat: svelte-check tsgo support with experimental api

jasonlyu123:
This adds a new `--tsgo-experimental-api` flag in svelte-check for tsgo support, without most of the limitations of the current `--tsgo`. But because this doesn't cache virtual code, it's slightly slower than --tsgo. This flag is considered experimental and might break in the future. 

In case people are wondering. No, the editor support is still not ready. The FS proxy approach this PR depends on isn't available under the api over LSP.


dummdidumm (approved):
great work!

PR #3037, opened 2026-05-29
fix(svelte-check): stop excluding workspaces under dot-prefixed ancestors

### LINKED ISSUE #3034: svelte-check skips every file when project lives under a dot-prefixed ancestor directory (e.g. .claude/, .worktrees/)
### Describe the bug

`svelte-check` silently emits zero virtual `.svelte` / Kit files when the project's absolute path contains any ancestor directory whose name begins with `.` (e.g. `/Users/me/repo/.claude/worktrees/foo/app`).

When `--tsgo` (or `--incremental`) is used the impact is user-visible:

- `+server.ts` handlers with un-annotated parameters (e.g. `export async function GET(req) { … }`) report spurious `Parameter 'req' implicitly has an 'any' type` errors — the annotated virtual copy that normally lives at `.svelte-kit/.svelte-check/svelte/src/routes/**/+server.ts` is never generated, so tsgo never sees the `req: import('./$types.js').RequestEvent` annotation that would otherwise be merged in via `rootDirs`.
- Other Svelte/Kit virtuals are also missing; only the bare shim files (and the overlay `tsconfig.json`) end up in `.svelte-kit/.svelte-check/`.

Without `--tsgo` / `--incremental` the bug is invisible because the non-virtual diagnostic path doesn't rely on those copies for endpoint handler typing.

### Reproduction

1. Place any SvelteKit project at a path that contains a dot-prefixed ancestor. Common cases:
    - `git worktree` under `.claude/worktrees/…` (Claude Code's default), `.worktrees/…`, `.dev/…`
    - Any project under `~/.config/…`, `~/.local/share/…`
2. Add a `+server.ts` with an un-annotated parameter:
    ```ts
    // src/routes/api/foo/+server.ts
    export async function GET(req) {
      return new Response();
    }
    ```
3. Run:
    ```sh
    rm -rf .svelte-kit
    pnpm exec svelte-kit sync
    pnpm exec svelte-check --tsconfig ./tsconfig.json --tsgo
    ```
4. Observe:
    ```sh
    $ find .svelte-kit/.svelte-check -type f | wc -l
    1
    ```
    Only the overlay `tsconfig.json` was emitted; no annotated virtual `+server.ts` / `.d.svelte.ts` files. svelte-check reports `Parameter 'req' implicitly has an 'any' type`.
5. Move the same project to a path with no `/.` segment (e.g. `/tmp/foo/app`) and repeat — the 1228+ virtuals are emitted and the error disappears.

### Expected behaviour

`findFiles()` should crawl the project's own contents regardless of what its ancestor directories happen to be named.

### System Info

- OS: macOS 15 (Darwin 24.6.0)
- node: 24.15.0
- svelte-check: 4.4.8
- @sveltejs/kit: 2.57.1
- @typescript/native-preview (tsgo): 7.0.0-dev.20260421.2
- Triggered originally by paths containing `/.claude/worktrees/…` — the directory the Claude Code CLI creates `git worktree`s in by default. Likely to bite anyone using AI coding tools that nest worktrees, or projects living under `~/.config/…` etc.

svelte-check

### Additional Information

**Root cause** (in `svelte-check@4.4.8`, `dist/src/index.js`, inside `findFiles`):

```js
return new fdir()
    .filter((filePath) => filter(filePath) && !isIgnored(filePath))
    .exclude((_, filePath) => {
        return filePath.includes('/node_modules/') || filePath.includes('/.');
    })
    .withPathSeparator('/')
    .withFullPaths()
    .crawl(workspacePath)
    .withPromise();
```

`filePath` is the absolute path of the directory `fdir` is considering descending into. The `'/.'` substring check was clearly meant to skip hidden directories *inside* the project (`/.git/`, `/.svelte-kit/`, `/.vscode/`, …), but it also matches any ancestor segment that starts with a dot. If the workspace itself lives under `…/.claude/worktrees/foo/app/`, every directory `fdir` considers (including the workspace root) already contains `/.claude/` in its absolute path, so the entire tree is excluded.

`emitSvelteFiles` therefore sees zero candidate files, writes no virtual copies, and tsgo loses the typing layer that the virtuals provide.

**Suggested fix** — check the *directory name being entered* instead of the whole path. `fdir`'s exclude callback already provides the dir name:

```js
.exclude((dirName, _filePath) => {
    return (
        dirName === 'node_modules' ||
        (dirName.startsWith('.') && dirName !== '.' && dirName !== '..')
    );
})
```

If preferred, equivalent form that scopes the dot check to segments below `workspacePath` (handles the edge case where `workspacePath` itself starts with a dot):

```js
const cwd = workspacePath.endsWith('/') ? workspacePath : workspacePath + '/';
.exclude((_, filePath) => {
    if (filePath.includes('/node_modules/')) return true;
    if (!filePath.startsWith(cwd)) return false;
    return filePath.slice(cwd.length).split('/').some((seg) => seg.startsWith('.'));
})
```

Either form preserves the original intent (skip `.git`, `.svelte-kit`, `.vscode`, …) without false-positives on the parent path chain.

**Workaround** — move the project so no ancestor directory name starts with `.`.

spokodev:
\`svelte-check\` silently emitted zero virtual files when the workspace lived under any dot-prefixed ancestor directory (e.g. \`.local/\`, \`.config/\`, custom worktree directories). The user-visible effect is invisible in the default check path but in \`--incremental\` / \`--tsgo\` mode it surfaces as missing endpoint-handler annotations and spurious \`Parameter 'req' implicitly has an 'any' type\` errors on \`+server.ts\` handlers.

The cause is in \`packages/svelte-check/src/utils.ts\` \`findFiles()\`: it routed every directory through \`filePath.includes('/.')\` to skip hidden directories, but that predicate also fires on any ancestor of the workspace whose name starts with a dot, so every descendant under the workspace gets pruned by fdir before the crawl reaches it.

## Change

\`findFiles()\` now exclude-matches by the directory name fdir already passes as the first argument:

\`\`\`ts
.exclude((dirName) => {
    return dirName === 'node_modules' || dirName.startsWith('.');
})
\`\`\`

\`node_modules\` is also matched by name rather than full path for the same ancestor-safety reason. \`createIgnored()\` and the user-supplied \`--ignore\` patterns are unchanged: they still operate on the workspace-relative path inside the \`filter\` predicate.

## Tests

\`packages/svelte-check/test-dot-prefix.js\` is a node-only regression test that:

- creates a fresh temp parent
- nests the workspace under a \`.dotted/\` subdirectory inside that parent
- runs \`bin/svelte-check\` in \`--incremental\` mode against the workspace
- asserts that the Svelte shim (\`Index.d.svelte.ts\`) lands in \`.svelte-check/svelte/src/\`

On \`master\` the shim never appears - findFiles returns an empty list, no emission happens. With the patch the shim is emitted and the regression test passes. The existing \`test:sanity\` (6 cases covering clean / error projects in plain / incremental cold / incremental warm) all still pass.

The new test is wired into \`package.json\` as \`pnpm test:dot-prefix\` next to \`test:sanity\` and \`test:flush\`.

## Verification

\`\`\`
$ pnpm test:dot-prefix
PASS: svelte-check emits Svelte shims under a workspace whose ancestor name starts with a dot

$ pnpm test:sanity
6 passed, 0 failed
\`\`\`

Changeset added (\`.changeset/fix-svelte-check-dot-prefix-ancestor.md\`, \`svelte-check: patch\`).

Fixes #3034


.changeset/fix-svelte-check-dot-prefix-ancestor.md
```diff
@@ -0,0 +1,5 @@
+---
+"svelte-check": patch
+---
+
+Fix `svelte-check` silently emitting zero virtual files when the workspace lives under any dot-prefixed ancestor directory (e.g. `.local/`, `.config/`, custom worktree directories). The directory exclusion in `findFiles()` now matches by directory name instead of full path, so ancestors of the workspace are no longer treated as hidden.
```
dummdidumm:
```suggestion
fix: stop excluding workspaces under dot-prefixed ancestors
```

packages/svelte-check/package.json
```diff
@@ -41,7 +41,8 @@
         "prepublishOnly": "pnpm build",
         "test": "pnpm build",
         "test:sanity": "node test-sanity.js",
-        "test:flush": "node test-flush.js"
+        "test:flush": "node test-flush.js",
+        "test:dot-prefix": "node test-dot-prefix.js"
```
dummdidumm:
```suggestion
        "test:flush": "node test-flush.js"
```

dummdidumm (approved):
thank you! (removed the test because it's overkill)

PR #3038, opened 2026-06-01
perf: remove lodash

ghostdevv:
Remove the dependency on `lodash` replacing with native code and `defu` for merging objects. The only thing I'm not certain is replacing `_.merge` with `defu` as there could be different behavior, but I _think_ it'll be fine. The new tests are AI generated, checked by me, and run both before and after this PR to help ensure I didn't break anything


ghostdevv:
> packages/language-server test:        provide event handler completions in svelte strict mode:
> packages/language-server test:       AssertionError [ERR_ASSERTION]: Expected values to be strictly deep-equal:
> packages/language-server test: + actual - expected
> packages/language-server test:   {
> packages/language-server test: +   newText: 'on:click$2={$1}',
> packages/language-server test: -   newText: 'on:click$2="{$1}"',

~~The test failure seems unrelated to this PR :thinking:~~ no, definitely my fault

dummdidumm (changes requested):
looks good except `groupBy`. groupBy creates an array per entry with all the objects that match the key

ghostdevv:
> looks good except `groupBy`. groupBy creates an array per entry with all the objects that match the key

yea! that's what it should do. It matches the [`Object.groupBy`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/groupBy) behaviour which appears to also match the lodash behaviour

dummdidumm:
Sorry that sentence was unclear - your implementation is _not_ doing what it's supposed to. Instead of creating an array per entry it just takes the first occurrence of a match and assigns it to the key

ghostdevv:
> Sorry that sentence was unclear - your implementation is _not_ doing what it's supposed to. Instead of creating an array per entry it just takes the first occurrence of a match and assigns it to the key

omg yea :woman_facepalming:

dummdidumm (approved):
thank you!

PR #3044, opened 2026-06-03
fix: mark optional members with a trailing ? in completion labels (#3043)

### LINKED ISSUE #3043: Optionally is missing from completions
### Describe the bug

When an option is optional, the completion label is missing the `?`

<table>
<tr>
 <td>Svelte Language Server</td>
 <td>vtsls</td>
<tr>
 <td>Image</td>
 <td>Image</td>
</table>

### Reproduction

```js
function foo(options: { bar: boolean; baz?: number }) {}
foo({ bar: false,   })
//                       ?^ See completions here 
```

### Expected behaviour

The label returned by the language server has a `?` when it's optional

### System Info

n/a

svelte-language-server

_No response_

MohammadYusif:
## Summary

- Optional object/class members were missing the trailing `?` in their completion labels (e.g. `baz?: number` completed as `baz`), unlike the TypeScript language service / vtsls.
- TypeScript reports optionality via the `optional` kind modifier on the completion entry, which `getCompletionLabelAndInsert` ignored.

## Changes

- `packages/language-server/src/plugins/typescript/features/CompletionProvider.ts`: detect `ts.ScriptElementKindModifier.optionalModifier` and append `?` to the label, keeping `insertText` as the bare name so the `?` is display-only.
- `packages/language-server/test/plugins/typescript/testfiles/completions/object-member.svelte`: add an optional-property scenario.
- `packages/language-server/test/plugins/typescript/features/CompletionProvider.test.ts`: regression test asserting the `baz?` label and bare insert text.
- `.changeset/optional-completion-label.md`: patch changeset for `svelte-language-server`.

## Test plan

`pnpm --filter svelte-language-server test` — 69/69 pass; new test fails without the source change.

Fixes #3043

### DIFF
--- a/packages/language-server/src/plugins/typescript/features/CompletionProvider.ts
+++ b/packages/language-server/src/plugins/typescript/features/CompletionProvider.ts
@@ -765,17 +765,13 @@ export class CompletionsProviderImpl implements CompletionsProvider<CompletionRe
             };
         }
 
-        // Optional members are reported by TypeScript through the `optional` kind
-        // modifier. Surface that in the label with a trailing `?` (matching the
-        // TypeScript language service), while keeping the inserted text as the
-        // bare name so the `?` is display-only.
         const isOptional = kindModifiers
             ?.split(/,|\s+/)
             .includes(ts.ScriptElementKindModifier.optionalModifier);
 
         return {
             label: isOptional ? `${name}?` : name,
-            insertText: isOptional ? (insertText ?? name) : insertText,
+            insertText,
             isSvelteComp,
             isRunesCompletion
         };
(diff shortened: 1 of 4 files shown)


packages/language-server/src/plugins/typescript/features/CompletionProvider.ts
```diff
@@ -765,9 +765,17 @@ export class CompletionsProviderImpl implements CompletionsProvider<CompletionRe
             };
         }
 
+        // Optional members are reported by TypeScript through the `optional` kind
+        // modifier. Surface that in the label with a trailing `?` (matching the
+        // TypeScript language service), while keeping the inserted text as the
+        // bare name so the `?` is display-only.
+        const isOptional = kindModifiers
+            ?.split(/,|\s+/)
+            .includes(ts.ScriptElementKindModifier.optionalModifier);
```
dummdidumm:
```suggestion
            insertText: isOptional ? (insertText ?? name) : insertText, // else label with ? is used if no insertText available
```

dummdidumm (approved):
thank you

Issue #3055, opened 2026-06-11
`svelte-check` and `load-config` don't support custom Vite config filenames


teemingc:
### Describe the bug

It's possible for users to specify a custom filename for their Vite config and then specify it via `vite dev -c filename` as shown by our SvelteKit test app.
The issue is that there's currently no way to specify the custom name for `svelte-check` and other tooling

### Reproduction

Running `svelte-check` on our SvelteKit test app without a `svelte.config.js` file
https://github.com/sveltejs/kit/blob/4298357331f4a35cf06e08b34bcb60f57b38235c/packages/kit/test/apps/options/vite.custom.config.js

### Expected behaviour

`svelte-check` should take a `-c vite.custom.config.js` option perhaps

### System Info

- OS: [e.g. Windows]
- IDE: [e.g. VSCode, Atom]

### Which package is the issue about?

svelte-check

PR #3062, opened 2026-06-18
fix: exclude vendor/ directory from config crawl

### LINKED ISSUE #3054: svelte-check: Config loader crawls vendor/ directory and logs non-fatal errors for unrelated vite.config.js files
## Description

When running `svelte-check` with `--tsconfig`, the config discovery mechanism recursively crawls from `dirname(tsconfigPath)` looking for `vite.config.*` and `svelte.config.*` files. The exclude filter in `ConfigLoader.loadConfigs()` only blocks `node_modules/` and dotfiles — it does **not** exclude `vendor/` directories.

In PHP/Laravel projects, the `vendor/` directory can contain `vite.config.js` files from packages (e.g., `vendor/laravel/wayfinder/vite.config.js`, `vendor/laravel/framework/.../vite.config.js`). These are unrelated to the project Svelte configuration but are still loaded and produce errors:

```
Error while loading config at /path/to/vendor/laravel/wayfinder/vite.config.js
Error: No Svelte configuration found in vite config. Is @sveltejs/vite-plugin-svelte configured?
    at ConfigLoader.loadConfig (svelte-check/dist/src/index.js:40183:13)

Error while loading config at /path/to/vendor/laravel/framework/src/Illuminate/Foundation/resources/exceptions/renderer/vite.config.js
Error: No Svelte configuration found in vite config. Is @sveltejs/vite-plugin-svelte configured?
    at ConfigLoader.loadConfig (svelte-check/dist/src/index.js:40183:13)
```

These errors are **non-fatal** and do not affect the check result (`0 errors, 0 warnings`), but they pollute stderr output in CI/CD pipelines.

## Impact

This issue is not limited to non-fatal stderr noise. As reported in [#issuecomment-4742937222](https://github.com/sveltejs/language-tools/issues/3054#issuecomment-4742937222), vendor vite configs can cause **fatal preprocessing failures** that affect every `.svelte` file:

```
vendor/laravel/telescope/vite.config.js (1:489) [UNRESOLVED_IMPORT] Could not resolve '@vitejs/plugin-vue2'
Preprocessing failed
svelte-check found 44 errors and 0 warnings in 44 files
```

This happens because the vendor vite config is loaded and used for preprocessing, even though it has no relation to the project's Svelte setup.

## Root Cause

In `ConfigLoader.loadConfigs()` (line ~40046), the glob crawl exclude filter:

```js
.exclude((_, path) => {
    return path.includes("node_modules/") || path.includes("/.") || path[0] === ".";
})
```

does not exclude `vendor/`. Additionally, the `--workspace` flag only controls the **file checking** scope — config loading always uses `dirname(tsconfigPath)` as its starting point (line 111608):

```js
return this.getTSServiceByConfigPath(
    this.options.tsconfigPath,
    dirname(this.options.tsconfigPath)  // workspace for config loading
);
```

## Reproduction

1. Create a Laravel project with Inertia.js + Svelte
2. Install packages that ship `vite.config.js` in vendor (e.g., `laravel/wayfinder`)
3. Run `svelte-check --tsconfig ./tsconfig.json`
4. Observe non-fatal errors about vendor vite configs on stderr

## Suggested Fix

Add `vendor/` to the exclude filter in `loadConfigs()`:

```js
.exclude((_, path) => {
    return path.includes("node_modules/") || path.includes("/vendor/") || path.includes("/.") || path[0] === ".";
})
```

Or provide a CLI option like `--ignore-paths` that works with `--tsconfig` (currently `--ignore` only works with `--no-tsconfig`).

ajangsupardi:
## Description

The config loader recursively searches for  and  files, but the exclude filter only blocked  and dotfiles. In PHP/Laravel projects, the  directory can contain  files from packages that are unrelated to the project's Svelte setup, causing either non-fatal stderr noise or fatal preprocessing failures.

This PR adds  to the exclude filter in both  and  to prevent crawling vendor directories.

## Changes

- : Add  to exclude filter in 
- : Add  to exclude filter in 

## Testing

Verified that vendor directories are now excluded from config crawling.

Fixes #3054

### DIFF
--- a/packages/language-server/test/lib/documents/configLoader.test.ts
+++ b/packages/language-server/test/lib/documents/configLoader.test.ts
@@ -409,7 +409,7 @@ describe('ConfigLoader', () => {
         );
     });
 
-    it('should exclude vendor directory from config crawl', async () => {
+    it('should exclude vendor directory when configured via setExcludeDirs', async () => {
         const configLoader = createConfigLoader(
             mockFdirWithExclude(['vendor/laravel/svelte.config.js', 'src/svelte.config.js']),
             {
@@ -421,6 +421,7 @@ describe('ConfigLoader', () => {
             (module: URL) => Promise.resolve({ default: { preprocess: module.toString() } }),
             process.features
         );
+        configLoader.setExcludeDirs(['vendor']);
         await configLoader.loadConfigs(normalizePath('/some/path'));
 
         // vendor config should NOT be found
--- a/packages/language-server/test/lib/documents/configLoader.test.ts
+++ b/packages/language-server/test/lib/documents/configLoader.test.ts
@@ -31,34 +31,29 @@ describe('ConfigLoader', () => {
         return path.join(...filePath.split('/'));
     }
 
-    function mockFdir(results: string[] | (() => string[])): any {
-        return class {
-            withPathSeparator() {
-                return this;
-            }
-            exclude() {
-                return this;
-            }
-            filter() {
-                return this;
-            }
-            withRelativePaths() {
-                return this;
-            }
-            crawl() {
-                return this;
-            }
-            sync() {
-                return typeof results === 'function' ? results() : results;
-            }
-        };
-    }
-
-    /**
-     * Like mockFdir, but actually applies the exclude callback so we can
-     * verify that node_modules, vendor, and dotfile exclusion works.
-     */
-    function mockFdirWithExclude(rawFiles: string[]): any {
+    function mockFdir(results: string[] | (() => string[]), applyExclude = false): any {
+        if (!applyExclude) {
+            return class {
+                withPathSeparator() {
+                    return this;
+                }
+                exclude() {
+                    return this;
+                }
+                filter() {
+                    return this;
+                }
+                withRelativePaths() {
+                    return this;
+                }
+                crawl() {
+                    return this;
+                }
+                sync() {
+                    return typeof results === 'function' ? results() : results;
+                }
+            };
+        }
         let excludeFn: (path: string) => boolean;
         return class {
             withPathSeparator() {
@@ -78,7 +73,8 @@ describe('ConfigLoader', () => {
                 return this;
             }
             sync() {
-                return rawFiles.filter((f) => !excludeFn(f));
+                const raw = typeof results === 'function' ? results() : results;
+                return raw.filter((f: string) => !excludeFn(f));
             }
         };
     }
@@ -411,7 +407,7 @@ describe('ConfigLoader', () => {
 
     it('should exclude vendor directory when configured via setExcludeDirs', async () => {
         const configLoader = createConfigLoader(
-            mockFdirWithExclude(['vendor/laravel/svelte.config.js', 'src/svelte.config.js']),
+            mockFdir(['vendor/laravel/svelte.config.js', 'src/svelte.config.js'], true),
             {
                 existsSync: (p) =>
                     typeof p === 'string' &&
@@ -439,7 +435,7 @@ describe('ConfigLoader', () => {
 
     it('should exclude node_modules directory from config crawl', async () => {
         const configLoader = createConfigLoader(
-            mockFdirWithExclude(['node_modules/some-pkg/svelte.config.js', 'src/svelte.config.js']),
+            mockFdir(['node_modules/some-pkg/svelte.config.js', 'src/svelte.config.js'], true),
             {
                 existsSync: (p) =>
                     typeof p === 'string' &&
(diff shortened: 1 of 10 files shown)


dummdidumm:
It's not guaranteed that this folder is not used with a config inside as this is not a "reserved" directory name in JavaScript projects. In other words we can't do it like this. 

Is there a definite way to know that this is a laravel project (e.g. some config file that is always present) and only then apply that filter?

If not then I would be ok with having an option to tell it which directories to ignore.

ajangsupardi:
Got it. I'll add a configuration option to let users specify directories to ignore.

Quick question: where should this config live? I'm thinking either in `svelte.config.js` or as a language server setting. Which approach fits better with the existing config patterns?

dummdidumm:
Is there a definite way to know that this is a laravel project (e.g. some config file that is always present) and only then apply that filter?

ajangsupardi:
Hi @dummdidumm, thanks for the feedback!

There's no reliable universal way to detect Laravel projects — `artisan` isn't guaranteed to exist, and `vendor/` with PHP class files also appears in other PHP frameworks (Symfony, CakePHP, etc.). Auto-detection would be fragile.

Instead, I've implemented a config option approach (`svelte.excludeDirs`) that works for **all** frameworks. Users can specify which directories to exclude from config crawling via their language server settings.

I also added `--exclude-dirs` support in `svelte-check` CLI for consistency.

The implementation is ready on the branch — happy to adjust if you have a different approach in mind.

dummdidumm:
you did not. please turn on your brain and don't just let your LLM make all the decisions

ajangsupardi:
You're right, that was careless of me. The config option is now actually implemented — see the latest commit. `svelte.excludeDirs` takes an array of directory names to exclude from config crawling. Works in both the language server settings and `svelte-check --exclude-dirs`.

packages/svelte-check/src/index.ts
```diff
@@ -580,6 +593,8 @@ parseOptions(async (opts) => {
     try {
         const writer = instantiateWriter(opts);
 
+        configLoader.setExcludeDirs(opts.excludeDirs);
```
dummdidumm: instead of exposing configLoader I'd rather have the SvelteCheck export have an option to set this.

packages/svelte-check/src/options.ts
```diff
@@ -58,6 +59,10 @@ export function parseOptions(cb: (opts: SvelteCheckCliOptions) => any) {
             '--ignore',
             'Only has an effect when using `--no-tsconfig` option. Files/folders to ignore - relative to workspace root, comma-separated, inside quotes. Example: `--ignore "dist,build"`'
         )
+        .option(
+            '--exclude-dirs',
```
dummdidumm: we already have `--ignore` - I'd rather repurpose that for this. In other words, ignore also does what `--exclude-dirs` does, and we adjust the wording of the `--ignore` option accordingly:

'Files/folders to ignore - relative to workspace root, comma-separated, inside quotes. Example: `--ignore "dist,build"`'. Use this to exclude directories from config crawling. When used together with the `--no-tsconfig` option, this also has an effect on the Svelte files that are checked.

packages/language-server/test/lib/documents/configLoader.test.ts
```diff
@@ -54,6 +54,35 @@ describe('ConfigLoader', () => {
         };
     }
 
+    /**
+     * Like mockFdir, but actually applies the exclude callback so we can
+     * verify that node_modules, vendor, and dotfile exclusion works.
+     */
+    function mockFdirWithExclude(rawFiles: string[]): any {
```
dummdidumm: no need to have this separate, just incorporate the new functionality into the existing mockFdir, and the default is to not exclude anything.

dummdidumm (changes requested):
thanks, left a few comments.
If this becomes a new VS Code option you also need to adjust packages/svelte-vscode/package.json and add that option there

dummdidumm:
You know what - implementing #3055 might actually be the better solution here. I assume you don't have multiple configs to load from a single svelte-check run. I'll implement it.

ajangsupardi:
Addressed all three review comments:

- `excludeDirs` is now a `SvelteCheck` option instead of exposing `configLoader`
- Removed `--exclude-dirs`, repurposed `--ignore` to also exclude dirs from config crawling
- Merged `mockFdirWithExclude` into `mockFdir` with an `applyExclude` param

ajangsupardi:
Closing in favor of #3066 which solves this by allowing explicit Svelte/Vite config file paths, turning off file crawling entirely.

PR #3067, opened 2026-06-23
fix (svelte-check): `--tsgo-experimental-api` option is not honored

### LINKED ISSUE #3068: --tsgo-experimental-api does not actually enable the new mode
### Describe the bug

The option is `--tsgo-experimental-api` but the code that checks the option looks for `--tsgo-experimental` so you can't actually enable it.

I already filed https://github.com/sveltejs/language-tools/pull/3067 with the fix, just opening this for completeness.

### Reproduction

Run `svelte-check --tsgo-experimental-api` and note that it takes the same amount of time, and other checks like running with `--incremental` do not throw the expected error.

### Expected behaviour

Should enable it

### System Info

N/A

_No response_

_No response_

dimfeld:
The option is `--tsgo-experimental-api` but when using it, the code was looking at `opts['tsgo-experimental']` so it wasn't possible to actually enable the new flag in the published 4.7.0.

Manually verified the fix by building svelte-check and running it against my projects.

Fixes #3068

### DIFF
--- a/.changeset/cold-ads-double.md
+++ b/.changeset/cold-ads-double.md
@@ -2,4 +2,4 @@
 'svelte-check': patch
 ---
 
-Fix command-liee option parsing of --tsgo-experimental-api flag
+fix: correctly parse --tsgo-experimental-api flag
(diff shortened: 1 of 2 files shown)


jasonlyu123 (approved):
Thanks. I renamed the flag and missed an important one😅.

PR #3076, opened 2026-07-08
feat: zero-config `+error.svelte` props

Rich-Harris:
No idea if I'm doing this right, but: this should allow zero-config types for the `error` prop passed to an `+error.svelte` file:

```svelte
<script>
  let { error } = $props(); // error is typed as `App.Error`
</script>

<h1>oops: {error.message}</h1>
```

I thought we'd need to do https://github.com/sveltejs/kit/pull/16272 first, but I'm not sure it's all that helpful since we'd need to be careful about the version. As it stands, this would allow someone to create the above props declaration with a version of SvelteKit that _didn't_ pass `error` to the component, but that's probably fine?

### DIFF
--- a/.changeset/grumpy-rabbits-chew.md
+++ b/.changeset/grumpy-rabbits-chew.md
@@ -1,5 +1,6 @@
 ---
-'svelte2tsx': minor
+'svelte2tsx': patch
+'svelte-check': patch
 ---
 
 feat: zero-config `+error.svelte` props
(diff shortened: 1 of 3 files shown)


.changeset/grumpy-rabbits-chew.md
```diff
@@ -0,0 +1,5 @@
+---
+'svelte2tsx': minor
```
dummdidumm: We're on 0.x versions so everything except breaking changes are patches. And since this also affects svelte-check (because it bundles svelte2tsx) it needs a changeset, too.

```suggestion
'svelte2tsx': patch
'svelte-check': patch
```

dummdidumm (approved):
👍

PR #3078, opened 2026-07-09
fix: don't emit invalid `export` on type/interface declarations in the instance script

### LINKED ISSUE #2832: "Modifiers cannot appear here" in unrelated `export type` when props don't have explicit type
### Describe the bug

When not declaring an explicit type for props, it seems `type` declarations cannot be `export`ed.

### Reproduction

This works:

```svelte
<script lang="ts">
	export type ExampleType = 'A' | 'B' | 'C'

	export interface ExampleProps {}

	const {}: ExampleProps = $props()
</script>
```

But when I remove the `: ExampleProps` from `const {} = $props()`, or when I remove the props line altogether (with or without the accompanying interface), I get this error:

```
Error: Modifiers cannot appear here.
If this is a declare statement, move it into <script context="module">..</script> (ts)
<script lang="ts">
        export type ExampleType = 'A' | 'B' | 'C'
</script>
```

(side-note: maybe the error can be updated to show `<script module>`, as the `context` tag is deprectated)

The `export` keyword is highlighted. `export function` gives no error, which is surprising because the effect of the function may differ between different instances of the component. Thankfully, even if `export`ed,  functions cannot be imported from other components, but maybe those *should* give an error.

Possibly related, this does not work regardless of explicitly typed props:

```svelte
<script lang="ts">
	const allExamples = ['A', 'B', 'C'] as const;
	export type ExampleType = typeof allExamples[number];
</script>
```

It gives the same error, but it's valid (though a bit obscure) typescript.

As a workaround, everything works properly in module context (by having separate `<script>` and `<script module>` blocks). But since we're also allowed to do `export interface Props` in normal context, I think it would be more intuitive if anything type-related could be exported in normal context as well (and possibly `static function`s, `readonly const`s, etc.), as types are inherently equal across all instances of the component.

### Logs

```shell

```

### System Info

```shell
System:
    OS: Linux 6.14 Ubuntu 24.04.3 LTS 24.04.3 LTS (Noble Numbat)
    CPU: (24) x64 AMD Ryzen 9 7900X 12-Core Processor
    Memory: 23.70 GB / 30.50 GB
    Container: Yes
    Shell: 5.2.21 - /bin/bash
  Binaries:
    Node: 22.18.0 - ~/.nvm/versions/node/v22.18.0/bin/node
    Yarn: 1.22.22 - ~/.nvm/versions/node/v22.18.0/bin/yarn
    npm: 10.9.3 - ~/.nvm/versions/node/v22.18.0/bin/npm
    pnpm: 10.13.1 - ~/.nvm/versions/node/v22.18.0/bin/pnpm
  Browsers:
    Chrome: 139.0.7258.127
  npmPackages:
    svelte: ^5.0.0 => 5.38.0
```

### Severity

annoyance

spokodev:
Fixes #2832

`export type`/`export interface` in the instance script keep their `export` keyword inside the generated render function when they aren't hoisted to the module scope, e.g. when `$props()` has no explicit type or there are no props at all. `export` isn't valid on a declaration inside a function body, so TypeScript reports "Modifiers cannot appear here." on an otherwise valid component.

Repro:

```svelte
<script lang="ts">
	export type ExampleType = 'A' | 'B' | 'C';
</script>
```

Adding a typed `const {}: Props = $props()` makes the error disappear, because the hoisting path then moves the type out to the module scope — which is why this is easy to miss.

Fix: strip the `export` keyword from top-level type/interface declarations that stay inside the render function (the ones that don't get hoisted), mirroring the existing handling for exported functions and classes in `handleExportFunctionOrClass`. Hoisted declarations are left untouched so they remain importable from other components.

Adds a regression test.

### DIFF
--- a/.changeset/tidy-experts-cheer.md
+++ b/.changeset/tidy-experts-cheer.md
@@ -1,0 +1,5 @@
+---
+'svelte2tsx': patch
+---
+
+fix: don't emit invalid `export` on type/interface declarations in the instance script
--- a/packages/svelte2tsx/src/svelte2tsx/nodes/ExportedNames.ts
+++ b/packages/svelte2tsx/src/svelte2tsx/nodes/ExportedNames.ts
@@ -54,6 +54,8 @@ export class ExportedNames {
     >();
     private doneDeclarationTransformation = new Set<ts.VariableDeclarationList>();
     private getters = new Set<string>();
+    private exportedTypesAndInterfaces: Array<ts.TypeAliasDeclaration | ts.InterfaceDeclaration> =
+        [];
 
     constructor(
         private str: MagicString,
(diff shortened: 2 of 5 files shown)


jasonlyu123:
I think removing the modifier will only make it even more confusing. You can't import it, and there is no error message indicating a problem.

PR #3091, opened 2026-08-04
fix: don't log an error for vite configs without a Svelte plugin

### LINKED ISSUE #3080: Reading svelte config from vite.config.js/ts files introduced in 4.6.0 seems flawed
### Describe the bug

With reference to these issues/PRs:
* https://github.com/sveltejs/language-tools/pull/3031  (initial change introduced in 4.6.0)
* https://github.com/sveltejs/language-tools/issues/3054 (bug report related to the change)
* https://github.com/sveltejs/language-tools/pull/3066  (fix introduced in 4.7.0)

It would seem that the short term "fix" was to allow a config file path to be explicitly provided to turn off recursive crawling, and a longer term "TODO" for the next major was mentioned:
> We should not recursively crawl for config files anymore if we find one in the root directory or next to the tsconfig.

I'm not sure either the short term fix or long term TODO fixes the following scenario:
(if it does, please let me know how)

### Reproduction

Assume the following monorepo style layout:
```
packages/
  package-a/    // this package does not use svelte or @sveltejs/vite-plugin-svelte
    vite.config.ts
  package-b/    // this package does use svelte and @sveltejs/vite-plugin-svelte
    vite.config.ts
  tsconfig.json    // shared tsconfig.json file used by all packages
```

With `svelte-check@4.5.0`, running `svelte-check` from the project root runs successfully:
```
Loading svelte-check in workspace: [REDACTED]
Getting Svelte diagnostics...

svelte-check found 0 errors and 0 warnings
```

...and just to be sure that it is actually checking, if I deliberately introduce an issue it correctly errors:

```
Loading svelte-check in workspace: [REDACTED]
Getting Svelte diagnostics...

[REDACTED]/packages/package-b/components/Foo.svelte:5:6
Error: 'foo' is declared but its value is never read. (ts)

        let foo = "";
</script>
====================================
svelte-check found 1 error and 0 warnings in 1 file
```

With `svelte-check@4.6.0`, running `svelte-check` from the project root now reports:

```
Error while loading config at  [REDACTED]/packages/package-a/vite.config.ts
Error: No Svelte configuration found in vite config. Is @sveltejs/vite-plugin-svelte configured?
    at ConfigLoader.loadConfig ([REDACTED]/node_modules/svelte-check/dist/src/index.js:40183:13)
Loading svelte-check in workspace: [REDACTED]
Getting Svelte diagnostics...

svelte-check found 0 errors and 0 warnings
```
Note that the error is complaining about `package-a`, which does not use svelte.

With `svelte-check@4.7.0`, running `svelte-check --config packages/package-b/vite.config.ts` from the project root suppresses the error.

However that only works because there is only one package in the repo that uses svelte.
How would this work if there was also a third package that used svelte, e.g.
```
packages/
  ...
  package-c/    // also uses svelte and @sveltejs/vite-plugin-svelte
    vite.config.ts
```

Would I need to run `svelte-check` separately for each package, passing the correct path to the `vite.config.ts` for that package?, e.g.
```
svelte-check --config packages/package-b/vite.config.ts && svelte-check --config packages/package-c/vite.config.ts
```

If we assume the long term TODO happens, note that my `vite.config.ts` files are neither in the root, nor are they adjacent to a `tsconfig` file, so it's not clear whether that would work here either.

### Expected behaviour

In `4.5.0`, running `svelte-check` at the root of a monorepo where individual packages may or may not use svelte would check successfully.

In `4.6.0` and later, that now errors unless a path to _one of the package's_ `vite.config.ts` is specified, but it is not clear how this works when there are >1 svelte packages mixed with >1 non-svelte packages, each with their own `vite.config.ts` and a shared `tsconfig` in a parent directory.

### System Info

- OS: [e.g. Windows] macOS
- IDE: [e.g. VSCode, Atom] VSCode

svelte-check

_No response_

Socialpranker:
Partially addresses #3080.

## What this fixes

Since 4.6.0 the config loader crawls for `vite.config.*` files. In a monorepo that means it also visits packages which don't use Svelte at all, and for each of those it logs an error:

```
Error while loading config at  .../packages/package-a/vite.config.ts
Error: No Svelte configuration found in vite config. Is @sveltejs/vite-plugin-svelte configured?
```

`package-a` not using Svelte isn't a problem, so reporting it as an error is misleading — the reporter in #3080 hit exactly this.

`loadConfig` already distinguishes the two cases, it just wasn't being used to: a config that genuinely fails to load comes back as `{ error, ... }`, while a config that loads fine but has no Svelte plugin falls through and comes back as `undefined`. Only the first is an error. This checks `result?.error` and keeps `Logger.error` for that case, logging the other at the normal level.

Everything else is deliberately untouched. The fallback preprocessor and `loadConfigError` are still set exactly as before, so the retry logic in `loadConfigs` (`!config || config.loadConfigError`), `addFallbackConfig`, and the diagnostics path in `getDiagnostics.ts` all behave identically. This is only about the log level.

## What this does not fix

#3080 raises a second, larger question: how `svelte-check` should work in a monorepo with **several** Svelte packages, each with its own `vite.config.ts`, when `--config` takes a single path. That's a design question about `--config`/`ExplicitConfigScope` rather than a bug, so I left it alone — happy to follow up if you decide on a direction.

## Verification

Two tests added, both driven from the actual behaviour rather than written to match the patch:

- **no svelte plugin → no error logged.** Fails on `main` (`Logger.error` is called with the "No Svelte configuration found in vite config" message), passes with the fix.
- **config genuinely fails to load → error still logged.** Passes both before and after, i.e. it pins the behaviour the fix must not break.

The second test builds `ConfigLoader` directly rather than through the `createConfigLoader` helper, because that helper always wraps its result in `{ config }` and so can't express a failed load. I found that out by tracing what the mock actually returned — my first attempt at the test passed for the wrong reason.

Checks run locally on Node 22:

- `configLoader.test.ts`: 15 passing.
- Full `svelte-language-server` suite: 819 passing, 21 pending, 3 failing — all three in `SvelteCheckTSGoDiagnosticsProvider`. Those also fail on a clean `main` (5 failing there, the set moves between runs), so they're pre-existing and unrelated to this change.
- `prettier --check` clean on the touched files, `tsc --noEmit` exits 0.

Changeset included.

Written with Claude Code (Claude Opus 5); I reviewed and tested everything before submitting.

### DIFF
--- a/.changeset/lucky-donkeys-repeat.md
+++ b/.changeset/lucky-donkeys-repeat.md
@@ -1,0 +1,6 @@
+---
+'svelte-language-server': patch
+'svelte-check': patch
+---
+
+fix: don't log an error for vite configs without a Svelte plugin
--- a/packages/language-server/src/lib/documents/configLoader.ts
+++ b/packages/language-server/src/lib/documents/configLoader.ts
@@ -286,6 +286,10 @@ export class ConfigLoader {
         }
 
         const configSource = result?.configSource ?? getConfigSource(configPath);
+        // A vite config that loads fine but has no Svelte plugin is not an error: in a monorepo
+        // the crawler also visits packages that don't use Svelte at all. Only report loading
+        // failures, which are the cases where `loadConfig` hands back an `error`.
+        const loadFailed = result?.error !== undefined;
         const error =
             result?.error ??
             new Error(
@@ -294,8 +298,12 @@ export class ConfigLoader {
                     : 'No Svelte configuration found'
             );
         const errorConfigPath = result?.configFilePath ?? configPath;
-        Logger.error('Error while loading config at ', errorConfigPath);
-        Logger.error(error);
+        if (loadFailed) {
+            Logger.error('Error while loading config at ', errorConfigPath);
+            Logger.error(error);
+        } else {
+            Logger.log('No Svelte config found at ', errorConfigPath);
+        }
 
         return {
             config: {
--- a/packages/language-server/test/lib/documents/configLoader.test.ts
+++ b/packages/language-server/test/lib/documents/configLoader.test.ts
@@ -1,4 +1,5 @@
 import { ConfigLoader } from '../../../src/lib/documents/configLoader';
+import { Logger } from '../../../src/logger';
 import path from 'path';
 import { pathToFileURL, URL } from 'url';
 import assert from 'assert';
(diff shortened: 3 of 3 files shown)


dummdidumm (approved):
thank you!

PR #3093, opened 2026-08-05
perf(svelte2tsx): use a module resolution cache in emitDts

### LINKED ISSUE #3094: emitDts resolves modules without a ModuleResolutionCache, dominating svelte-package build time
### Describe the bug

`emitDts` (used by `@sveltejs/package` to generate declaration files) resolves every import through its compiler host with bare `ts.resolveModuleName` calls and no `ts.createModuleResolutionCache`. Nothing is cached across the program's files, so each of the typically hundreds of node_modules declaration files in the program re-runs the full node_modules resolution walk (open/stat/realpath syscalls) for every import statement.

In a pnpm + Turborepo monorepo with 29 Svelte 5 component packages (TypeScript 6.0.3, svelte2tsx 0.7.57, ~880 node_modules declaration files per program), V8 profiles showed ~82% of each `svelte-package` run inside `emitDts`, and ~52% of total build CPU inside uncached module resolution. The Svelte compile itself is under 1%.

For comparison, the language server in this repo already maintains a deliberate module resolution cache (`module-loader.ts`; see #2754, #2902) - `emitDts` appears to be the remaining resolution path without one.

### Reproduction

Run `svelte-package` on any package with a realistic dependency tree and profile it (`node --cpu-prof node_modules/.bin/svelte-package -i ./src`); the profile is dominated by `resolveModuleName` -> fs syscalls under `emitDts`.

### Expected behaviour

Resolution results should be cached per program, as TypeScript's `ts.createModuleResolutionCache` provides.

### Impact of fixing

Passing a `ModuleResolutionCache` per resolution host (measured via a pnpm patch carrying the exact diff in the linked PR): the 29-package cold build dropped from 173.0s to 113.2s of summed build time (21.0s to 13.3s wall on 14 cores), with byte-identical emitted output verified by recursive diff across all packages.

### Proposed fix

PR: https://github.com/sveltejs/language-tools/pull/3093

### Package

svelte2tsx (via @sveltejs/package)

### System Info

- svelte2tsx 0.7.57 via @sveltejs/package 2.5.8
- TypeScript 6.0.3, Svelte 5.56.8
- Node 22, macOS arm64 (profiling) and Linux CI (production use of the patch)

sam-easygo:
### perf(svelte2tsx): use a module resolution cache in emitDts

Fixes #3094.

#### Summary

`emitDts` resolves every import through its compiler host with bare `ts.resolveModuleName` calls, without a `ts.createModuleResolutionCache`. Nothing is cached across the program's files, so each of the (typically hundreds of) node_modules declaration files re-runs the full node_modules resolution walk (open/stat/realpath syscalls) for every import statement.

This PR creates a `ModuleResolutionCache` per resolution host in `createTsCompilerHost` and passes it to the two existing `ts.resolveModuleName` calls. Each of the two hosts (plain `ts.sys`, and the svelte-aware system) gets its own cache because they disagree about whether virtual `.svelte.ts`/`.svelte.js` paths exist, so their results must not be mixed.

This brings `emitDts` in line with the rest of the repo: the language server already maintains a deliberate module resolution cache in its `module-loader.ts` (see #2754, #2902); `emitDts` was the remaining resolution path without one. It is also complementary to #2963, which caches script-AST parsing on the language-server path; this PR addresses module resolution on the `emitDts`/`svelte-package` build path.

#### Benchmarks

Profiled in a pnpm + Turborepo design-system monorepo with 29 Svelte 5 component packages, each built with `svelte-package` (TypeScript 6.0.3, svelte2tsx 0.7.57, ~880 node_modules declaration files per program, macOS/arm64). V8 profiles showed ~82% of each `svelte-package` run inside `emitDts`, and ~52% of total build CPU in uncached module resolution.

With this change applied via a pnpm patch (identical diff):

| Cold build of the 29 packages | Before | After |
| --- | --- | --- |
| Sum of per-package build times | 173.0s | 113.2s |
| Process CPU (user+sys) | 133s | 99s |
| Wall time (14 cores) | 21.0s | 13.3s |

A representative single package dropped from ~10.9s to ~6.7s under parallel load, and to ~2s when built alone. The monorepo has been running this exact diff in CI since.

#### Correctness

- Within one `emitDts` call file contents do not change, so caching resolutions is sound.
- The svelteMap population is unaffected: the first resolution of each `(directory, specifier)` pair still goes through `svelteSys.fileExists`, which registers `.svelte` files; cache hits only replay results whose first resolution already did that work.
- Output verified byte-for-byte identical: recursive diff of every emitted `dist/` tree (d.ts + d.ts.map) across all 29 packages, patched vs unpatched, from both clean and warm starting states.

#### Test plan

- `pnpm test` in `packages/svelte2tsx`: 370 passing, including the `emitDts` suite (no new tests: the change is a pure caching pass-through with no observable behaviour, and byte-identical output is verified above).
- `pnpm lint` (prettier check): clean.
- `pnpm format`: no changes needed.
- Changeset included (`svelte2tsx`: patch).

#### A note on authorship

I use AI tooling in my workflow and it helped discover and author this change. If AI-assisted contributions are against the maintainers' stance, I am entirely happy for this to be closed.

### DIFF
--- a/packages/svelte2tsx/src/emitDts.ts
+++ b/packages/svelte2tsx/src/emitDts.ts
@@ -278,6 +278,20 @@ async function createTsCompilerHost(options: any, svelteMap: SvelteMap, absDecla
         });
     };
 
+    const getCanonicalFileName = createGetCanonicalFileName(ts.sys.useCaseSensitiveFileNames);
+    // Each resolution host gets its own cache: the two hosts disagree about
+    // whether virtual .svelte.ts/.svelte.js paths exist.
+    const tsModuleResolutionCache = ts.createModuleResolutionCache(
+        ts.sys.getCurrentDirectory(),
+        getCanonicalFileName,
+        options
+    );
+    const svelteModuleResolutionCache = ts.createModuleResolutionCache(
+        ts.sys.getCurrentDirectory(),
+        getCanonicalFileName,
+        options
+    );
+
     function resolveModuleName(name: string, containingFile: string, compilerOptions: any) {
         // Delegate to the TS resolver first.
         // If that does not bring up anything, try the Svelte Module loader
@@ -286,14 +300,20 @@ async function createTsCompilerHost(options: any, svelteMap: SvelteMap, absDecla
             name,
             containingFile,
             compilerOptions,
-            ts.sys
+            ts.sys,
+            tsModuleResolutionCache
         ).resolvedModule;
         if (tsResolvedModule && !isVirtualSvelteFilepath(tsResolvedModule.resolvedFileName)) {
             return tsResolvedModule;
         }
 
-        return ts.resolveModuleName(name, containingFile, compilerOptions, svelteSys)
-            .resolvedModule;
+        return ts.resolveModuleName(
+            name,
+            containingFile,
+            compilerOptions,
+            svelteSys,
+            svelteModuleResolutionCache
+        ).resolvedModule;
     }
 
     return host;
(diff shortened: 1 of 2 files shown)


packages/svelte2tsx/src/emitDts.ts
```diff
@@ -278,6 +278,22 @@ async function createTsCompilerHost(options: any, svelteMap: SvelteMap, absDecla
         });
     };
 
+    const getCanonicalFileName = ts.sys.useCaseSensitiveFileNames
```
jasonlyu123: We use a slightly different "to lower case" logic in the language server. Can you copy the function here and use that instead?
https://github.com/sveltejs/language-tools/blob/61ae3d8cddc47d78b5b220662ede110318da9081/packages/language-server/src/utils.ts#L351

sam-easygo: Thanks @jasonlyu123, good call. Let me know if that looks as expected. Copied verbatim including comments referencing where its sourced from. Also took the `createGetCanonicalFileName` wrapper too.

jasonlyu123 (approved):

jasonlyu123:
Weird. The head of the branch seems a bit strange after I updated the branch. Now "Squash and merge" preview show commits in the main branch as changes in this PR. Can you try to rebase it again?

sam-easygo:
> Weird. The head of the branch seems a bit strange after I updated the branch. Now "Squash and merge" preview show commits in the main branch as changes in this PR. Can you try to rebase it again?

@jasonlyu123 strange, no longer had the button to update the branch just on the github UI but rebased locally on latest and force pushed but will need the workflow approval again

jasonlyu123:
Oh, it's from your commit message. Can you clean that up?

sam-easygo:
> Oh, it's from your commit message. Can you clean that up?

@jasonlyu123 Oh sorry about that not even sure how that happened, cleaned up now

PR #3097, opened 2026-08-06
fix: don't double-nest kit options when loading from a vite config

Nic-Polumeyv:
`api.options` on the `vite-plugin-sveltekit-setup` plugin has been the already-split config (`{ compilerOptions, preprocess, extensions, kit }`) since sveltejs/kit#15944, so the flat destructure here picked the existing `kit` key up into the rest spread and consumers got `config.kit.kit.files` and so on. Top-level fields were unaffected, which is why the eslint recipe from sveltejs/eslint-plugin-svelte#1550 still worked.

Repro before the fix, with options passed to `sveltekit()` in `vite.config.js`:

```js
const result = await loadConfig('./', { traverse: false });
result.config.kit; // { kit: { files: {...}, ... } }
```

### DIFF
--- a/.changeset/tall-hats-shake.md
+++ b/.changeset/tall-hats-shake.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/load-config': patch
+---
+
+fix: don't double-nest kit options when loading from a vite config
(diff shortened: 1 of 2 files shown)


dummdidumm (approved):
Good Catch!

PR #3113, opened 2026-08-26
fix: hide imports from document symbols

colecrouter:
While not explicitly outlined, VSCode's current behavior for the TS LSP is to omit top-level imports from document symbols (see [#50829](https://github.com/microsoft/vscode/issues/50829) and [f1efd11](https://github.com/microsoft/vscode/commit/f1efd11c2a6da4533f65d66bf7921c552270ca4e)). This has the nice side effect of reducing clutter in the "Outline" view.

**Before**:
image

**After**:
image

Expanded tests to cover default, namespace, named, renamed, type-only, and `import = require` imports, while preserving other symbols.

### DIFF
--- a/.changeset/hide-import-document-symbols.md
+++ b/.changeset/hide-import-document-symbols.md
@@ -1,0 +1,5 @@
+---
+'svelte-language-server': patch
+---
+
+Hide imports from document symbols so they do not appear in the editor outline.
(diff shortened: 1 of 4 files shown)


jasonlyu123:
I am not really sure about this one. TypeScript 7's native LSP server doesn't filter out imports. So it'll show up again once TypeScript 7 support is out.

colecrouter:
Opened an issue here: https://github.com/microsoft/TypeScript/issues/64155

PR #3114, opened 2026-08-26
fix: hide generated template functions from document symbols

colecrouter:
Multi-line snippets currently generate garbage output. Detect generated anonymous functions by first checking if they are in a template & checking if they were written by the user, then skip those symbols.

**Before**:
image

**After**:
image

Tests cases include parametrized and non-parametrized snippets, as well as user-defined anonymous functions.

### DIFF
--- a/packages/language-server/src/plugins/typescript/TypeScriptPlugin.ts
+++ b/packages/language-server/src/plugins/typescript/TypeScriptPlugin.ts
@@ -107,6 +107,7 @@ import {
 import { CallHierarchyProviderImpl } from './features/CallHierarchyProvider';
 import { CodeLensProviderImpl } from './features/CodeLensProvider';
 import { WorkspaceSymbolsProviderImpl } from './features/WorkspaceSymbolProvider';
+import type { SvelteDocumentSnapshot } from './DocumentSnapshot';
 
 export class TypeScriptPlugin
     implements
@@ -326,6 +327,23 @@ export class TypeScriptPlugin
             }
 
             if (symbol.name === '<function>') {
+                // Generated functions in the template can cover arbitrary template text, which
+                // results in malformed entries. Keep functions that map to an expression the user
+                // actually wrote.
+                const symbolStartOffset = document.offsetAt(symbol.location.range.start);
+                // svelte2tsx replaces the instance script's end tag with the generated template
+                // callback. Its start therefore maps to scriptInfo.end, which is also considered
+                // inside the script by isInScript. Treat this boundary as template code here so the
+                // generated callback is filtered out.
+                const startsAtScriptEnd = symbolStartOffset === document.scriptInfo?.end;
+                if (
+                    startsAtScriptEnd ||
+                    (!isInScript(symbol.location.range.start, document) &&
+                        !this.isInUserWrittenTemplateFunction(tsDoc, symbolStartOffset))
+                ) {
+                    continue;
+                }
+
                 let name = getTextInRange(symbol.location.range, document.getText()).trimLeft();
                 if (name.length > 50) {
                     name = name.substring(0, 50) + '...';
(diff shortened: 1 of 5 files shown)


colecrouter:
I've since noticed this solves a second bug with comments inside templates:

image
image

—though, this is also a TextMate grammar bug (and I'd interject that Prettier shouldn't be forcefully collapsing template brackets here in the first place). I will open a separate PR for that.

Edit: See #3115 & #3116.

packages/language-server/src/plugins/typescript/TypeScriptPlugin.ts
```diff
@@ -326,6 +331,16 @@ export class TypeScriptPlugin
             }
 
             if (symbol.name === '<function>') {
+                // Generated functions in the template can cover arbitrary template text, which
+                // results in malformed entries. Keep functions that map to an expression the user
+                // actually wrote.
+                if (
+                    !isInScript(symbol.location.range.start, document) &&
+                    !isInFunctionExpression(tsDoc.svelteNodeAt(symbol.location.range.start))
```
jasonlyu123: Instead of finding the exact node, then walk up the AST tree. It'll be slightly more performant to walk down the AST until a function expression or a snippet block where the document symbol position is inside the start tag(`{#snippet ...}`)

colecrouter: Thanks for the feedback. I've applied that change.

jasonlyu123:
While testing the result, I found another generated function in the document symbols. Can you also fix this one? It will be mapped to the script end tag `</script>`. So we can probably filter it out with a simple string check or by checking if the position is the end tag. 

```svelte
<script>
  import Test1 from "./Test1.svelte";
</script>

<Test1></Test1>
```

colecrouter:
> While testing the result, I found another generated function in the document symbols. Can you also fix this one? It will be mapped to the script end tag `</script>`,

@jasonlyu123 Could you share which document you are checking against? I wasn't able to recreate any issues across a generic set of Svelte template syntax.

jasonlyu123:
The one in my example. It seems like it only reproduces when there isn't a final newline in the file.

colecrouter:
Thanks @jasonlyu123, I wouldn't have caught this; Prettier guarantees trailing newlines for fixtures in this repo.

This uncovers a pre-existing quirk to how `isInScript` is implemented.  It treats the `<` in `</script>` as within the script, which causes my code to end up in a bogus state. The trailing newline causes the invalid symbol to be silently discarded.

Changing this behavior risks changing IDE behavior (e.g. cursor before `</script>` could report as outside of script tag), so I have patched my code to watch for this specific case. Let me know if you agree with my direction.

packages/language-server/src/plugins/typescript/TypeScriptPlugin.ts
```diff
@@ -326,6 +327,22 @@ export class TypeScriptPlugin
...
             if (symbol.name === '<function>') {
+                // Generated functions in the template can cover arbitrary template text, which
+                // results in malformed entries. Keep functions that map to an expression the user
+                // actually wrote.
+                const symbolStartOffset = document.offsetAt(symbol.location.range.start);
+                // svelte2tsx replaces the instance script's end tag with the generated template
+                // callback. Its start therefore maps to scriptInfo.end, which is also considered
+                // inside the script by isInScript. Treat this boundary as template code here so the
+                // generated callback is filtered out.
+                const startsAtScriptEnd = symbolStartOffset === document.scriptInfo?.end;
```
jasonlyu123, line 338: If `startsAtScriptEnd ` is true, we don't need to check the next condition. It'll be in the script tag. And there is no space left for a function, so the symbol won't be a user-written function.

jasonlyu123 (approved):

colecrouter:
Thanks, rebased.

PR #3116, opened 2026-08-31
fix: highlight comments on first line of template expressions

### LINKED ISSUE #3115: Comments on the first line of a template don't highlight
### Describe the bug

Image

Image

Both compile, and the former is what Prettier suggests.

### Reproduction

Inside VSCode enter:

```svelte
{// Comment 1
// Comment 2
"true" == "true"
	? "foo"
	: "bar"}
```

### Expected behaviour

Comment receives comment symbol/grammar.

### System Info

    OS: macOS 26.6.2
    Node: 26.7.0
    VSCode: 1.135.0

svelte-language-server, svelte2tsx

_No response_

colecrouter:
Fixes #3115.

The `/` in comments always triggers the grammar for the "end template block" (e.g. `{/if}`).

This PR adds a negative lookahead to prevent `//` or `/*` from hitting that branch. It also includes a small test fixture for comments inside templates.

**Before**:
image

**After**:
image


jasonlyu123 (approved):
👍

